PR #4, opened 2015-12-09
MLDB-1182 clang support

jeremybarnes:
This allows building with

```
make ... toolchain=clang
```

to compile with the clang compiler.  Almost all warnings are enabled; they helped to drive significant cleanups in the code.

The default toolchain is gcc.

There is still one test failing (the git dataset test); it's under investigation as it only started failing once the latest master was pulled in.

I'd like to get this one merged soon-ish as it touches a lot of files.


arch/bit_range_ops.h
```diff
@@ -1,18 +1,18 @@
-// This file is part of MLDB. Copyright 2015 Datacratic. All rights reserved.
-
 /* bit_range_ops.h                                                 -*- C++ -*-
    Jeremy Barnes, 23 March 2009
    Copyright (c) 2009 Jeremy Barnes.  All rights reserved.
 
+   This file is part of MLDB. Copyright 2015 Datacratic. All rights reserved.
```
jeremybarnes: The extra header stopped emacs from being able to detect C++ files, which caused indenting issues, so I fixed the headers in the files I edited.  We should probably do a round of improvements to the header messages.

ml/bucketing_probabilizer.cc
```diff
@@ -271,7 +271,7 @@ getProbForBucket(size_t bucket) const
     if(ctr_buckets.size()==0)
         throw ML::Exception("ctr buckets is empty!");
 
-    if(bucket < 0 || bucket > ctr_buckets.size() - 1)
+    if(bucket >= ctr_buckets.size())
```
jeremybarnes: Bucket is an unsigned integer, so it can't be negative.

ext/judy/judy.mk
```diff
@@ -26,6 +26,6 @@ LIBJUDY_SOURCES := \
 LIBJUDY_LINK :=
 
 # gcc 4.9 compilation requirements
-$(eval $(call set_compile_option,$(LIBJUDY_SOURCES),-fno-strict-aliasing -Wno-array-bounds))
+$(eval $(call set_compile_option,$(LIBJUDY_SOURCES),-fno-strict-aliasing -Wno-array-bounds -Wno-all -Wno-error))
```
jeremybarnes: You're right; it's not required.

PR #5, opened 2015-12-09
Added Excel .xlsx importer (MLDB-1011)

jeremybarnes:
Adds a basic, experimental .xlsx importer, with a basic set of
tests that uses the input files of an external .xlsx -> .csv
converter.


steadtler:
Looks good to me +1

PR #6, opened 2015-12-10
MLDB-1186 make default sql builtins functions available in column expr

steadtler:


sql/binding_contexts.cc
```diff
@@ -169,13 +169,24 @@ doGetFunction(const Utf8String & tableName,
...
+         return {[=] (const std::vector<ExpressionValue> & args,
                  const SqlRowScope & context)
             {
                 auto & col = static_cast<const ColumnContext &>(context);
                 return fn(col.columnName, args);
             },
             std::make_shared<Utf8StringValueInfo>()};
+    }
+
+    auto sqlfn = SqlBindingScope::doGetFunction(tableName, functionName, args);
```
guyd: Will this break training code like svd which do not support operation on columns but support column expression?

guyd: After discussion with Mathieu, some check will have to be written to ensure that the training does not fail badly on column expression with functions.  I will do that as part of MLDB-1104.

guyd:
Mathieu, are you planning to add the test case you mentioned to me?

steadtler:
@guyd le vrai test ca va être Mich! J'ai ajouté un test de base en attendant.

guyd:
+1

PR #7, opened 2015-12-10
Mldb 1104 combined sql and dataset

guyd:
- combined sql fragments into a single sql statement
- modified the svd, svm, tsne and kmeans
- added validation that the select statement fullfil the requirements of the above training algorithms
- adapted the tests accordingly


sql/sql_expression.cc
```diff
@@ -3560,6 +3581,58 @@ SelectStatementDescription()
...
+void
+InputQueryDescription::
+parseJsonTyped(InputQuery * val,
+               JsonParsingContext & context) const
+{
+    if (context.isString())
+        val->stm = make_shared<SelectStatement>(SelectStatement::parse(context.expectStringUtf8()));
+    else if (context.isObject()) {
+        Json::Value v = context.expectJson();
+        //        val->stm.reset(new DatasetSelectStatement(v));
```
steadtler: ?

steadtler:
That makes it so much cleaner thank you +1

sql/sql_expression.cc
guyd: left-over from a different impl - will remove

PR #9, opened 2015-12-11
Mldb 1181 - add an option to not push empty rows

steadtler:
In the transform procedure.


plugins/sql_functions.cc
```diff
@@ -484,13 +490,16 @@ run(const ProcedureRunConfig & run,
                     cols.push_back(c);
                 }
 
-                auto & rows = accum.get();
-                rows.reserve(10000);
-                rows.emplace_back(RowName(calc.at(0).toString()), std::move(cols));
+                if (!skipEmptyRows || cols.size() > 0)
```
guyd: Are you sure this is the condition you want?

guyd:
+1

nicolaskruchten:
Hmmm I'm not sure I like the process used in this PR... There was no DOD for this option. Why did we need a new option for this as opposed to just doing it within the SQL query of the transform?

steadtler:
We found no elegant and performant way of doing this - it either takes two queries and some custom code from the client, or one much more expensive query. You weren't there so I checked with Frank beforehand.

nicolaskruchten:
The "SQL way" to do this, to my mind, would be to repeat the `select` expression in the `where` and for MLDB to cache the selection to reduce the expense. That said, this is acceptable for now.

PR #11, opened 2015-12-14
Mldb 1190 segfault sqlexpr jseval biz

guyd:
When running the jseval query below a seg fault was occurring after several calls.  The problem is that we were performing a static_cast on an object of the wrong type.  This happens when a function calls BoundSqlExpression.constantValue() on an SQL expression bound with a ReadThroughBindingContext.  This is the case of jseval and regexp functions when called from within an sql.expression.


guyd:
@jeremybarnes This is the problem I was describing earlier this morning.

steadtler:
I guess you meant static_cast in your description, not const_cast :)

sql/binding_contexts.cc
```diff
@@ -32,8 +32,19 @@ rebind(BoundSqlExpression expr)
                      ExpressionValue & storage)
         -> const ExpressionValue &
         {
-            auto & row = static_cast<const RowContext &>(context);
```
steadtler: Could / should this be a virtual method instead?

steadtler:
I'll let you decide for the virtual method, take it as a suggestions, +1 when you're decided.

guyd:
regarding the description, yes my bad, I meant static_cast.

jeremybarnes:
Hmmm, this smells like a design issue we're plastering over.

guyd:
It not only smells, it is a plaster!  I will revisit once I'm done with my current task.

guyd:
I revisited the code.  This is more aligned with the design.  Let me know what you think.

steadtler:
Well that is much better heh! + 1

jeremybarnes:
Indeed... much much better :)

PR #12, opened 2015-12-16
Don't expose CurlWrapper in HttpRestProxy (MLDB-1200)

jeremybarnes:
Simple one: we don't expose the CurlWrapper so that the HttpRestProxy can be part of our API without pulling too many extra dependencies.

### DIFF
--- a/http/http_rest_proxy.h
+++ b/http/http_rest_proxy.h
@@ -1,9 +1,8 @@
-// This file is part of MLDB. Copyright 2015 Datacratic. All rights reserved.
-
 /* http_rest_proxy.h                                               -*- C++ -*-
    Jeremy Barnes, 10 April 2013
    Copyright (c) 2013 Datacratic Inc.  All rights reserved.
 
+   This file is part of MLDB. Copyright 2015 Datacratic. All rights reserved.
 */
 
 #pragma once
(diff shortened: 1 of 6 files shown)


http/http_rest_proxy.h
```diff
@@ -258,36 +256,28 @@ struct HttpRestProxy {
...
     struct Connection {
-        Connection(CurlWrapper::Easy * conn,
+        Connection(ConnectionHandler * conn,
                    HttpRestProxy * proxy)
             : conn(conn), proxy(proxy)
         {
         }
 
-        ~Connection();
+        ~Connection() noexcept;
```
FinchPowers: Je croyais que noexcept pour un destructeur était implicite à partir de C++11. Est-ce que je me trompe?

FinchPowers:
+1

http/http_rest_proxy.h
jeremybarnes: Je ne suis pas sur.

wsourdeau: @FinchPowers  a raison, pour une certaine définition de "par défaut": (from http://en.cppreference.com/w/cpp/language/noexcept_spec):

"Inheriting constructors and the implicitly-declared default constructors, copy constructors, move constructors, destructors, copy-assignment operators, move-assignment operators are all noexcept(true) by default, unless they are required to call a function that is noexcept(false), in which case these functions are noexcept(false). "
+
"Any user-defined destructor is noexcept(true) by default, unless the declaration specifies otherwise, or the destructor of any base or member is noexcept(false). "

PR #14, opened 2015-12-16
Removed __int128_t from interface (MLDB-1201)

jeremybarnes:
This hoists the non-standard type (which is also not available on 32 bit machines or some compilers) out of the header files, and into implementation files.


steadtler:
+1

PR #15, opened 2015-12-16
Fixed out of bounds D value (MLDB-1202)

jeremybarnes:
Looks like sometimes the SVD will return less columns than we asked for, but we calculated our loop bounds based upon what we asked for not what we got.  This should fix a potential out-of-bounds.

@mailletf please verify the fix.


steadtler:
Well there is nothing wrong with the code. +1

jeremybarnes:
Keeping the branch open until @mailletf verifies.

mailletf:
Did not crash again so I'd say fixed

PR #18, opened 2015-12-17
MLDB-1058: reduced string allocations/copies by using std::move on st…

wsourdeau:
…ring parameters

### DIFF
--- a/http/http_socket_handler.h
+++ b/http/http_socket_handler.h
@@ -1,6 +1,4 @@
-// This file is part of MLDB. Copyright 2015 Datacratic. All rights reserved.
-
-/* http_socket_handler.h                                           -*- C++ -*-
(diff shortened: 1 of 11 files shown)


http/http_socket_handler.h
```diff
@@ -1,6 +1,4 @@
-// This file is part of MLDB. Copyright 2015 Datacratic. All rights reserved.
-
-/* http_socket_handler.h                                           -*- C++ -*-
+/* http_socket_handler.h - This file is part of MLDB               -*- C++ -*-
```
guyd: I am pretty sure we want to keep the copyright notice.

wsourdeau: It's still present below. But I concede the style is not consistent with the rest.

guyd: Oh right.  I did not realized it was duplicated.

guyd:
+1

PR #20, opened 2015-12-17
Mldb 1058 sub trim copy replacement

wsourdeau:


jml/utils/testing/string_functions_test.cc
```diff
@@ -46,3 +46,27 @@ BOOST_AUTO_TEST_CASE( test_string_replace_in_place )
     num_found = replace_all(my_str, "$toRep", "hoho");
     BOOST_CHECK_EQUAL(my_str, "hohoBouyahoho");
 }
+
+BOOST_AUTO_TEST_CASE( test_string_trim )
```
guyd: please add these `""`, `"\t\t\ta"` and `"   "`.

guyd:
I guess you have notice that the boost implementation is not fast enough.  How bad is it?  Is the boost implementation more versatile and when we improve our trim implementation to handle these other cases we will loose the performance benefit?

wsourdeau:
@guyd the boost implementation is slower because each invocation makes a call to "std::locale", which is very costly. Most of the cases where we use "trim_copy" is on ascii strings, and even if we were using iso-8859-x or utf-8, the way the trimming is done in my implementation would still be valid.

guyd:
+1

PR #23, opened 2015-12-18
fixed the experiment procedure where training clause.

guyd:


nicolaskruchten:
+1 once the build clears

PR #26, opened 2015-12-19
MLDB-1212 csv import long quoted lines

jeremybarnes:
Fixes issues with parsing quoted CSV fields longer than 4096 bytes.

@mailletf FYI


mailletf:
Confirmed fix; thanks! +1

PR #27, opened 2015-12-21
Mldb 1104 combined sql and dataset

guyd:
- Grouped the probabilizer SQL clauses into a single one like `select expr as score, expr as label, expr as weight`.
- Move some code outside procedure.h to avoid polluting our interface.
- Minor refactoring to avoid code duplication.


plugins/probabilizer.cc
```diff
@@ -1,10 +1,10 @@
-// This file is part of MLDB. Copyright 2015 Datacratic. All rights reserved.
-
 /** probabilizer.cc
     Jeremy Barnes, 16 December 2014
     Copyright (c) 2014 Datacratic Inc.  All rights reserved.
 
-    Implementation of an algorithm to transform an arbitrary score into a
+    This file is part of MLDB. Copyright 2015 Datacratic. All rights reserved.
+    
+Implementation of an algorithm to transform an arbitrary score into a
```
mailletf: indentation

mailletf:
+1

PR #29, opened 2015-12-22
Streamlined and standardized ML::Spinlock (MLDB-1211)

jeremybarnes:
Very simple change.  We take the GCC-specific spinlock implementation we had, simplify it and make it use C++11 atomics.


guyd:
+1

FinchPowers:
+1 aussi ;)

PR #32, opened 2015-12-22
MLDB-1217: homogenisation of well known sql expressions

wsourdeau:


jeremybarnes:
+1 (apart from the build failure, which I suspect is not related since your changes are simple).

nicolaskruchten:
There were some comments on HipChat about this change breaking things... have those been resolved?

Guy said "@wolfgang On a different topic, I believe your changes to use SqlExpression::TRUE is breaking one of the test.  I looked a bit into it.  I think there might be a copy that do not copy the surface of the expression."

/cc @guyd @wsourdeau

wsourdeau:
@nicolaskruchten the fundamental cause has not been solved and I would like to solve it before merging.

Thus, -1 for now.

wsourdeau:
Oh, apparently I already merged it. +1 then. The problem itself is still on my list of things to solve though.

nicolaskruchten:
I'm confused... this PR has already been merged. do you have another one open that fixes the problem? Is there an issue associated with the bug

wsourdeau:
Context restored.... The test was failing due to the (reverted) change to those lines: https://github.com/mldbai/mldb/blob/master/plugins/experiment_procedure.cc#L313

Since there was only one failing test and since that part of the code is enabled only in the context of testing, I considered that it was safe to just revert that part and tackle the issue separately.

nicolaskruchten:
Conclusion following in-person conversation: this PR will be reverted in toto.

PR #33, opened 2015-12-22
added the 'try' spin lock since this is required by some code in mldb…

guyd:
… pro.


jeremybarnes:
This was superseded by #42 and is thus being closed.

PR #36, opened 2015-12-23
filter_stream can return object info (MLDB-1218)

jeremybarnes:
Provides access to information and metadata (etag, permissions, size, etc) that is already known for most implementations without requiring a separate, non-atomic call to getUriObjectInfo.


FinchPowers:
+1

PR #37, opened 2015-12-23
MLDB-1058: avoid allocating a std::string each time a ML::Parse_Conte…

wsourdeau:
…xt is instantiated

### DIFF
--- a/base/parse_context.cc
+++ b/base/parse_context.cc
@@ -28,7 +28,7 @@ namespace ML {
 /* PARSE_CONTEXT                                                             */
 /*****************************************************************************/
 
-const std::string Parse_Context::NO_FILE("-");
+const std::string Parse_Context::CONSOLE("-");
 
 Parse_Context::
 Parse_Context()
--- a/base/parse_context.cc
+++ b/base/parse_context.cc
@@ -28,6 +28,8 @@ namespace ML {
 /* PARSE_CONTEXT                                                             */
 /*****************************************************************************/
 
+const std::string Parse_Context::NO_FILE("-");
+
 Parse_Context::
 Parse_Context()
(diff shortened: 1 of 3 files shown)


jeremybarnes:
-1.  This will make the error messages much less helpful, as they won't tell us what the expression we were parsing was.  I agree with the end but not the means.  Can we do it a different way?

wsourdeau:
@jeremybarnes for case such as in sql_expressoin.cc, I understand, but in other cases it may not be relevant. The use of "NO_FILE" is still optional. Would it be ok for you if I reverted sql_expression.cc?

jeremybarnes:
No, the Parse_Context's error messages are a big part of the usability in all the places that they are used.  You're making a big change here when we only need a small one.  If a user of Parse_Context wants to have no filename, then it can do so externally, but that should be enabled at the caller.

nicolaskruchten:
What's the status of this PR? What's the performance improvement like from this?

wsourdeau:
@nicolaskruchten the technical aspect of the comments above have been discussed and settled with Jeremy before christmas. Benchmarks under way...

nicolaskruchten:
We still need the +1 from @jeremybarnes to merge.

wsourdeau:
@nicolaskruchten @jeremybarnes gain = 5kqps

base/parse_context.cc
```diff
@@ -28,6 +28,8 @@ namespace ML {
 /* PARSE_CONTEXT                                                             */
 /*****************************************************************************/
 
+const std::string Parse_Context::NO_FILE("-");
```
jeremybarnes: Should be "CONSOLE", not NO_FILE.

jeremybarnes:
With the renaming I just proposed, this is good to merge.

PR #38, opened 2015-12-23
MLDB-1058: "getResponseReasonPhrase" now returns a const string ref

wsourdeau:
This change enables a single instantiation of std::string per reason code, in order to reduce the number of allocations taking place during the returning of an HTTP response.

### DIFF
--- a/http/http_header.cc
+++ b/http/http_header.cc
@@ -20,6 +20,53 @@ using namespace std;
 using namespace ML;
 
 
+namespace {
+
+const string Continue("Continue");
+const string SwitchingProtocols("Switching Protocols");
+const string OK("OK");
+const string Created("Created");
+const string Accepted("Accepted");
+const string NonAuthoritativeInformation("Non-Authoritative Information");
+const string NoContent("No Content");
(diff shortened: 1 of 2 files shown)


jeremybarnes:
What is the rationale?  You should put one in the description.

Also, why not return a const char * from the function?

wsourdeau:
@jeremybarnes for the "const char *": because it's more idiomatic and safer. Technically, "const char *" can be freed without a compiler warning.

wsourdeau:
@jeremybarnes descriptoin amended.

jeremybarnes:
If you free a const char * from read-only memory where these strings come from, you will get a crash instantly.

The change is OK, I guess, but it's bigger than it needs to be and I really don't see the need for the static initialization everywhere.

+1 anyway.

nicolaskruchten:
What is the performance improvement from this PR?

wsourdeau:
@nicolaskruchten 2 kqps
I am still recovering the hair I've lost by benchmarking this.

nicolaskruchten:
ok. please merge, you already have a +1 from Jeremy

PR #40, opened 2015-12-23
MLDB-1058: stopped logging stats uselessly until performance is improved

wsourdeau:
Those statistics are not being used currently and the mechanism in use harms the performance of the systme.

### DIFF
--- a/server/mldb_server.cc
+++ b/server/mldb_server.cc
@@ -319,24 +319,6 @@ initCollections(std::string configurationPath,
     procedures->loadConfig();
     functions->loadConfig();
 
-    logRequest = [&] (const HttpRestConnection & conn, const RestRequest & req)
-        {
-            this->recordHit("rest.request.count");
-            this->recordHit("rest.request.verbs.%s", req.verb.c_str());
-        };
-
-    logResponse = [&] (const HttpRestConnection & conn,
-                       int code,
-                       const std::string & resp,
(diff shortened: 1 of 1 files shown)


jeremybarnes:
-1.  Let's put it behind a switch instead of removing it entirely.  In any case, it shouldn't cause much of a performance difference, does it?

nicolaskruchten:
I vote to remove this: there are no users of this data and it seems trivial to add back in. AFAIK it does have a performance impact.

nicolaskruchten:
@wsourdeau what is the performance improvement of this PR?

wsourdeau:
@nicolaskruchten 11kqps

nicolaskruchten:
@jeremybarnes thoughts?

wsourdeau:
Btw, personally, I think the idea of a command-line switch is reasonable. However, I am willing to dig further in the understanding of why these simple function calls take so much cpu time. Since those are basically no-ops, they should take no time at all, presumably...

jeremybarnes:
I would put it behind an if (false) {
}.  That way we can modify the condition to something that looks to see if anyone has registered a metrics handler, and have the metrics just work again.  For any production usage, we are going to want to keep them.

wsourdeau:
@jeremybarnes done

jeremybarnes:
+1

PR #41, opened 2015-12-23
MLDB-1058: restored in-house implementation of "trim"

wsourdeau:
Restored ML::trim, that was disabled without a bug report.


FinchPowers:
Still brisé.

wsourdeau:
Le test qui échoue n'est pas influencé par l'une ou l'autre des implémentations. C'est un hasard.

jeremybarnes:
The reason it was disabled without a bug report was that it broke the SQL parsing (which needs the behavior to include stripping all characters including newlines), and so was reverted.  Until we understand what's going on there, we shouldn't change it back.

Also, in the case that there is no space at the beginning or end of a string, we should simply return a copy of the input string.  In the shared string implementation of our compiler, this is a significant performance gain.

guyd:
@wsourdeau I created the JIRA to avoid any additional back-and-forth.  See https://datacratic.atlassian.net/browse/MLDB-1221.

nicolaskruchten:
-1

I will create a test case to catch the regressions which caused this revert.

wsourdeau:
@nicolaskruchten thanks!

nicolaskruchten:
test case created and issue assigned

wsourdeau:
@guyd all characters reported as spaces by "isspace" are now handled.

wsourdeau:
@nicolaskruchten  all test cases pass now. The demos needs to be tested yet.

guyd:
+1

nicolaskruchten:
The demos can be tested by building a docker container, booting it, and then executing the notebooks through Jupyter. If the tests pass it should be fine though, so please merge when ready.

wsourdeau:
(tests run 3 times, best result kept)
With fix: 80kqps
Without this fix: 75kqps
Gain: 5kqps

nicolaskruchten:
you already have a +1, please merge :)

On Wed, Jan 13, 2016 at 3:05 PM, Wolfgang Sourdeau <notifications@github.com
> wrote:

> (tests run 3 times, best result kept)
> With fix: 80kqps
> Without this fix: 75kqps
> Gain: 5kqps
>
> —
> <https://github.com/mldbai/mldb/pull/41#issuecomment-171417111>.
>



-- 
http://datacratic.com/

wsourdeau:
Squash merged

PR #42, opened 2015-12-23
Added back in try_lock to spinlock (MLDB-1218)

jeremybarnes:
This replaces #33 with a simpler implementation.  There were external plugins that were using the exposed try_lock() method, so it's been added back in.


FinchPowers:
+1

PR #44, opened 2015-12-23
MLDB-1058: added a "num-threads" parameter to mldb_runner

wsourdeau:


server/mldb_runner.cc
```diff
@@ -95,6 +95,7 @@ int main(int argc, char ** argv)
     options_description script_options("Script options");
     options_description plugin_options("Plugin options");
 
+    int numThreads(1);
```
FinchPowers: You should have left the default to 16 I suppose.

jeremybarnes:
Agreed with Mich.  Having not enough threads causes deadlocks.  It should always have at least 16 threads, and probably 32, unless it auto-scales the number of threads.

wsourdeau:
-1 until MLDB-1058-one-thread-per-loop is mergable.

wsourdeau:
@FinchPowers @jeremybarnes fixed.

server/mldb_runner.cc
```diff
@@ -134,6 +135,7 @@ int main(int argc, char ** argv)
         ("etcd-path", value(&etcdPath),
          "Base path in etcd")
 #endif
+        ("num-threads,t", value(&numThreads), "Number of workers threads")
```
guyd: workers -> worker

guyd:
+1

FinchPowers:
If MLDB can deadlock with a single thread I think we should make it impossible to set it to that value.

jeremybarnes:
We're missing an error message (or at least a loud warning) if we set num threads < 4 or so.

nicolaskruchten:
Agreed that we should not allow fewer than 4 threads.

wsourdeau:
@FinchPowers @jeremybarnes @nicolaskruchten I have set the default number of threads to 16, and a strong requirement of 4. The program dies with an error message when using a lesser value.

nicolaskruchten:
+1, please merge

PR #45, opened 2015-12-23
Mldb 1058 various string opts

wsourdeau:


FinchPowers:
+1

sql/sql_expression.cc
```diff
@@ -1208,7 +1208,7 @@ parse(ML::Parse_Context & context, int currentPrecedence, bool allowUtf8)
         ExpressionValue constant;
         if (matchConstant(context, constant, allowUtf8)) {
             lhs = std::make_shared<ConstantExpression>(constant);
-            lhs->surface = ML::trim(token.captured());
+            lhs->surface = token.captured();
```
guyd: Why removing the trim here?

rest/http_rest_service.cc
```diff
@@ -321,9 +321,13 @@ std::string
...
-    std::string s = Date::now().print(9) + ML::format("%d", random());
-    uint64_t jobId = CityHash64(s.c_str(), s.size());
-    return ML::format("%016llx", jobId);
+    union {
+        char string[sizeof(long int)];
+        long int rnd;
+    } requestId;
+    requestId.rnd = random();
+    auto jobId = CityHash64(requestId.string, sizeof(long int));
+    return to_string(jobId);
```
guyd: What would be the effect of having two active requests with the same id?  I am asking because the new code seems to be more likely to collision than the previous one.

rest/http_rest_endpoint.cc
```diff
@@ -204,12 +204,48 @@ sendResponse(int code,
...
              RestParams headers)
 {
-    for (auto & h: endpoint->extraHeaders)
-        headers.push_back(h);
 
-    putResponseOnWire(HttpResponse(code,
-                                   std::move(contentType), std::move(body),
-                                   std::move(headers)));
+    string responseStr;
+    responseStr.reserve(16384 + body.length());
```
guyd: Isn't 16K an order of magnitude too much for most response?

rest/http_rest_endpoint.cc
```diff
@@ -204,12 +204,48 @@ sendResponse(int code,
              std::string body, std::string contentType,
              RestParams headers)
 {
-    for (auto & h: endpoint->extraHeaders)
-        headers.push_back(h);
```
guyd: If we are in-lining the code to avoid this push_back loop then I would be tempted to use boost::join instead which should avoid all copies or simply modify the HttpResponse to take two sets of headers instead.

PR #46, opened 2015-12-28
Mldb 1189 join options

steadtler:
For Left, Right and Outer join.

The idea in both the regular and pipeline executors is that if we want the outer join on a side, we replace the condition ("where") on that side with a TRUE (to get all row) and instead calculate the condition in the select. Then we can check its value to record the row anyway.

Obviously we also need to record the row whenever it has a valid condition but fails to match with the other side.


builtin/joined_dataset.cc
```diff
@@ -181,115 +159,8 @@ struct JoinedDataset::Itl
...
-                {
-                    auto generator = dataset.queryBasic
-                    (context, side.select, side.when, *side.where, side.orderBy,
-                     0, -1, true /* allowParallel */);
-                    auto rows = generator(-1);
-                
-                    if (debug)
-                        cerr << "got rows " << jsonEncode(rows) << endl;
-                    
-                    // Now we extract all values and re-sort.  This is necessary because
```
steadtler: Could I get a comment on that @jeremybarnes ? Now that I put it in an embedding I explicitely get a single value not a row. Im not sure how multiple values could have happened before and it doesnt break any test.

steadtler:
Ps: no hurry for the review, enjoy your holidays

builtin/joined_dataset.h
```diff
@@ -25,6 +25,7 @@ struct JoinedDatasetConfig {
     std::shared_ptr<TableExpression> left;
     std::shared_ptr<TableExpression> right;
     std::shared_ptr<SqlExpression> on;
+    JoinQualification qualification;
```
guyd: We should consider grouping these fields into one SQL expression as we have done at many places.

builtin/joined_dataset.cc
```diff
@@ -181,115 +159,8 @@ struct JoinedDataset::Itl
...
-                                const RowHash & rightHash = std::get<2>(*it2a);
-
-                                if (debug)
-                                    cerr << "rows " << leftName << " and " << rightName << " join on value " << val1 << endl;
-                                
-                                gotRow(leftName, leftHash, rightName, rightHash);
-                            }
-                        }
-                    }
+            makeJoinFast(condition, context, left, right, joinConfig.qualification);
```
guyd: Consider renaming the function to mention that this is a join with a constant where.  The fact that this is handling the easier and faster case could be put as a comment.

builtin/joined_dataset.cc
```diff
@@ -299,10 +170,18 @@ struct JoinedDataset::Itl
                     // This has xxx results
```
guyd: noise?

builtin/joined_dataset.cc
```diff
@@ -299,10 +170,18 @@ struct JoinedDataset::Itl
...
+                    for (; i+ 2 < res->values.size(); i+=2)
+                    {
+                        leftNameUtf8 += "-" + res->values.at(i).toUtf8String();
+                    }                        
 
-                    gotRow(leftName, leftName, rightName, rightName);
+                    RowName leftName(leftNameUtf8);
+                    RowName rightName(res->values.at(i).toUtf8String());
+
+                    recordJoinRow(leftName, leftName, rightName, rightName);
```
guyd: The function can take only the row names and compute the row hashes internally.

builtin/joined_dataset.cc
```diff
@@ -360,6 +239,223 @@ struct JoinedDataset::Itl
...
+            -> std::vector<std::tuple<ExpressionValue, RowName, RowHash> >
+            {
+                auto condition = side.where;
+
+                std::vector<std::shared_ptr<SqlExpression> > clauses = { side.selectExpression };
+
+                if (outer)
+                {
+                    //return all rows
+                    condition = SqlExpression::TRUE;
```
guyd: Why is this set only for outer?  I thought that this function was call only for constant true where expression.

builtin/joined_dataset.cc
```diff
@@ -360,6 +239,223 @@ struct JoinedDataset::Itl
...
+                                          //"joinOn", joinConfig.on);
+            }
+            break;
+        }
+        case AnnotatedJoinCondition::EQUIJOIN: {
+            // Join on f(leftrow) = f(rightrow)
+            break;
+        }
+        default:
+            throw HttpReturnException(400, "Join expression requires an equality operator; needs to be in the form f(left) = f(right)",
```
guyd: If I am not mistaken the equality expression is required only for inner and outer joins.  We should modify the error message to be more specific.

builtin/joined_dataset.cc
```diff
@@ -360,6 +239,223 @@ struct JoinedDataset::Itl
...
+                                      "condition", condition);
+        }
+
+        // Finally, perform the join
+        // We keep a list of the row hashes of those that join up
+        auto it1 = leftRows.begin(), end1 = leftRows.end();
+        auto it2 = rightRows.begin(), end2 = rightRows.end();
+
+        while (it1 != end1 && it2 != end2) {
+            // TODO: there could be multiple values...
```
guyd: Is this tracked somewhere?

sql/execution_pipeline_impl.cc
```diff
@@ -743,6 +845,27 @@ take()
             } while (r && r->values.back() < lField);
         }
     }
+
+    if (outerLeft && l)
+    {
+        l->values.pop_back();
+        l->values.emplace_back(ExpressionValue("", Date()));
```
guyd: I am really puzzled by what we are trying to do here.  A comment would help.

sql/sql_expression.cc
```diff
@@ -2960,6 +2960,20 @@ printJsonTyped(const SelectExpression * val,
...
+/*****************************************************************************/
+/* JOIN QUALIFICATION                                                        */
+/*****************************************************************************/
+
+DEFINE_ENUM_DESCRIPTION(JoinQualification);
+
+JoinQualificationDescription::
+JoinQualificationDescription()
+{
+    addValue("JOIN_INNER", JOIN_INNER, "Inner join");
```
guyd: So "join" is no longer supported.  I thought it would be an inner join by default.

guyd: Any reason why this is not the more natural wording  "INNER_JOIN"?

sql/sql_expression_operations.cc
```diff
@@ -1112,6 +1113,7 @@ bind(SqlBindingScope & context) const
                 std::make_shared<BooleanValueInfo>()};
     }
     else if (op == "NOT" && !lhs) {
+        cerr << "not";
```
guyd: debug code

builtin/joined_dataset.cc
steadtler: In many (other) cases we already have the rowHash so we'll avoid computing it twice

testing/MLDB-180-basic-join.js
```diff
@@ -91,6 +91,8 @@ expected = [
    [ "ex2-ex5-ex5", 2, 4, 2, 2, 2, 2 ]
 ];
 
+//THIS HERE
```
guyd: noise

guyd:
I would add more test for the join dataset and other use of join in procedure for example.

guyd:
I am not familiar with the join implementation so there are many parts of the code I do not fully grasp.

builtin/joined_dataset.cc
steadtler: ugh, I renamed a parameter with a local, I'll change that. The crossWhere are the non-pivot conditions, but as I understand each side of the pivot can still have conditions in the fast path.

builtin/joined_dataset.cc
steadtler: This is originally Jeremy's comment, ref. my question to him above.

sql/sql_expression.cc
steadtler: I think its nice to have all members of the same Enum start by the same prefix.

guyd: I was more concerned with the associated string.  If I am not mistaken, someone would have to specify "JOIN_INNER" in a given JSON payload which, I feel, is not natural.

nicolaskruchten:
are we waiting for a +1 from Jeremy to merge this? I will take care of the documentation for this before we do the next release.

testing/MLDB-180-basic-join.js
```diff
@@ -251,4 +251,105 @@ expected = [
...
+// MLDB-1189 OUTER JOINS
+
+expected = [
+   [ "_rowName", "test1.x", "test1.y", "test2.x", "test2.z" ],
+   [ "ex1-ex4", 1, 2, 1, 2 ],
+   [ "ex1-ex6", 1, 2, null, 3 ],
+   [ "ex1-ex5", 1, 2, 2, 2 ]
+];
+
+testQuery('SELECT * FROM test1 INNER JOIN test2 ON test1.x = 1',
```
nicolaskruchten: need test case for the optional use of the `INNER` keyword:

* `INNER JOIN` === `JOIN`

need test cases for the optional use of the `OUTER` keyword:

* `OUTER JOIN` === `FULL OUTER JOIN` === `FULL JOIN` 
* `RIGHT OUTER JOIN` === `RIGHT JOIN`
* `LEFT OUTER JOIN` === `LEFT JOIN`

need test cases for N-way joins (where N > 2 :)

sql/sql_expression.cc
jeremybarnes: This is an internal class, which is only printed out for debugging.  It's not part of the MLDB API so it probably doesn't really matter.

builtin/joined_dataset.cc
jeremybarnes: The data model allows for multiple values for the same column with different timestamps.  It's extensively used eg in lookalike modelling, so it's not an optional extra.  In general, we can't expect that joinable rows can be put into an embedding; if they can it's an optimization but it won't always be possible.  So we probably need to have the slow and fast paths that are chosen based upon the incoming data.

testing/MLDB-180-basic-join.js
steadtler: As discussed, we'll solve the pre-existing 3+ way join issue in another PR.

builtin/joined_dataset.cc
steadtler: So as discussed it seems right now, pre and post this PR, join conditions will operate on the LATEST value, which is what we want *for the moment*, until we put the proper operators in place

nicolaskruchten: for now the ON clause should behave like the WHERE clause: equality and math operators operate on the latest value only.

guyd:
@nicolaskruchten It would be great if you can add examples of each of the join type in the doc.

guyd:
I am fine with the latest changes and very happy with the extra coverage in the tests.  :-)

nicolaskruchten:
+1 to merge from me. Not sure if we still need @jeremybarnes to weigh in?

sql/execution_pipeline_impl.cc
```diff
@@ -743,6 +845,29 @@ take()
             } while (r && r->values.back() < lField);
         }
     }
+
+    //Return unmatched rows if we have a LEFT/RIGHT/OUTER join
+    //Fill unmatched with empty values
+    if (outerLeft && l)
+    {
+        l->values.pop_back();
+        l->values.emplace_back(ExpressionValue("", Date()));
```
jeremybarnes: Date() is certainly wrong.  It should be Date::notADate().  Date() returns 1-1-1970.

sql/execution_pipeline_impl.cc
```diff
@@ -743,6 +845,29 @@ take()
...
+        auto result = std::move(l);                
+        l = left->take();
+        return result;
+    }
+
+    if (outerRight && r)
+    {
+        r->values.pop_back();
+        r->values.insert(r->values.begin(), ExpressionValue("", Date()));
```
jeremybarnes: Same here.  Date() should basically never be used.

sql/sql_expression.h
```diff
@@ -1332,6 +1308,10 @@ struct TableExpression: public std::enable_shared_from_this<TableExpression> {
     */
     virtual UnboundEntities getUnbound() const = 0;
 
+    /** Match join qualification (Inner, left, right, full)
```
jeremybarnes: Do we need to expose this and JoinQualification as part of the API?  I would have them as free functions defined inside the class.

jeremybarnes:
Nice work.  I have some comments about the dates (Date() should not be used) and one about leaking an internal class into the API.  Once addressed, +1.

sql/execution_pipeline_impl.cc
steadtler: Right.

PR #47, opened 2016-01-04
[MLDB-1104] merged the bucketize query fields into one

guyd:


plugins/bucketize_procedure.cc
```diff
@@ -127,12 +117,10 @@ run(const ProcedureRunConfig & run,
 {
     SqlExpressionMldbContext context(server);
 
-    auto boundDataset = procedureConfig.inputDataset->bind(context);
+    auto boundDataset = procedureConfig.inputData.stm->from->bind(context);
 
     SelectExpression select(SelectExpression::parse("1"));
     vector<shared_ptr<SqlExpression> > calc;
-    ssize_t offset = 0;
```
guyd: @FinchPowers Not sure why passing an offset and limit was not allowed in bucketize.  Now this is allowed so if there  is a good reason for not allowing it, let me know.

steadtler:
+ 1

guyd:
C'est un + 1 ou un - 1? :-)

steadtler:
whaaaat? c'etait un +1!

PR #48, opened 2016-01-04
Mldb 1104 transform

guyd:
Note that the changes to the bucketize procedure are part of a different PR.

- modified the transform procedure to take only one SQL statement as input.
- adapted the tests accordingly
- adapted the demos and tutorials (changes to python notebooks generate a lot of noise)

Tests and benchmarks in mldbpro have also been modified.  Another PR on that project will be created.


steadtler:
Fun times. +1!

PR #49, opened 2016-01-05
Mldb 1198 sum inconsistency test

steadtler:
Make vector aggregators ignore values with multiple timestamps.


sql/expression_value.cc
```diff
@@ -1438,6 +1438,67 @@ forEachColumnDestructive(const std::function<bool (Id & columnName, ExpressionVa
     }
 }
 
+ExpressionValue::Row
+ExpressionValue::
+getFiltered(const VariableFilter & filter /*= GET_LATEST*/) const
```
guyd: This is very similar to doSearchRow.  Can we avoid the code duplication?

guyd:
You might want to confirm that this is also fixing the `count({*})` problem as reported in the JIRA.  Other than that +1

PR #51, opened 2016-01-05
MLDB-1205 make sure we skip newlines in sql_expression

steadtler:


sql/sql_expression.cc
```diff
@@ -1119,8 +1119,10 @@ parse(ML::Parse_Context & context, int currentPrecedence, bool allowUtf8)
             lhs = arg;  
         }
 
-        context.skip_whitespace();
+        cerr << "test1" << endl;
```
guyd: debug code

sql/sql_expression.cc
```diff
@@ -1119,8 +1119,10 @@ parse(ML::Parse_Context & context, int currentPrecedence, bool allowUtf8)
...
+        skip_whitespace(context);
         context.expect_literal('}');
+        cerr << "test2" << endl;
```
guyd: here also

guyd:
+1

sql/sql_expression.cc
steadtler: the hell, I removed that before commit. Git is wierd sometimes

PR #53, opened 2016-01-05
MLDB-1210 change error message

steadtler:
As requested by Nick


guyd:
+1

nicolaskruchten:
This message only shows up in that specific case, right? Or what else can trigger that error response?

steadtler:
Obviously I havent tested all possible error cases, but it should only show up in the aggregator or group by queries

PR #56, opened 2016-01-08
Mldb 1226 persist model

guyd:
- modified the kmeans procedure to persist the centroids as a model file - that model is necessary to create a kmeans function
- similarly modified tf-idf procedure to persist the document frequencies of the terms in the corpus along with the corpus size - that model is also necessary when creating a kmeans function
- adapted and augmented the test coverage of the kmeans and tf-idf functions

### DIFF
--- a/plugins/tfidf.cc
+++ b/plugins/tfidf.cc
@@ -243,8 +243,11 @@ apply(const FunctionApplier & applier,
 
                 Utf8String word = get<0>(col).toUtf8String();
                 const CellValue& cell = get<1>(col);
-                maxNt = std::max(maxNt, cell.toInt());
-                bagOfWords[word] = cell.toInt();               
+                // if the value is not set it is because the term
+                // was never seen in the corpus
+                int64_t value = cell.isInteger() ? cell.toInt() : 0;
+                maxNt = std::max(maxNt, value);
+                bagOfWords[word] = value;               
             }
 
             return true;
@@ -289,20 +292,20 @@ apply(const FunctionApplier & applier,
     }
 
     // the different possible IDF scores
-    auto idf_unary = [=] (double numberOfRelevantDoc) {
+    auto idf_unary = [=] (unsigned long long numberOfRelevantDoc) {
         return 1.0f;
     };
-    auto idf_inverse = [=] (double numberOfRelevantDoc) {
-        return std::log(N / (1 + numberOfRelevantDoc));
+    auto idf_inverse = [=] (unsigned long long numberOfRelevantDoc) {
+        return std::log(N / std::max(1ULL, numberOfRelevantDoc));
     };
-    auto idf_inverseSmooth = [=] (double numberOfRelevantDoc) {
-        return std::log(1 + (N / (1 + numberOfRelevantDoc)));
+    auto idf_inverseSmooth = [=] (unsigned long long numberOfRelevantDoc) {
+        return std::log(1 + (N / std::max(1ULL, numberOfRelevantDoc)));
     };
-    auto idf_inverseMax = [=] (double numberOfRelevantDoc) {
-        return std::log(1 + (maxNt)/ (1 + numberOfRelevantDoc));
+    auto idf_inverseMax = [=] (unsigned long long numberOfRelevantDoc) {
+        return std::log(1 + (maxNt)/ std::max(1ULL, numberOfRelevantDoc));
     };
-    auto idf_probabilistic_inverse = [=] (double numberOfRelevantDoc) {
-        return std::log((N - numberOfRelevantDoc) / (1 + numberOfRelevantDoc));
+    auto idf_probabilistic_inverse = [=] (unsigned long long numberOfRelevantDoc) {
+        return std::log((N - numberOfRelevantDoc) / std::max(1ULL, numberOfRelevantDoc));
     };
 
     std::function<double(double)> idf_fct = idf_unary;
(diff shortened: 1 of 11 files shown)


plugins/tfidf.cc
```diff
@@ -325,13 +390,23 @@ apply(const FunctionApplier & applier,
...
+
     // Compute the score for every word in the input
+    //cerr << "corpus size " << corpusSize << " document size " << documentSize << endl;
+
     for (auto& col : inputVal.getRow() ) {
-        Utf8String colName = std::get<0>(col).toUtf8String();
+        Utf8String term = std::get<0>(col).toUtf8String(); // the term is the columnName
+        double frequency = (double) std::get<1>(col).getAtom().toUInt() / documentSize;
 
-        double frequency = std::get<1>(col).toDouble() / documentCount;
```
guyd: @mailletf @steadtler Here we divide the term frequency (# of times a term appears in the document) by the document size (the number of terms it contains).  This is a good approach in some cases especially when documents are varying in size greatly but it may not be what the user wants (BM25 is likely a better option).  Since the term frequencies are passed as an input to the tf-idf function, I suggest we do not perform that normalization here and let the user decide if he/she wants to use raw term frequencies or normalized ones.  Do you agree?

steadtler:
I guess the future EM clustering should also save clusters as a model file?

guyd:
I guess so for uniformity.

nicolaskruchten:
@guyd @steadtler yes please for the model file :)

re: normalization, I would like to get @mailletf 's opinion

plugins/tfidf.cc
mailletf: Agreed. Normalizing by docSize should be a tf weighting scheme.

jeremybarnes:
I don't like this change.  The point of having the kmeans write to a dataset was to avoid having custom model file formats: a model in the end is a dataset containing its parameters/

For the kmeans function, we should use a sql.query function that operates on the centroid dataset and queries the nearest neighbours.  Or that was at least the original intent.

guyd:
@jeremybarnes If we follow the dataset approach, how would we keep the extra parameters that are required by the model?  Kmeans does not have such parameters but tf-idf does.

nicolaskruchten:
kmeans does have the extra parameter: metric

guyd:
Sorry.  That's right.

nicolaskruchten:
@jeremybarnes the problem is that a dataset is not enough to fully define the model: we need associated data (metric, numdocs) and then we need to be able to very simply serialize/deserialize. Artifacts fit the bill well, and in any case that's what we use for SVD, Classifier, StatsTable etc. The point of what Guy is working on is standardization.

jeremybarnes:
We would store the extra parameters in rows in the dataset.  Alternatively, assuming that the training procedure creates the function, it would create the function with the correct parameters which would be passed to the function as part of its configuration.

jeremybarnes:
Having model parameters be in a dataset is really powerful and introspectable, and I think we should standardize *towards* this, not away from it.

plugins/tfidf.cc
```diff
@@ -23,9 +22,73 @@
 #include "mldb/server/analytics.h"
 #include "mldb/types/any_impl.h"
 #include "mldb/types/optional_description.h"
+#include "mldb/vfs/filter_streams.h"
 
 using namespace std;
 
+namespace {
+void
+serialize(ML::DB::Store_Writer & store,
```
jeremybarnes: If we do want to serialize artifacts, let's at least put them in JSON format rather than a proprietary binary format.

nicolaskruchten: my thought was that we would have both TF-IDF and K-Means serialized to CSVs-with-headers.

nicolaskruchten:
So the current proposal still outputs the datasets for introspection, it just doesn't use them to define the function. Having the function created by the procedure is fine when they are running on the same machine but the requirement here is that it should be very easy to train on one machine and score on another.

At the moment, we can train on machine X, then we have to run a CSV export procedure on machine X to create an artifact on S3, then we have to run a CSV import procedure on machine Y using the S3 URI and then we have to create a function on machine Y and we still need to remember to carry the appropriate metadata along when creating the function. Compare this with the classifier or svd or statstable procedures: train on machine X to create artifact on S3, create a function on machine Y using S3 URI..

jeremybarnes:
OK.  That works for introspection.  But if I want to run a k-means or TF-IDF with modified parameters I still can't (though JSON or especially CSV would help with that).

My thoughts on model portability distribution are: if we select a universal location for the file, and we allow implicit dataset loading, then we could use the following in the function config:

```
params: {
    modelDatasetThatContainsParametersForFunction: {
        type: "csv.import",
        dataFileUri: "s3://model.file.lives.here"
   }
}
````

or even modelDataset: "SELECT * FROM load_dataset({ id: 'xxx', type: 'yyy', params: { modelFileId: 's3://xxx' } } )"

I think that we should aim

nicolaskruchten:
I don't disagree that we can do better than just persisting to files, but I think that in the short run, standardization is better than this mixed mess we have right now, so we should move forward with this change, and schedule a bigger move towards another model at some later date.

Also, FWIW, I would rather that we run discussions about the DOD (the 'what') in the JIRA issue (i.e. https://datacratic.atlassian.net/browse/MLDB-1226) and keep the implementation details (the 'how') in the Github PR wherever possible :)

nicolaskruchten:
+1

PR #57, opened 2016-01-11
Don't stack overflow on JSON encoding long strings (MLDB-1238)

jeremybarnes:
@wsourdeau FYI.  The MLDB versions are a little different and we took the opportunity to avoid making a character by character copy of the string passed in if it doesn't need escaping.

### DIFF
--- a/types/json_printing.cc
+++ b/types/json_printing.cc
@@ -1,8 +1,8 @@
-// This file is part of MLDB. Copyright 2015 Datacratic. All rights reserved.
-
 /* json_printing.cc
    Jeremy Barnes, 8 March 2013
    Copyright (c) 2013 Datacratic Inc.  All rights reserved.
+
+   This file is part of MLDB. Copyright 2015 Datacratic. All rights reserved.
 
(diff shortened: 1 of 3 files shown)


types/json_printing.cc
```diff
@@ -19,17 +19,31 @@ using namespace std;
 
 namespace Datacratic {
 
+namespace {
+
+static constexpr char * BUFFER_TOO_SMALL = (char *)0;
```
wsourdeau: (cosmetic) "static" is not required in anonymous namespace.

types/json_printing.cc
```diff
@@ -19,17 +19,31 @@ using namespace std;
 
 namespace Datacratic {
 
+namespace {
```
wsourdeau: mmm. Anonymous namespace embedded within the "datacratic" namespace. Typo?

wsourdeau:
+1

types/json_printing.cc
jeremybarnes: No... it's file scope still.

PR #61, opened 2016-01-12
[MLDB-1111] added a check for limit

guyd:


plugins/for_each_line.cc
```diff
@@ -238,17 +238,17 @@ forEachLineStr(const std::string & filename,
 /*****************************************************************************/
 
 void forEachLineBlock(std::istream & stream,
-                      int64_t lineOffset,
```
jeremybarnes: We shouldn't need to modify the interface here.  The bool which is returned from the function can be used to skip.

guyd: Sure, I can do that.  However, all the other functions in for_each_line have this extra parameter and this function `forEachLineBlock` was also defining a `maxLines` variable internally (see below).

jeremybarnes:
+1

PR #62, opened 2016-01-12
MLDB-1244 Add 'finish' and logging script to mldb service

jraby:
The docker init system (based on runsvdir) will call
/etc/service/mldb_runner/finish when mldb_runner exits.
The finish script will print a message with the signal number that killed the
process if it is available.

The 'run' script has been tweaked to 'exec' mldb_runner instead of forking it.
This is required for the finish script to properly get the signal code when
mldb_runner exits.

Since we now 'exec', we cannot use a pipeline to redirect stderr to stdout and
stdout to tee.  Instead, we use a logging script, which is a simple wrapper
around tee. runsv will arrange to wire up the file descriptor correctly
(mldb_runner's stdout to the logger's stdin).

### DIFF
--- a/container_files/init/mldb_finish.py
+++ b/container_files/init/mldb_finish.py
@@ -14,13 +14,30 @@
 import os
 import sys
 
+sigmap = { 4: "SIGILL: illegal instruction - internal error",
+           6: "SIGABRT: abort(3) called - internal error",
+           9: "SIGKILL: killed from outside, did we run out of memory (OOM killed?)",
+           11: "SIGSEGV: segfault - internal error",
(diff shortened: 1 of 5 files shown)


jraby:
Example output:

```
MLDB ready




MLDB exited due to signal 15. SIGTERM: regular shutdown


[...]



MLDB ready




MLDB exited due to signal 9. SIGKILL: killed from outside, did we run out of memory (OOM killed?)

[...]

MLDB exited due to signal 11. SIGSEGV: segfault - internal error


[...]

MLDB exited due to signal 1.
```

nicolaskruchten:
+1, merging!

PR #63, opened 2016-01-13
ML::Thread_Specific -> thread_local (MLDB-1249)

jeremybarnes:
Since the current compilers aren't horrendously slow for thread_local, we can use it instead of the workaround.


wsourdeau:
+1

jeremybarnes:
@wsourdeau I deepened the changes; please re-check.

jeremybarnes:
@jraby FYI once this patch lands, we'll no longer need boost-thread or boost-thread-dev to build or run MLDB.  We should update documentation and the Docker build.

jeremybarnes:
Another spurious test failure...

PR #64, opened 2016-01-13
Mldb 1242 sampled dataset

mailletf:


builtin/sampled_dataset.cc
```diff
@@ -0,0 +1,368 @@
...
+/*****************************************************************************/
+
+DEFINE_STRUCTURE_DESCRIPTION(SampledDatasetConfig);
+
+SampledDatasetConfigDescription::
+SampledDatasetConfigDescription()
+{
+    nullAccepted = true;
+
+    addField("rows", &SampledDatasetConfig::rows,
```
guyd: I suggest we use _size_ instead of _rows_.

builtin/sampled_dataset.cc
```diff
@@ -0,0 +1,368 @@
...
+            "Number of rows to sample. It cannot be set when fraction is "
+            "set.");
+    addField("fraction", &SampledDatasetConfig::fraction,
+            "Fraction of rows in dataset to sample. It cannot be set when "
+            "rows is set.");
+    addField("withReplacement", &SampledDatasetConfig::withReplacement,
+            "Sample with or without replacement", false);
```
guyd: This would require more documentation.   It was not clear to me what this option was controlling.

builtin/sampled_dataset.h
```diff
@@ -0,0 +1,66 @@
...
+
+/*****************************************************************************/
+/* SAMPLED DATASET CONFIG                                                    */
+/*****************************************************************************/
+        
+struct SampledDatasetConfig {
+
+    SampledDatasetConfig();
+
+    PolyConfigT<const Dataset> dataset;
```
guyd: I would put a std::shared_ptr<TableExpression>.  This way one can simply put a string in the config instead of a json object.

builtin/sampled_dataset.cc
```diff
@@ -0,0 +1,368 @@
...
+            const SampledDatasetConfig config)
+        : dataset(dataset),
+          matrix(dataset->getMatrixView()),
+          index(dataset->getColumnIndex()),
+          columnCount(matrix->getColumnNames().size())
+    {
+        // get all existing  rows
+        auto rows = matrix->getRowHashes();
+
+        if (config.rows != 0 && config.fraction != 0) {
```
guyd: This validation should go in the onPostValide of the config for consistency.

builtin/sampled_dataset.cc
```diff
@@ -0,0 +1,368 @@
...
+            "Number of rows to sample. It cannot be set when fraction is "
+            "set.");
+    addField("fraction", &SampledDatasetConfig::fraction,
+            "Fraction of rows in dataset to sample. It cannot be set when "
+            "rows is set.");
+    addField("withReplacement", &SampledDatasetConfig::withReplacement,
+            "Sample with or without replacement", false);
+    addField("dataset", &SampledDatasetConfig::dataset,
+             "Dataset to sample");
+    addField("seed", &SampledDatasetConfig::seed,
```
guyd: This is optional, we should mention it.

builtin/sampled_dataset.cc
nicolaskruchten: I think rows makes more sense: we are sampling either n rows OR f fraction (see next option)

builtin/sampled_dataset.cc
```diff
@@ -0,0 +1,368 @@
...
+             "Seed value for the random number generator");
+}
+
+
+/*****************************************************************************/
+/* SAMPLED INTERNAL REPRESENTATION                                        */
+/*****************************************************************************/
+
+
+struct SampledDataset::Itl
```
guyd: The FilteredDataset class was attempting a similar wrapping.  Not sure if it would help deriving from it.

builtin/sampled_dataset.cc
nicolaskruchten: We should the following to the `withReplacement` option: "Sampling with replacement means that the same input row can appear in the output more than once." 

We should also add the following to the `rows` option: "This option cannot be set higher than the number of rows of input unless withReplacement is set to true."

builtin/sampled_dataset.cc
```diff
@@ -0,0 +1,368 @@
...
+
+            // if we're not sampling with replacement, check if
+            // we already sampled this index
+            if(!config.withReplacement) {
+                sampledIt = sampledIndexes.find(sample_index);
+                if(sampledIt != sampledIndexes.end())
+                    continue;
+            }
+
+            sampledIndexes.insert(sample_index);
```
guyd: Shouldn't this be in a else { }?

builtin/sampled_dataset.cc
```diff
@@ -0,0 +1,368 @@
...
+
+            auto rowName = matrix->getRowName(rows[sample_index]);
+            sampledRows.emplace_back(rowName);
+            sampledRowsIndex.insert(rowName);
+        }
+    }
+    
+    virtual RowName getRowName(const RowHash & row) const 
+    {
+        return matrix->getRowName(row);
```
guyd: Shouldn't we return that it is an unknown row when it is not in the sample?

builtin/sampled_dataset.cc
nicolaskruchten: We should add: "The purpose of this parameter is to permit reproducible random samples. This parameter is optional, with the default value being selected randomly for each sample. "

builtin/sampled_dataset.cc
```diff
@@ -0,0 +1,368 @@
...
+    }
+
+    virtual size_t getRowCount() const
+    {
+        return sampledRows.size();
+    }
+
+    virtual size_t getColumnCount() const
+    {
+        return matrix->getRowCount();
```
guyd: getRowCount -> getColumnCount

builtin/sampled_dataset.cc
```diff
@@ -0,0 +1,368 @@
...
+    }
+
+    std::pair<Date, Date>
+    getTimestampRange() const
+    {
+        // TODO
```
guyd: We should but a tracking JIRA here.

builtin/sampled_dataset.cc
guyd: _Sample size_ being so common that I felt it was the right term to use.

builtin/sampled_dataset.cc
guyd: Should we rename it to `allowDuplicates`?

guyd:
SampledDataset are immutable right?  Can you add a test to ensure this is returning a meaningful error.

builtin/sampled_dataset.cc
nicolaskruchten: No, the term of art is in fact "sampling with replacement"

guyd:
+1

PR #65, opened 2016-01-13
MLDB-1247 logging process revamp

jraby:
Tornado based 'logger':

 - Reads lines from stdin
 - keeps a buffer of 1024 lines in memory
 - writes every lines to stdout (flush after every write)
 - The content of the buffer can be queried over HTTP at `127.0.0.1:HTTP_LISTEN_PORT/logs/mldb`
   the last 'n' logs lines can be fetched by passing `?n=number_of_lines` (ex: `127.0.0.1:HTTP_LISTEN_PORT/logs/mldb?n=32`)
 - Root privileges are dropped on start

There some more work to be done in there, namely regarding the logging of the http server which goes to stderr and adding a route to listen for log lines in realtime. But it is probably good enough to be merged.


container_files/assets/www/crash.html
```diff
@@ -3,15 +3,16 @@
...
           <h2>Well this is embarrasing... it looks like MLDB crashed!</h2>
 
           <p> The process currently restarting, please try again in a few moments.</p>
 
-          <p> More information about the crash can be found in the <a href="/logs/mldb" target="_blank">mldb log file</a>. </p>
-
           <p>The MLDB Team at Datacratic would love to help you debug this problem! Please feel free to reach out to us at <a href="mailto:mldb@datacratic.com" target="_blank">mldb@datacratic.com</a>.</p>
 
+          <p> More information about the crash can be found in the MLDB log: </p>
+          <iframe src="/logs/mldb" height="500" width="800" />
```
nicolaskruchten: missing the #end

nicolaskruchten:
mega-beau travail!

nicolaskruchten:
+1 once comment is addressed :)

jraby:
For the record, this thing was handling 160k lines/sec on a dumb benchmark while using ~20mb of ram.
Should be fast enough for our use.

PR #66, opened 2016-01-13
MLDB-1253 concat function + test

FinchPowers:


sql/builtin_functions.cc
```diff
@@ -1307,6 +1307,92 @@ RegisterVectorOp<SumOp> registerVectorSum("vector_sum");
...
+                        else {
+                            result += separator;
+                        }
+                        result += columnValue ?
+                            val.toUtf8String() : columnName.toUtf8String();
+                    }
+                    return true;
+                };
+
+                args.at(0).forEachAtom(onAtom);
```
FinchPowers: I need assistance. This does not iterate over null values but I need it to. How do I do?

mailletf:
Missing the new test file

testing/MLDB-1253_concat_test.py
```diff
@@ -0,0 +1,102 @@
...
+            log(test_case.__name__ + ' SUCCESS')
+            success += 1
+        except Exception:
+            log(test_case.__name__ + ' FAILED')
+            errors += 1
+            log(traceback.format_exc())
+
+    log("{}/{} test(s) passed".format(success, success + errors))
+    if errors:
+        raise Exception("{} test(s) failed".format(errors))
```
FinchPowers: @nicolaskruchten From the top of the file to here, these are functions that I reuse in most of my codes. I am considering creating an utility library. As for the test_case + run_tests functions, I tried to import unittest but it doesn't work because sys.argv is undefined. I may be worth it to spend more time to make it work.

testing/MLDB-1253_concat_test.py
```diff
@@ -0,0 +1,102 @@
...
+
+@test_case
+def test_default():
+    res = query('SELECT concat({*}) FROM sample')
+    assert res[1][1] == 'val1A,val1B'
+    assert res[2][1] == 'val2A,val2C'
+
+
+@test_case
+def test_keep_nulls():
```
FinchPowers: Currently failing as I can't get the null values in the function.

testing/MLDB-1253_concat_test.py
nicolaskruchten: I wouldn't put too much effort into a utility library for now. We will be upgrading the Python API soon to add `mldb.query()` and to have non-2xx responses raise exceptions, which will require us to revisit a whole pile of tests anyway. Let's take a look once this is done to see what's missing.

sql/builtin_functions.cc
guyd: Je ne suis pas sûr de comprendre le problème.  Le code de `forEachAtom` semble appeler `onAtom` pour les nulls aussi.

sql/builtin_functions.cc
```diff
@@ -1307,6 +1307,92 @@ RegisterVectorOp<SumOp> registerVectorSum("vector_sum");
...
+        else if (columnName == ColumnName("separator")) {
+            assertArg(1, "separator");
+            separator = std::get<1>(arg).toUtf8String();
+        }
+        else if (columnName == ColumnName("columnValue")) {
+            assertArg(2, "columnValue");
+            columnValue = std::get<1>(arg).asBool();
+        }
+        else {
+            throw HttpReturnException(400, "Unknown argument in tokenize",
```
guyd: tokenize -> concat

sql/builtin_functions.cc
```diff
@@ -1307,6 +1307,92 @@ RegisterVectorOp<SumOp> registerVectorSum("vector_sum");
...
+            throw HttpReturnException(
+                400, "Argument " + name + " is specified more than once");
+        }
+        check[field] = true;
+    };
+
+    for (auto& arg : argRow) {
+        const ColumnName& columnName = std::get<0>(arg);
+        if (columnName == ColumnName("skipNulls")) {
+            assertArg(0, "splitNulls");
```
guyd: splitNulls -> skipNulls

sql/builtin_functions.cc
```diff
@@ -1307,6 +1307,92 @@ RegisterVectorOp<SumOp> registerVectorSum("vector_sum");
...
+BoundFunction concat(const std::vector<BoundSqlExpression> & args)
+{
+    return {[=] (const std::vector<ExpressionValue> & args,
+                 const SqlRowScope & context) -> ExpressionValue
+            {
+                bool skipNulls = true;
+                Utf8String separator(",");
+                bool columnValue = true;
+
+                if (args.size() == 0) {
```
guyd: The two first check can be done at bind time, that is, above the return.

sql/builtin_functions.cc
```diff
@@ -1307,6 +1307,92 @@ RegisterVectorOp<SumOp> registerVectorSum("vector_sum");
...
+                    throw HttpReturnException(
+                        400, "concat requires at least one argument");
+                }
+
+                if (args.size() > 2) {
+                    throw HttpReturnException(
+                        400, "concat requires at most two arguments");
+                }
+
+                if (args.size() == 2) {
```
guyd: Actually, this one too can be handled at bind time.  This would avoid reparsing that static input for all rows in the dataset.

sql/builtin_functions.cc
```diff
@@ -1307,6 +1307,92 @@ RegisterVectorOp<SumOp> registerVectorSum("vector_sum");
...
+                    ParseConcatArguments(skipNulls, separator, columnValue,
+                                         args.at(1).getRow());
+                }
+
+                Utf8String result = "";
```
guyd: Not sure how much we would save on reallocation if we reserve some space in the string first.  Finding the right size to reserve is the tough part!

testing/MLDB-1253_concat_test.py
guyd: Ah!  When you say _null values_ you mean _missing values_!  There are places where we access the _known columns_ but I don't think this is accessible in ExpressionValue.

guyd:
Reviewed with a few minor comments.  Regarding the missing columns problem, I am not sure how to tackle it.

sql/builtin_functions.cc
FinchPowers: This is somewhat crap. Do we have a utf8stringstream?

guyd: Not that I know and if we have I would doubt it to be faster than a Utf8String.

FinchPowers: Then can I deduce that Utf8String is closer to a stringstream than std::string?

sql/builtin_functions.cc
```diff
@@ -1307,6 +1307,93 @@ RegisterVectorOp<SumOp> registerVectorSum("vector_sum");
...
+                 const SqlRowScope & context) -> ExpressionValue
+        {
+            bool skipNulls = true;
+            Utf8String separator(",");
+            bool columnValue = true;
+
+            if (args.size() == 2) {
+                ParseConcatArguments(skipNulls, separator, columnValue,
+                                        args.at(1).getRow());
+            }
```
FinchPowers: @guyd Si j'extrait cette fonction à l'extérieur, j'ai
```
./mldb/sql/builtin_functions.cc:1362:41: error: ‘const value_type’ has no member named ‘getRow’
                              args.at(1).getRow());
```

guyd: Tu dois exécuter le _BoundSqlExpression_ avant.  

```
args.at(1)(context).getRow()
```

sql/builtin_functions.cc
guyd: Not sure what you mean.  My point is that neither Utf8String nor stringstream will fix the problem of the multiple reallocations as the string grows.  I cannot think of a simple solution right now so let's keep this in mind as revisit when we will have a performance issue.

testing/MLDB-1253_concat_test.py
```diff
@@ -0,0 +1,141 @@
...
+    assert res['statusCode'] in [200, 201], str(res)
+    return res
+
+
+def query(query):
+    res = perform('GET', '/v1/query', [['q', query], ['format', 'table']])
+    return json.loads(res['response'])
+
+
+def record_row(ds_url, row_name, cols):
```
mailletf: Did you see that we have a dataset object wrapped with methods to addRows and commit? example: https://github.com/mldbai/mldb/blob/master/testing/MLDB-878_experiment_proc.py#L8

sql/builtin_functions.cc
FinchPowers: Mais j'ai pas le context alors.

guyd: Ah.  En effet.  Regarde comment c'est fait pour les regex.
https://github.com/mldbai/mldb/blob/master/sql/builtin_functions.cc#L390
Encore une fois c'est une optimization sans trop connaitre l'impact sur les performances.  Donc si ça ne passe pas avec cette approche je ne passerai pas beaucoup de cycles supplémentaire sur ce changement.

guyd:
+1

FinchPowers:
Squash merged.

PR #67, opened 2016-01-13
Sets python sys.argv, allows unittest to work

FinchPowers:


plugins/lang/python/python_plugin_context.cc
```diff
@@ -199,6 +199,8 @@ class CatchOutErr:
         self.method=method
     def write(self, txt):
         self.catchOut.write(txt, self.method)
+    def flush(self):
```
nicolaskruchten: whoa, we can inject Python code in the context of every execution??? if so, we can bolt on all sorts of useful stuff to the mldb object right now :) /cc @mailletf ?

jeremybarnes:
Looks like there is an infinite loop; the tests appear to have hung.

plugins/lang/python/python_loader.cc
```diff
@@ -443,9 +443,13 @@ runPythonScript(std::shared_ptr<PythonContext> titl,
     // if we're simply executing the body of the script
     try {
         if(elementToRun == PackageElement::MAIN) {
-                JML_TRACE_EXCEPTIONS(false);
-                boost::python::object obj
-                    = boost::python::exec(scriptSourceStr, pyControl.main_namespace);
+            JML_TRACE_EXCEPTIONS(false);
+            int argc = 1;
```
jeremybarnes: this should be sizeof(argv1) / sizeof(char *), and you can do const char * argv[] = { "arg1", "arg2" }.  That is clearer and less error prone.

FinchPowers: Thanks for that. I coded it when all c++ kings were gone and I had the feeling it was wrong. ;)

FinchPowers: @jeremybarnes
```
const char * argv[] = { "mldb-boost-python"};
```
doesn't work because
1. error: invalid conversion from ‘const char**’ to ‘char**’ [-fpermissive] PySys_SetArgv(argc, argv);
2. error: deprecated conversion from string constant to ‘char*’

FinchPowers: Are you sure argc shouldn't be equal to 1? It is supposed to be the count of strings pointed to by argv. Argv holds one string hence argc should be 1 no?

FinchPowers:
MLDB-951-run-on-creation.py crashes with that patch. I'm trying to see why.

plugins/lang/python/python_plugin_context.cc
mailletf: Yup

plugins/lang/python/python_loader.cc
```diff
@@ -443,9 +443,13 @@ runPythonScript(std::shared_ptr<PythonContext> titl,
...
     try {
         if(elementToRun == PackageElement::MAIN) {
-                JML_TRACE_EXCEPTIONS(false);
-                boost::python::object obj
-                    = boost::python::exec(scriptSourceStr, pyControl.main_namespace);
+            JML_TRACE_EXCEPTIONS(false);
+            char argv1[] = "mldb-boost-python";
+            char *argv[] = {argv1};
+            int argc = sizeof(argv[0]) / sizeof(char *);
+            PySys_SetArgv(argc, argv);
```
mailletf: Would we need the same PySys_SetArgv call for the python::exec call at line 471?

FinchPowers: Dis-moi comment déclencher ce code et je vais vérifier.

mailletf: Faut faire un plugin avec des routes custom. Check ici http://mldb.ai/doc/#builtin/lang/Python.md.html dans la struct PackageElementSources il y a l'élément "routes". Donc mettre du code dans routes, loader le plugin, caller une route custom.

mailletf:
+1

PR #69, opened 2016-01-14
Fix a segfault in rest_collection

FinchPowers:
It occurrs when the object is destroyed before threads have finished
running. I added a counter and a condition variable to have the dtor
wait for them before dtoring.

### DIFF
--- a/rest/rest_collection_impl.h
+++ b/rest/rest_collection_impl.h
@@ -1,9 +1,8 @@
-// This file is part of MLDB. Copyright 2015 Datacratic. All rights reserved.
-
 /** rest_collection_impl.h                                         -*- C++ -*-
     Jeremy Barnes, 21 January 2014
     Copyright (c) 2014 Datacratic Inc.  All rights reserved.
 
+    This file is part of MLDB. Copyright 2015 Datacratic. All rights reserved.
 */
 
 #pragma once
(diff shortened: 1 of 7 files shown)


rest/rest_collection_impl.h
```diff
@@ -82,6 +82,10 @@ struct RestCollection<Key, Value>::Impl {
         protected by the RCU lock.
     */
     MutateMutex mutateMutex; // not mutable on purpose
+
+    int runningThreads;
```
FinchPowers: I added a counter, but my knowledge of that class is insufficient to know whether there can be more than one or not.

rest/rest_collection_impl.h
```diff
@@ -143,6 +147,13 @@ void
 RestCollection<Key, Value>::
 clear()
 {
+    {
+        std::unique_lock<std::mutex> lock(impl->runningThreadsMutex);
+        while (impl->runningThreads != 0) {
+            std::cerr << "running threads: " << impl->runningThreads << std::endl;
```
FinchPowers: Debug output. Can be removed.

FinchPowers:
One bug with that fix: If the thread crashes before the counter can be decremented, the program will hang there.

guyd:
We are getting closer to a fix.
Deferring to Jeremy to confirm that such an approach can work.

jeremybarnes:
Bravo for the diagnosis @FinchPowers .  It doesn't entirely fix the problem (I have a testcase that still triggers an error, although less frequently with your patch).  I'm working on a better fix.

jeremybarnes:
@FinchPowers FYI.

guyd:
Looking forward to see that merged into master.

PR #70, opened 2016-01-14
MLDB-1250 fix bad downcast

steadtler:


sql/binding_contexts.cc
```diff
@@ -66,13 +66,14 @@ doGetFunction(const Utf8String & tableName,
...
-            //<< ML::type_name(context) << " outer type is "
-            //<< ML::type_name(row.outer) << endl;
-
-            return outerFunction(args, row.outer);
+            const RowContext * row = dynamic_cast<const RowContext*>(&context);
+            if (row) {
+                return outerFunction(args, row->outer);
+            }
+            else {
+                return outerFunction(args, context);
```
guyd: Peut-être expliquer dans quel cas le contexte ne sera pas du bon type?

guyd:
+1

jeremybarnes:
At some point in the past, the dynamic_cast calls were contributing a major part of the runtime, and so I had modified things so that RowContext was not an abstract class (had no VMT).  I added it back only so that we could check for debugging.  In what circumstance could it not be a row context?  It smells bad to me but it's probably because I don't understand something.  -1 until we figure out where the misunderstanding is.

steadtler:
Alright, I think I got the underlying cause. I still dislike unwrapped static_casts though, I hope we can find a way around it eventually.

jeremybarnes:
I'd prefer that we include an int cookie in the underlying class that we can test against a magic value or something like that, to validate the static cast.  It's a pretty major speed increase, so probably worth paying a small bad smell for.

steadtler:
So the main problem I found was that FunctionExpressionContext inherited from ReadThroughBindingContext without its RowContext following suit. Its been a bit of a pain passing the outer context around...

sql/binding_contexts.cc
```diff
@@ -66,12 +66,7 @@ doGetFunction(const Utf8String & tableName,
     result.exec = [=] (const std::vector<ExpressionValue> & args,
                        const SqlRowScope & context)
         {
-            auto & row = static_cast<const RowContext &>(context);
-            //cerr << "rebinding to apply function " << functionName
-            //<< ": context type is "
-            //<< ML::type_name(context) << " outer type is "
-            //<< ML::type_name(row.outer) << endl;
-
+             const RowContext & row = dynamic_cast<const RowContext&>(context);
```
guyd: The dynamic cast goes away right?  Also the commented cerr turned out to be useful for me a couple of time.  I would leave it.

server/function_contexts.h
```diff
@@ -66,9 +66,9 @@ struct ExtractContext: public SqlBindingScope {
 
 struct FunctionExpressionContext: public ReadThroughBindingContext {
 
-    struct RowContext: public SqlRowScope {
-        RowContext(const FunctionContext & input)
-            : input(input)
+    struct RowContext: public ReadThroughBindingContext::RowContext {
```
guyd: This might be the problem I am seeing with my changes that segfault on some function calls.

sql/binding_contexts.cc
steadtler: Oups my mistake

core/function.cc
```diff
@@ -684,10 +684,10 @@ toRowInfo() const
 
 FunctionOutput
 FunctionApplier::
-apply(const FunctionContext & input) const
-{
+apply(const SqlRowScope & outer, const FunctionContext & input) const
```
jeremybarnes: I would add some documentation to outer parameter, describing exactly why it's needed and what should be passed in.

guyd:
Reviewed the added detection of stack overflow.  +1

PR #71, opened 2016-01-14
Update services to use 'exec' and drop tee -a log

jraby:
Allows all services to be properly controlled by sv {status,start,stop} service

The in-container log file for all services will not be generated anymore.
If we ever need those, we can use a 'log' script like mldb_runner does.


nicolaskruchten:
+1

PR #72, opened 2016-01-14
HTTP 5xx handling update

jraby:
- crash.html -> 5xx.html
  - Send an application/json response for 5xx that happen on /v1/* routes

@nicolaskruchten this is a POC, feel free to update the json payload.


nicolaskruchten:
+1. I will review the exact wording for 5xx.* before the next release

PR #74, opened 2016-01-14
Mldb 1258 nofrom segfault

mailletf:


guyd:
+1 once the failing tests are passing

PR #75, opened 2016-01-18
MLDB-1256 fix typo in ambiguous aggregator state

steadtler:


testing/MLDB-703-count-not-null.js
```diff
@@ -29,12 +29,14 @@ var expected = [
 ];
 assertEqual(resp.json, expected)
 
-var resp1 = mldb.get('/v1/query', { q: 'select x, count(x) as a from test1 group by x', format: 'table' });
+//  MLDB-1256
+var resp1 = mldb.get('/v1/query', { q: 'select x, count(x) as a from test1 where x is not null group by x', format: 'table' });
```
steadtler: I had to add a "is not null" to both queries because count(*) skips null values

guyd:
+1

testing/MLDB-703-count-not-null.js
FinchPowers: If count(*) skips nulls, adding "is not null" is a synonym and should make no difference.

PR #79, opened 2016-01-19
MLDB-1267 bucketize latest timestamp

jeremybarnes:
Causes the latest timestamp for the order by column to be used by the bucketize procedure.

Also fixes a bug in CSV timestamps that caused the test to fail originally.

@FinchPowers FYI

### DIFF
--- a/plugins/bucketize_procedure.cc
+++ b/plugins/bucketize_procedure.cc
@@ -122,10 +122,27 @@ run(const ProcedureRunConfig & run,
     SelectExpression select(SelectExpression::parse("1"));
     vector<shared_ptr<SqlExpression> > calc;
 
+    // We calculate an expression with the timestamp of the order by
+    // clause.  First, we need to calculate each of the order by clauses
+    for (auto & c: procedureConfig.inputData.stm->orderBy.clauses) {
+        auto whenClause = std::make_shared<FunctionCallWrapper>
+            ("", "when", vector<shared_ptr<SqlExpression> >(1, c.first),
+             nullptr /* extract */);
+        calc.emplace_back(whenClause);
+    }
+
     vector<Id> orderedRowNames;
+    Date globalMaxOrderByTimestamp = Date::negativeInfinity();
     auto getSize = [&] (const MatrixNamedRow & row,
                         const vector<ExpressionValue> & calc)
     {
+        for (auto & c: calc) {
+            auto ts = c.getAtom().toTimestamp();
+            if (ts.isADate()) {
+                globalMaxOrderByTimestamp.setMax(c.getAtom().toTimestamp());
+            }
+        }
+
         orderedRowNames.emplace_back(row.rowName);
         return true;
     };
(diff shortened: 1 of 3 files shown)


plugins/bucketize_procedure.cc
```diff
@@ -122,10 +122,28 @@ run(const ProcedureRunConfig & run,
     SelectExpression select(SelectExpression::parse("1"));
     vector<shared_ptr<SqlExpression> > calc;
 
+    // We calculate an expression with the timestamp of the order by
+    // clause.  First, we need to calculate each of the order by clauses
+    for (auto & c: procedureConfig.inputData.stm->orderBy.clauses) {
+        cerr << "when clause for " << c.first->print() << endl;
```
steadtler: debug output?

jeremybarnes: thanks, fixed

steadtler:
Looks fine besides the debug leftover + 1

testing/MLDB-1043-bucketize-procedure.js
```diff
@@ -62,6 +62,11 @@ function runTest(testNumber, buckets, goldStandard) {
     assertEqual(res['responseCode'], 200);
     checkResults(res, goldStandard);
 
+    res = mldb.get("/v1/query", {q: "SELECT *, when({*}) FROM " + datasetName,
+                                 format: "full"});
+
+    // MLDB-1267
+    assertEqual(res.json[0].columns[1][1].ts, "2015-12-22T14:26:51Z");
```
guyd: Wouldn't this be dependent on when the check-out was made?

plugins/bucketize_procedure.cc
```diff
@@ -143,26 +161,29 @@ run(const ProcedureRunConfig & run,
...
                                 nullptr, true /*overwrite*/);
 
-    typedef tuple<ColumnName, CellValue, Date> cell;
-    PerThreadAccumulator<vector<pair<RowName, vector<cell>>>> accum;
+    typedef tuple<ColumnName, CellValue, Date> Cell;
+    PerThreadAccumulator<vector<pair<RowName, vector<Cell> > > > accum;
 
     for (const auto & mappedRange: procedureConfig.percentileBuckets) {
-        string bucketName(mappedRange.first);
+        ColumnName bucketColumn("bucket");
```
FinchPowers: This variable is unused.

plugins/bucketize_procedure.cc
```diff
@@ -177,14 +197,14 @@ run(const ProcedureRunConfig & run,
...
-             << " to " << higherBound << endl;
+        //cerr << "Bucket " << mappedRange.first << " from " << lowerBound
+        //     << " to " << higherBound << endl;
 
         ML::run_in_parallel_blocked(lowerBound, higherBound, applyFct);
     }
 
     // record remainder
-    accum.forEach([&] (vector<pair<RowName, vector<cell>>> * rows)
+    accum.forEach([&] (vector<pair<RowName, vector<Cell> > > * rows)
```
FinchPowers: Why the extra whitespaces? Is it auto formatting? I think old c++ compilers required it, is that the reason?

testing/MLDB-1043-bucketize-procedure.js
FinchPowers: This is indeed a bug.

PR #82, opened 2016-01-20
MLDB-1273 flattening nested rows

jeremybarnes:
Fixes column names coming out as row..a instead of row.a when flattening.


guyd:
+1

plugins/json_importer.cc
```diff
@@ -167,7 +167,7 @@ struct JSONImporter: public Procedure {
                             for(int i=0; i<val.size(); i++) {
                                 string key;
                                 if(val[i].isString())      key = val[i].asString();
-                                else if(val[i].isBool())   key = val[i].asBool() ? "true" : false;
+                                else if(val[i].isBool())   key = val[i].asBool() ? "true" : "false";
```
mailletf: nice catch

PR #84, opened 2016-01-21
Mldb 1111 csv tabular limit

guyd:
I made these changes related to limit:
- empty lines are handled like error lines and contribute to the limit count
- fixed a bug in forEachLineBlock where new block reading were scheduled even when the limit was reached (this actually made query with limit slower than those without limit)
- stop early the reading of memory mapped file when the limit is reached


steadtler:
+1

plugins/csv_dataset.cc
```diff
@@ -897,7 +908,7 @@ struct CsvDataset::Itl: public TabularDataStore {
                 //threadAccum.emplace_back(std::move(lineEntry));
             };
 
-        forEachLineBlock(stream, onLine, lineOffset, config.limit);
```
jeremybarnes: If we don't pass lineOffset, how do we a) print the correct line numbers in error messages, and b) deal with the fact that the header is actually a line?  In other words, the for_each_line function was designed to care about *lines* only, and so we need to tell it how many lines we have already skipped when we start so it can tell us the appropriate error message.

testing/MLDB-1155_csv_line_endings.py
```diff
@@ -199,7 +197,7 @@
     assert json.loads(result['response'])[1][2] == 1
 
 
-    # CSV importation should accept a blank trailing line with DOS
+    # CSV importation should accept a blank trailing line with DOS if ignoreBadLines is set to true
```
jeremybarnes: It should accept a blank trailing line no matter what ignoreBadLines is set to; it's too common for us tom consider that a bad line.

jeremybarnes:
We need test cases that check that the line number reported in errors is the actual physical line number in the file, both in the presence and absence of headers.  Once we have that and they pass, I'm a +1.

plugins/csv_dataset.cc
guyd: forEachLineBlock does not return error messages.  The caller is responsible to know how many lines were skipped (because of header, offset) and return the proper line number in error messages (see actualLineNum).  I changed it because it was too misleading since the code do not actually skip any lines.

testing/MLDB-1155_csv_line_endings.py
guyd: This is not easily fixed.  The callers of forEachLineBlock have no idea if a line is the last one and forEachLineBlock should not drop lines.  Not sure what to do here.

guyd:
Agreed for the test case.  I've verified that but forgot to cast it into a test case.

guyd:
Changed the code of forEachLineBlock to drop the last line if it is empty.  In order word, a CSV with *one* empty line at the end will not cause an error no matter how _ignoreBadLines_ is set.

PR #86, opened 2016-01-21
Limit and Group By queries memory and cpu optimization

steadtler:
The goal of this change is to avoid creating an array of all the dataset's rowNames when the number of rows returned will be small (group queries and queries with limit). This is done by adding an interface to the dataset to stream rowNames one by one. The interface is optional, the only dataset that does not support it right now is the SQLite dataset. Sorted queries not supported yet. Each working thread has its own copy of the stream.

The streams could be made to be more efficient, most of it is a naive implementation for now.

This should also make it easier to multithread the pipeline execution model.

Queries that use this codepath should use about # of row * sizeof(rowName) less memory. According to my tests, group by queries are also a couple of time fasters, while limit 1 queries are almost 2 order of magnitude faster.


server/bound_queries.cc
```diff
@@ -625,6 +722,150 @@ struct RowHashOrderedExecutor: public BoundSelectQuery::Executor {
...
+            MatrixNamedRow row = std::move(matrix->getRow(r));
+            auto rowContext = context.getRowContext(row);
+
+            whenBound.filterInPlace(row, rowContext);
+            NamedRowValue outputRow;
+            outputRow.rowName = row.rowName;
+            outputRow.rowHash = row.rowName;
+
+            vector<ExpressionValue> calcd(boundCalc.size());
+            for (unsigned i = 0;  i < boundCalc.size();  ++i) {
```
FinchPowers: Purement sur le style, mais y'a une dualité au niveau des brackets entre ici et la ligne 832.

FinchPowers:
+1
J'ai hâte de voir les graphiques dans Kibana demain.

builtin/joined_dataset.cc
```diff
@@ -59,6 +59,30 @@ struct JoinedDataset::Itl
...
+            
+        }
+
+        virtual std::shared_ptr<RowStream> clone() const{
+            auto ptr = std::make_shared<JoinedRowStream>(source);
+            return ptr;
+        }
+
+        virtual void initAt(size_t start){
+            iter = source->rows.begin() + start;
```
mailletf: Ça peut pas être out of bounds ça?

builtin/joined_dataset.cc
```diff
@@ -59,6 +59,30 @@ struct JoinedDataset::Itl
...
+        }
+
+        virtual RowName next() {
+            return (iter++)->rowName;
```
mailletf: out of bounds?

builtin/joined_dataset.cc
mailletf: Per discussion with Mathieu, on purpouse. Responsbility passed on to the caller to no go out of bounds. Will be documented in the code

steadtler: I purposefully dont check for outofbounds for performance reasons. The existing interface for datasets means that we always have an exact Row Count, so the executor knows exactly how many times it should call next on the stream. I'll add a comment to specify that.

plugins/sparse_matrix_dataset.cc
```diff
@@ -473,11 +509,11 @@ struct SparseMatrixDataset::Itl
             return result;
         }
 
+        result.erase(result.begin(), result.begin() + start);
```
mailletf: nice catch

builtin/id_hash.h
```diff
@@ -203,6 +203,53 @@ struct IdHashes {
...
+                if (bucket == NBUCKETS)
+                    break;
+                else
+                    bucketIter = source->buckets[bucket].begin();
+            }
+        }
+        private:
+
+        const IdHashes* source;
+        unsigned bucket;
```
jeremybarnes: uint64_t or size_t?

builtin/id_hash.h
```diff
@@ -203,6 +203,53 @@ struct IdHashes {
...
+                bucket++;
+                if (bucket == NBUCKETS)
+                    break;
+                else
+                    bucketIter = source->buckets[bucket].begin();
+            }
+        }
+        private:
+
+        const IdHashes* source;
```
jeremybarnes: Here, we put spaces before the *.  If you want no spaces, go work for Google.

builtin/id_hash.h
```diff
@@ -203,6 +203,53 @@ struct IdHashes {
         }
         return true;
     }
+
+    struct const_iterator
```
jeremybarnes: I find boost::iterator_facade very helpful for getting this kind of iterator right.

builtin/joined_dataset.cc
```diff
@@ -59,6 +59,30 @@ struct JoinedDataset::Itl
...
+        virtual std::shared_ptr<RowStream> clone() const{
+            auto ptr = std::make_shared<JoinedRowStream>(source);
+            return ptr;
+        }
+
+        virtual void initAt(size_t start){
+            iter = source->rows.begin() + start;
+        }
+
+        virtual RowName next() {
```
jeremybarnes: Instead of next(), I would create next_n() which returns more than one in an array you pass in.  That will amortize the cost of the virtual method call, which could be quite significant here.

jeremybarnes: Also, some basic method comments would help.

builtin/joined_dataset.cc
```diff
@@ -59,6 +59,30 @@ struct JoinedDataset::Itl
...
+            return (iter++)->rowName;
+        }
+
+        std::vector<RowEntry>::const_iterator iter;
```
jeremybarnes: Private?

plugins/sparse_matrix.h
```diff
@@ -71,6 +71,16 @@ struct BaseEntry {
 struct MatrixWriteTransaction;
 
 struct MatrixReadTransaction {
+
+      struct Stream {
```
guyd: Why can we use interface RowStream here?  They seem identical.

builtin/joined_dataset.cc
```diff
@@ -694,6 +718,13 @@ getColumnIndex() const
     return itl;
 }
 
+std::shared_ptr<RowStream> 
+JoinedDataset::
+getRowStream() const
```
jeremybarnes: I think getRowGenerator is a better name.

builtin/joined_dataset.cc
jeremybarnes: Also, do we really need rowNames?  Or would it be better to allow the dataset to return some kind of row token, which could be used to avoid a by-name lookup when actually getting the row later on?

builtin/merged_dataset.cc
```diff
@@ -183,6 +183,38 @@ struct MergedDataset::Itl
...
+
+        virtual std::shared_ptr<RowStream> clone() const
+        {
+            return make_shared<MergedRowStream>(source);
+        }
+
+        /* set where the stream should start*/
+        virtual void initAt(size_t start)
+        {
+            it = source->rowIndex.begin();
```
jeremybarnes: Ugh, this is horrible.  HORRIBLE!  How often is it used?

builtin/joined_dataset.cc
steadtler: Yes, I thought about the next_n, but I wanted to wrap. Ok if I do it as a follow-up story?

core/dataset.h
```diff
@@ -173,6 +173,20 @@ struct ColumnIndex {
...
 
+/*****************************************************************************/
+/* ROW STREAM                                                                */
+/*****************************************************************************/
+
+struct RowStream {
+
+    virtual std::shared_ptr<RowStream> clone() const = 0;
+
+    /* set where the stream should start*/
```
jeremybarnes: I wonder: could we change this to an interface like splitIntoManyStreams(numStreams)?  To avoid having to implement initAt, which must be slow?

plugins/embedding.cc
```diff
@@ -286,6 +286,37 @@ struct EmbeddingDataset::Itl
...
     }
 
+    struct EmbeddingRowStream : public RowStream {
+
+        EmbeddingRowStream(const EmbeddingDataset::Itl* source) : index(0), source(source)
+        {
+         
+        }
+
+        virtual std::shared_ptr<RowStream> clone() const{
```
jeremybarnes: Why do we need to clone the row streams?  Let's at least make sure that we understand what the requirements are in a comment, eg where should the cloned stream point to?  The start?  The current element?

plugins/sparse_matrix_dataset.cc
jeremybarnes: oups

plugins/sparse_matrix.h
steadtler: They are similar. RowStream's next returns RowNames, this one returns the uint64_t hashes specific to the sparse datasets implementation

server/bound_queries.cc
```diff
@@ -188,6 +173,100 @@ struct UnorderedExecutor: public BoundSelectQuery::Executor {
...
+        // Do we select *?  In that case we can avoid a lot of copying
+        bool selectStar = boundSelect.expr->isIdentitySelect(context);
+
+        // Do we have where TRUE?  In that case we can avoid evaluating 
+        bool whereTrue = whereBound.expr->isConstantTrue();
+
+        size_t numPerBucket = (size_t)std::ceil((float)whereGenerator.upperBound / numBuckets);
+
+        int numRows = whereGenerator.upperBound;
+        auto doBucket = [&] (int bucketNumber) -> bool
```
jeremybarnes: Instead of initAt, we should ask the thing to create us N parallel streams and return them via a lambda, which we can then use to fire off to a thread.  In that case we can avoid a lot of iterating.

builtin/joined_dataset.cc
steadtler: Followup story for the token? Or as we said we could skip it entirely and pass the boundselect (or a lambda of sort) to the dataset instead.

server/bound_queries.cc
```diff
@@ -625,6 +722,150 @@ struct RowHashOrderedExecutor: public BoundSelectQuery::Executor {
         return;
     }
 
+    virtual void execute_iter(std::function<bool (const NamedRowValue & output,
```
jeremybarnes: Comment what's the difference between execute_iter (executeIterative?) and execute_bloc (executeBlock?)

server/bound_queries.cc
```diff
@@ -1044,7 +1291,9 @@ BoundGroupByQuery(const SelectExpression & select,
     boundRowName = rowName.bind(*groupContext);
 
     size_t maxNumRow = from.getMatrixView()->getRowCount();
-    numBuckets = maxNumRow <= 32 ? 1 : (maxNumRow / 32);
+    int numThreads = ML::num_threads();
```
jeremybarnes: We should have around 8 * num threads buckets, to avoid a single slow thread (eg, due to it being on a core that's running a pinned or kernel task) slowing down throughput.

server/bound_queries.cc
```diff
@@ -1044,7 +1291,9 @@ BoundGroupByQuery(const SelectExpression & select,
...
+    numBuckets = maxNumRow <= numThreads*32? maxNumRow / numThreads : numThreads;
```
guyd: Consider naming this magic number (e.g. minBucketSize).

jeremybarnes:
+1 modulo a bunch of minor comments from me and others.  Up to you to determine when and how to improve the getRowStream versus getNEquallySpacedRowStreams(lambda_to_call_when_one_is_created) interface.

plugins/embedding.cc
steadtler: We need to clone the stream because in the boundquery each task needs its own. If / when we do the "splitIntoManyStreams" we shouldnt need the clone anymore. Im commenting the methods in dataset.h

steadtler:
I created a followup story MLDB-1278 for some of the suggested improvements

PR #87, opened 2016-01-21
MLDB-1274 Make stemmer sum cells instead of count them

simlmx:
Make stemmer function sum the values in cells instead of just counting them.


mailletf:
+1

PR #88, opened 2016-01-21
Mldb 1276 unpack json

mailletf:


steadtler:
looks fine to me +1

nicolaskruchten:
I would put the unpacking logic documentation in the function documentation and have the procedure point there, rather than the other way around

PR #89, opened 2016-01-21
upgrade jdcal version to 1.0.1

co-dh:
The version 1.0 no longer exists and cannot be installed on pypi.
https://github.com/phn/jdcal/issues/5

### DIFF
--- a/python_constraints.txt
+++ b/python_constraints.txt
@@ -62,7 +62,7 @@ ipykernel==4.0.3
 ipython-genutils==0.1.0
 ipywidgets==4.0.2
 itsdangerous==0.24
-jdcal==1.0
+jdcal==1.0.1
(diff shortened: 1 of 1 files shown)


python_constraints.txt
```diff
@@ -62,7 +62,7 @@ ipykernel==4.0.3
 ipython-genutils==0.1.0
 ipywidgets==4.0.2
 itsdangerous==0.24
-jdcal==1.0
+jdcal==1.0.1
```
jraby: any reason not to use the latest available, 1.2 ?

co-dh: no, 1.2 should do according to it's change log. but not tested.

co-dh:
please use https://github.com/mldbai/mldb/pull/90

PR #90, opened 2016-01-21
bump jdcal to 1.0.1 because 1.0 no longer available

co-dh:


nicolaskruchten:
@jraby are we shielded from this sort of problem on khan because of the wheelhouse?

jraby:
Yes.

nicolaskruchten:
@jraby should we accept this PR? What else is required? You need to add the new version to the wheelhouse etc? Rebuild mldb_base?

jraby:
Once this is in, we need to add the package to the wheelhouse and rebuild mldb_base.

However, I think we should upgrade to the latest version (1.2)  instead of 1.0.1, unless there's some compatibility issue.

co-dh:
There should be no according to there change log.
somebody need to make sure the unit test covers this package well, and then
run the unit test.

nicolaskruchten:
Where does this dep come from? Jupyter?

jraby:
jdcal comes from openpyxl:

```
Collecting jdcal (from openpyxl==2.2.6->-r /home/jean/ws/src/mldbpro/python_requirements.txt (line 14))
  Downloading jdcal-1.2.tar.gz
```

I will update it to 1.2

jraby:
Closing, I will use 1.2.

PR #91, opened 2016-01-21
Fix date iso8601 tz parsing

FinchPowers:


wsourdeau:
I am surprised we (including I) did not notice that earlier. I think the commit messages lack a ticket number.

guyd:
It would nice to have tests around midnight to confirm days, months, years are changed correctly.

FinchPowers:
@guyd tests added.

guyd:
+1

PR #92, opened 2016-01-22
MLDB-1233 small fix for bucket number in group by

steadtler:
Fix (for now) a small issue where some rows could report a bucket index == bucketnumber.


FinchPowers:
+1 (tests pending)

PR #93, opened 2016-01-22
[MLDBFB-320] force lowecase lookup on aggregates

guyd:
I know that this pull request is not on master.  I will fix that before merging.

The lookup of the aggregate function was done in a case-sensitive way in some cases. This resulted in SelectExpression::findAggregator to return no aggregators when there were some. An immediate consequence is that the BoundGroupByQuery was not inserting the aggregator expression in the calc vector resulting in an out of bound lookup when evaluating the aggregators.

### DIFF
--- a/sql/sql_expression.cc
+++ b/sql/sql_expression.cc
@@ -297,8 +297,9 @@ ExternalAggregator lookupAggregator(const Utf8String & name)
 
 ExternalAggregator tryLookupAggregator(const Utf8String & name)
 {
+    Utf8String functionNameLower(boost::algorithm::to_lower_copy(name.extractAscii()));
     std::unique_lock<std::recursive_mutex> guard(externalAggregatorsMutex);
-    auto it = externalAggregators.find(name);
+    auto it = externalAggregators.find(functionNameLower);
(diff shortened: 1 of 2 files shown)


guyd:
@FinchPowers this is the first part of what you need to fix MLDBFB-320.

FinchPowers:
Does it make the assert reproduced by the test MLDBFB-320 go away? If so, +1! If not, we need a test case.

guyd:
The assert does not go away.  To the contrary, the assert is now happening all the time.  I did not add a test cases because there are already many tests on aggregate functions which are all passing.

nicolaskruchten:
what is being lowercased here? I'm not sure I follow :)

nicolaskruchten:
Function names (including those of aggregators) are meant to be case-sensitive.

guyd:
The aggregate function name.  Mich was using _COUNT_ in an expression.  So it should be _count_?

guyd:
After discussion with @nicolaskruchten, we must use lowercase for built-in functions and aggregators.  I will fix the code that force a lowercase before a lookup of the name as this is not required and causing issues.

PR #94, opened 2016-01-25
Allow sql.expression to be prepared which makes bind() a no-op (MLDB-1293)

jeremybarnes:
Setting prepare to true massively reduces the latency of starting a query, which helps with ad-hoc query performance.


nicolaskruchten:
Cool. If `a` calls `b` and `a` is prepared, does it slow things down if `b` is not? I.e. does the whole tree need to be prepared or just the root?

Second question: if `a` calls `b` and `a` is prepared, and then `b` is deleted/recreated, what happens? Does `a` call the new `b`?

plugins/sql_functions.cc
```diff
@@ -229,25 +249,39 @@ getStatus() const
 struct SqlExpressionFunctionApplier: public FunctionApplier {
     SqlExpressionFunctionApplier(SqlBindingScope & outerContext,
                                  const SqlExpressionFunction * function,
-                                 const SelectExpression & expression,
                                  const FunctionValues & input)
-        : FunctionApplier(function),
-          context(outerContext, input),
-          bound(expression.bind(context))
+        : FunctionApplier(function), function(function),
+          innerScope(const_cast<SqlExpressionMldbContext &>
```
steadtler: Should we create a story to resolve this const_cast?

steadtler:
Code looks fine, but Im curious as to Nick's second question, so +1 when answered.

plugins/sql_functions.cc
```diff
@@ -204,15 +203,36 @@ SqlExpressionFunctionConfigDescription()
...
              "'x, y + 1 AS z'");
+    addField("prepared", &SqlExpressionFunctionConfig::prepared,
+             "Do we pre-prepare the expression to be run many times quickly?  "
+             "If this is true, it will only be bound once, for generic "
+             "inputs, and so will allow for quick individual queries, "
+             "possibly at the expense of batch queries being slower.  "
+             "If this is false, the default, then for every query the "
+             "expression will be specialized for that query's data type. "
+             "This can lead to faster batch queries, at the expense of a "
+             "possibly high per-query overhead for individual queries.");
```
guyd: _false_ should be added as the third parameter (default).

plugins/sql_functions.cc
```diff
@@ -289,25 +321,9 @@ FunctionInfo
 SqlExpressionFunction::
 getFunctionInfo() const
 {
+    return this->info;
```
guyd: Why keep the lines after the first return?

guyd:
This is breaking for nested function calls as shown by the failing test.

plugins/sql_functions.cc
jeremybarnes: Probably a comment will do.  Since none of the methods of SqlExpressionMldbContext modify the object, it can't actually cause any harm.  I didn't want to create a new object for no reason, but maybe I should just do it and make it simple.

jeremybarnes:
In answer to Nic's questions:

1.  Only the root of the tree needs to be prepared.  The rest will be bound at the same time.
2.  If b is deleted after being pre-bound to A, then it there should still be a reference to everything needed and it should continue to function.  That hasn't been tested, though.

jeremybarnes:
Note: tests all pass on my dev machine and Jenkins is slow.  I'll merge immediately once I get a +1

mailletf:
+1 but we should probably add a testcase that's using the new option

jeremybarnes:
There is a testcase in MLDBpro... but you're right.  @nicolaskruchten would be worth adding another issue for that.

PR #97, opened 2016-01-25
[MLDBFB-320] related to MLDB-957, built-in function names should be c…

guyd:
…ase-sensitive

This was causing some instability since part of the code was interpreting a function as being an aggregator when some other parts were not!


FinchPowers:
+1

PR #98, opened 2016-01-25
MLDB-1295 python mldb wrapper

FinchPowers:


plugins/lang/python/python_loader.cc
```diff
@@ -497,6 +501,127 @@ runPythonScript(std::shared_ptr<PythonContext> titl,
...
+            self.elapsed           = 'unimplemented'
+            self.encoding          = 'unimplemented'
+            self.history           = 'unimplemented'
+            self.iter_content      = 'unimplemented'
+            self.iter_lines        = 'unimplemented'
+            self.links             = 'unimplemented'
+            self.ok                = 'unimplemented'
+            self.raise_for_status  = 'unimplemented'
+            self.reason            = 'unimplemented'
+            self.request           = 'unimplemented'
```
FinchPowers: @nicolaskruchten I left it there to show the list that requests Response object has. We can remove them if needed.

testing/python_mldb_interface_test.py
```diff
@@ -0,0 +1,122 @@
...
+
+    res = unittest.main(exit=False, argv=argv).result
+    mldb.log(res)
+    got_err = False
+    for err in res.errors + res.failures:
+        got_err = True
+        mldb.log(str(err[0]) + "\n" + err[1])
+
+    if not got_err:
+        mldb.script.set_return("success")
```
FinchPowers: It would be great if the mldb wrapper had that bloc integrated. It would allow to call "mldb.run_tests()" rather than copy paste that bloc from test case to test case.

plugins/lang/python/python_loader.cc
```diff
@@ -497,6 +501,127 @@ runPythonScript(std::shared_ptr<PythonContext> titl,
...
+            return response
+
+        def log(self, thing):
+            self.mldb.log(str(thing))
+
+        @property
+        def script(self):
+            return self.mldb.script
+
+        def _is_args_configured(self, args, kwargs):
```
nicolaskruchten: we should just choose either args or kwargs, not both. I'm torn but I'm leaning towards just supporting kwargs

plugins/lang/python/python_loader.cc
```diff
@@ -497,6 +501,127 @@ runPythonScript(std::shared_ptr<PythonContext> titl,
...
+            self.raise_for_status  = 'unimplemented'
+            self.reason            = 'unimplemented'
+            self.request           = 'unimplemented'
+
+        def json(self):
+            return self.__class__.jsonlib.loads(self.text)
+
+    class wrap(object):
+        def __init__(self, mldb):
+            self.mldb = mldb
```
nicolaskruchten: the 'old' mldb should be called `_mldb` or something, no?

FinchPowers: +1

plugins/lang/python/python_loader.cc
FinchPowers: will do

testing/python_mldb_interface_test.py
```diff
@@ -0,0 +1,122 @@
...
+# Mich, 2016-01-25
+# This file is part of MLDB. Copyright 2015 Datacratic. All rights reserved.
+#
+import unittest
+
+if False:
+    mldb = None
+    mldb_wrapper = None
+
+mldb = mldb_wrapper.wrap(mldb)
```
nicolaskruchten: if `mldb_wrapper` is a global, then we don't need to call a method with an argument, right? We can just do something simpler like `mldb = api_v2()` and that will return a nice object?

testing/python_mldb_interface_test.py
nicolaskruchten: approved.

plugins/lang/python/python_loader.cc
nicolaskruchten: GET -> kwargs
PUT/POST -> one arg

testing/python_mldb_interface_test.py
```diff
@@ -0,0 +1,92 @@
+#
+# python_mldb_interface_test.py
+# Mich, 2016-01-25
+# This file is part of MLDB. Copyright 2016 Datacratic. All rights reserved.
+#
+import unittest
+
+if False:
+    mldb_wrapper = None
```
nicolaskruchten: is this now unnecessary because of the `# noqa` ?

plugins/lang/python/python_loader.cc
```diff
@@ -497,6 +501,125 @@ runPythonScript(std::shared_ptr<PythonContext> titl,
...
+
+        def _perform(self, method, url, *args, **kwargs):
+            raw_res = self._mldb.perform(method, url, *args, **kwargs)
+            response = mldb_wrapper.Response(url, raw_res)
+            if response.status_code < 200 or response.status_code >= 400:
+                raise mldb_wrapper.ResponseException(response)
+            return response
+
+        def log(self, thing):
+            self._mldb.log(str(thing))
```
nicolaskruchten: can this pretty-print dictionaries? i.e. with indentation and without the `u'...'` that comes from the output of `json.loads()`?

plugins/lang/python/python_loader.cc
```diff
@@ -497,6 +501,125 @@ runPythonScript(std::shared_ptr<PythonContext> titl,
...
+
+        @property
+        def script(self):
+            return self._mldb.script
+
+        def get(self, url, **kwargs):
+            iter_on = None
```
nicolaskruchten: unnecessary line?

plugins/lang/python/python_loader.cc
```diff
@@ -497,6 +501,125 @@ runPythonScript(std::shared_ptr<PythonContext> titl,
...
+            query_string = [[str(k), str(v)] for k, v in kwargs.iteritems()]
+            return self._perform('GET', url, query_string)
+
+        def _post_put(self, verb, url, data=None):
+            if data is None:
```
nicolaskruchten: I'm ok with no default value for data here...

testing/python_mldb_interface_test.py
```diff
@@ -0,0 +1,92 @@
...
+
+    def test_get(self):
+        res = mldb.get('/ping')
+        mldb.log(res)
+        mldb.log(res.text)
+        mldb.log(type(res))
+        mldb.log(dir(res))
+
+        with self.assertRaises(mldb_wrapper.ResponseException):
+            mldb.get("/unexisting")
```
nicolaskruchten: can we get a small test case that shows the contents of the `ResponseException` and how to access them?

FinchPowers: Good idea.

testing/python_mldb_interface_test.py
FinchPowers: No because ```mldb_wrapper``` would be undefined at line 26

FinchPowers: To spare a line I could do
```
mldb_wrapper = mldb_wrapper or None # noqa
```

plugins/lang/python/python_loader.cc
FinchPowers: To remove the prefix u we have to reencode utf-8 to ascii, which will break when we have utf-8.
Alternative: python 3. ;)

testing/python_mldb_interface_test.py
```diff
@@ -0,0 +1,98 @@
...
+        mldb.post(url + '/rows', {
+            'rowName' : 'row1',
+            'columns' : [['colA', 1, 0]]
+        })
+        mldb.post(url + '/commit')
+
+        res = mldb.query('SELECT * FROM ds').json()
+        self.assertEqual(res, [['_rowName', u'colA'], ['row1', 1]])
+
+    def test_response_exception(self):
```
FinchPowers: @nicolaskruchten The new test for ResponseException

guyd:
Briefly reviewed.  Looks good to me.

nicolaskruchten:
waiting for Jenkins and then +1

PR #99, opened 2016-01-25
Mldb 1272 regression training

guyd:


guyd:
Merging right away since this is only the addition of a test.

PR #100, opened 2016-01-25
[MLDB-1299] returning a descriptive error message when trying to map …

guyd:
…an empty file.

This is actually fixing MLDB-1303 which is deprecating MLDB-1299.
We want to be able to csv import empty files.  This required to work around a failure (by design) in mmap to create mapped file for an empty file.


mailletf:
+1

plugins/csv_dataset.cc
```diff
@@ -575,7 +575,13 @@ struct CsvDataset::Itl: public TabularDataStore {
         filename = config.dataFileUrl.toString();
         
         // Ask for a memory mappable stream if possible
-        ML::filter_istream stream(filename, { { "mapped", "true" } });
+        ML::filter_istream stream;
+        try {
+            stream.open(filename, { { "mapped", "true" } });
+        }
+        catch (const std::exception & e) {
+             throw HttpReturnException(400, "dataFileUrl pointed to an empty file");
```
jeremybarnes: Two problems here:
1.  It's not at all true that this exception means the file is empty.  It could also be because it doesn't exist.
2.  You're eating the helpful error message (eg, "file not found") and replacing with another one ("dataFileUrl pointed to an empty file)
3.  You don't give the user context: that they are opening a CSV file called xxx.csv, which will make it harder for them to know where the error comes from
4.  I can't count

testing/MLDB-1155_csv_line_endings.py
```diff
@@ -219,4 +219,21 @@
     assert json.loads(result['response'])[1][2] == 1
 
 
+# importing an empty CSV gives a specific error message
+# the problem is with boost implementation of mapped file
+# which is only possible when the file is not compressed
+path = "tmp/csv_empty.csv"
```
jeremybarnes: Add  test-case for a non-existant file?

plugins/csv_dataset.cc
jeremybarnes: Also, you should use the rethrowHttpException() in http_exception.h when your error was caused by an existing exception, not throw a new one.

jeremybarnes: Finally, the filter_istream should not fail to map an empty file; it should return a zero length mapping.  Once that is fixed, the check for zero length can be made elsewhere (and @nicolaskruchten does it make sense to disallow an empty csv file?  One with no rows and no columns is empty, but a valid (if useless) CSV file.

nicolaskruchten: Does it make more sense to create empty datasets for empty files? I guess I
can see that... Non-existent files should cause an exception though, right?

Is it possible to specify a csv dataset with no columns? Should we return
something different if headers are specified and the file is empty vs no
headers and empty file!

On Monday, 25 January 2016, Jeremy Barnes <notifications@github.com> wrote:

> In plugins/csv_dataset.cc
> <https://github.com/mldbai/mldb/pull/100#discussion_r50762565>:
>
> > @@ -575,7 +575,13 @@ struct CsvDataset::Itl: public TabularDataStore {
> >          filename = config.dataFileUrl.toString();
> >
> >          // Ask for a memory mappable stream if possible
> > -        ML::filter_istream stream(filename, { { "mapped", "true" } });
> > +        ML::filter_istream stream;
> > +        try {
> > +            stream.open(filename, { { "mapped", "true" } });
> > +        }
> > +        catch (const std::exception & e) {
> > +             throw HttpReturnException(400, "dataFileUrl pointed to an empty file");
>
> Finally, the filter_istream should not fail to map an empty file; it
> should return a zero length mapping. Once that is fixed, the check for zero
> length can be made elsewhere (and @nicolaskruchten
> <https://github.com/nicolaskruchten> does it make sense to disallow an
> empty csv file? One with no rows and no columns is empty, but a valid (if
> useless) CSV file.
>
> —
> <https://github.com/mldbai/mldb/pull/100/files#r50762565>.
>


-- 
http://datacratic.com/

jeremybarnes: I would not make it fail with empty files.  In my experience, in real world situations they often occur, and really do have no rows.  Eg, if there was a data collection process that run hourly, but for some hours there was no activity.

guyd:
All valid points.  I changed the code to rethrow the catched std::exception.

plugins/csv_dataset.cc
guyd: The problem is that boost chokes on these empty files.  The problem does not come from filter_istream.  We can modify filter_istream to test for empty files before trying to map them.

nicolaskruchten:
I've created MLDB-1303 to cover the work to properly handle empty files. The idea of MLDB-1299 was to just return a more informative message, without incurring too too much rework.

PR #102, opened 2016-01-26
Mldb 12550 3 way outer join

steadtler:
This fixes calling dataset.rowName() in 3 way joins by allowing the joinDataset to override the getFunction with the table name in order to check what was the original rowName from that dataset in the joined row.

Does *not* fix it for N-way joins yet, that will be next, I want to pull this to unblock some users.

### DIFF
--- a/testing/MLDB-180-basic-join.js
+++ b/testing/MLDB-180-basic-join.js
@@ -28,7 +28,6 @@ function testQuery(query, expected) {
 
 var dataset1 = mldb.createDataset({type:'sparse.mutable',id:'test1'});
 var dataset2 = mldb.createDataset({type:'sparse.mutable',id:'test2'});
-var dataset2 = mldb.createDataset({type:'sparse.mutable',id:'test2'});
 
 var ts = new Date("2015-01-01");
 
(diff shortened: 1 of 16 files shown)


testing/MLDB-180-basic-join.js
```diff
@@ -380,4 +379,56 @@ testQuery('SELECT * FROM test1 RIGHT JOIN test2 ON test1.x = test2.x AND test2.x
...
+dataset5.recordRow("ex6", [ [ "x", 2, ts ], ["z", 2, ts] ]);
+dataset5.recordRow("ex1", [ [ "x", null, ts ], ["z", 3, ts] ]);
+
+dataset3.commit()
+dataset4.commit()
+dataset5.commit()
+
+expected = [
+   [ "_rowName", "test3.x", "test3.y", "test3.z", "test4.x", "test4.z" ],
+   [ "ex1-", 1, 2, null, null, null ],
```
guyd: I thought we had decided to use a period instead of a dash?

guyd:
I did not have a chance to dig into the join code yet but this seems ok on first look. +1

testing/MLDB-180-basic-join.js
steadtler: As discussed, dot is for flattened column names.

PR #104, opened 2016-01-27
Don't require query interfaces to share ownership of their arguments (MLDB-1318)

jeremybarnes:
Rationale
- Currently, we force callers of the query functions (in dataset.h, analytics.h, bound_queries.h) to pass shared pointers to owned versions of the where expressions
- Not everything owns those pointers, and so this decision trickles down through all of the interfaces
- Since these are passive interfaces, they should accept const SqlExpression & parameters, not just std::shared_ptr<SqlExpression>

DOD
- Passive query interfaces take const references to the expressions themselves
- The enable_shared_from_this or shallow_copy() interface is used in the rare case that a pointer is actually needed.


jeremybarnes:
Note again: only failing test is MLDB-1273, which was already a regression.

server/bound_queries.cc
```diff
@@ -935,16 +935,16 @@ BoundSelectQuery(const SelectExpression & select,
 {
     try {
 
-        ExcAssert(where);
+        ExcAssert(&where);
```
guyd: This is probably not required.

server/bound_queries.h
```diff
@@ -143,7 +143,7 @@ struct BoundGroupByQuery {
 
     const Dataset & from;
     WhenExpression when;
-    std::shared_ptr<SqlExpression> where;
+    const SqlExpression & where;
```
guyd: Who own that reference?

guyd:
One small comment and one question +1

server/bound_queries.h
jeremybarnes: By convention, things passed as a const reference are owned by the caller.

server/bound_queries.cc
jeremybarnes: It catches errors in the upstream components.  Agree we should remove it once it hasn't triggered for a while.

server/bound_queries.h
guyd: My concern was that the owner would free that object before BoundGroupByQuery is done with it.

jeremybarnes: Fair point.  Probably that will need to be handled by documentation... although the other parameters are in the same boat.

guyd: A comment in the code is probably good enough for the time being.

PR #105, opened 2016-01-27
Small cleanups in sql expression ops (MLDB-1321)

jeremybarnes:
- Improves the scope of SqlExpression::constantValue() allowing for it to be used in more situations
- Better error messages
- Improved accuracy in isConstant()
- Make function binding exception safer

### DIFF
--- a/sql/builtin_dataset_functions.cc
+++ b/sql/builtin_dataset_functions.cc
@@ -133,11 +133,16 @@ BoundTableExpression sample(const SqlBindingScope & context,
                             const Utf8String& alias)
 {
     if (args.size() != 1)
-        throw HttpReturnException(500, "sample() takes 1 dataset as input, "
-                "followed by a options");
+        throw HttpReturnException(400, "sample() takes 1 dataset as input, "
+                                  "followed by a row of dataset options",
+                                  "options", options,
+                                  "alias", alias);
 
     if(!options.isRow())
-        throw ML::Exception("options should be a row");
+        throw HttpReturnException(400, "options should be a row; we got "
+                                  + jsonEncodeStr(options),
+                                  "options", options,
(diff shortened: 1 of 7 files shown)


sql/builtin_dataset_functions.cc
```diff
@@ -133,11 +133,16 @@ BoundTableExpression sample(const SqlBindingScope & context,
                             const Utf8String& alias)
 {
     if (args.size() != 1)
-        throw HttpReturnException(500, "sample() takes 1 dataset as input, "
-                "followed by a options");
+        throw HttpReturnException(400, "sample() takes 1 dataset as input, "
+                                  "followed by a options",
```
guyd: a -> an

sql/sql_expression.cc
```diff
@@ -1539,16 +1539,21 @@ bool
...
 }
 
 ExpressionValue
 SqlExpression::
 constantValue() const
 {
-    throw HttpReturnException(400, "Expression is not provably constant",
-                              "surface", surface,
-                              "ast", print());
+    SqlExpressionConstantScope scope;
```
guyd: Should we call isConstant first or will this throw something meaningful for non-constant expression?

guyd:
existing typo that was not fixed + one question
+1

sql/sql_expression.cc
jeremybarnes: The isConstant has a slightly tighter scope (eg, it's possible that something is a constant because it has a case which avoids the evaluation of  non-constant part).  I will improve the error message.

sql/builtin_dataset_functions.cc
jeremybarnes: done

PR #106, opened 2016-01-27
Mldb 1322 sum stem token

simlmx:
Test case for my sum(stemmer(tonenizer(... usercase.
It turns out the bug I was trying to point out was very recently fixed. I figured I would add the test case anyway.


guyd:
+1

testing/MLDB-1322-sum_stem_token.py
```diff
@@ -0,0 +1,206 @@
...
+# Step 6: same thing but from a merged dataset
+ds = mldb.create_dataset({
+    'type': 'sparse.mutable',
+    'id': 'label2'})
+ds.record_row('row_0', [['label2',2, 0]])
+ds.record_row('row_1', [['label2', 2, 0]])
+ds.record_row('row_2', [['label2',3, 0]])
+
+q = '''
+SELECT sum(stem({words: {tokenize(txt) as *}})[words]) as *
```
nicolaskruchten: can you please add a case for `sum(stem({words: tokenize(txt) })[words]) as *`? i.e. without the "{ ... as *}` around the `tokenize()`? If MLDB doesn't support this, please comment out the test case and I will create an issue.

PR #107, opened 2016-01-27
Improved function binding for new executor (MLDB-1319)

jeremybarnes:
DOD
- sql.query of x JOIN y on x.rowName() = y.rowName() works
- sql.query of x JOIN y on CAST (x.rowName() AS INTEGER) = CAST (y.rowName() AS INTEGER) works
- Test cases for the above
- Use of a parameter (eg, $1) in sql.query join condition works


testing/MLDB-1319-new-executor-function-binding.js
```diff
@@ -0,0 +1,87 @@
+// This file is part of MLDB. Copyright 2015 Datacratic. All rights reserved.
```
guyd: We may want to update you emacs template to say 2016!

guyd:
Briefly reviewed.  +1

PR #108, opened 2016-01-27
Implemented whole table mode for sql.query (MLDB-1320)

jeremybarnes:
This allows sql.query to use the whole table and turn it into a row, which enables much more sophisticated output formatting (especially using joins, etc).


container_files/assets/www/doc/builtin/functions/SqlQueryFunction.md
```diff
@@ -6,10 +6,22 @@ The query is parameterized on the function input values.
...
 
-Functions created this way executes the given query against the given dataset
-and outputs the `SELECT` expression applied to the first matching
-row (ie, with an `OFFSET` of 0 and a `LIMIT` of 1).  If multiple
-rows are matched, all but the first will be ignored.
+Functions created this way executes the given query against the given dataset.
+
+There are two possibilities for the output:
+
+1.  If the `output` field is set to `FIRST_ROW` (the default), then
```
guyd: This is not completely clear at first read.  Consider adding an example of both modes.

plugins/sql_functions.cc
```diff
@@ -47,6 +61,13 @@ SqlQueryFunctionConfigDescription()
              "SQL query to run.  The values in the dataset, as "
              "well as the input values, will be available for the expression "
              "calculation");
+    addField("output", &SqlQueryFunctionConfig::output,
+             "Controls how the query output is converted into a row. "
+             "'FIRST_ROW' (default) will return only the first row produced "
+             "by the query.  'NAMED_COLUMNS' will construct a row from the "
+             "whole returned table, which must have a 'value' column "
+             "containing the value.  If there is a 'column' column, it will "
+             "be used as a column name, otherwise the row name will be used.");
```
guyd: We should populate the third parameter with 'FIRST_ROW' to better identify the default.

plugins/sql_functions.cc
```diff
@@ -138,17 +166,83 @@ struct SqlQueryFunctionApplier: public FunctionApplier {
...
+                        if (col == ColumnName("column")) {
+                            foundCol = ColumnName(val.getAtom().toUtf8String());
+                            ++numFoundCol;
+                        }
+                        else if (col == ColumnName("value")) {
+                            foundVal = std::move(val);
+                            ++numFoundVal;
+                        }
+                        else {
+                            throw HttpReturnException(400, "Row returned from NAMED_COLUMNS SQL query can only contain 'column' and 'value' columns",
```
guyd: Row -> Rows

testing/MLDB-1320-sql-query-whole-table.js
```diff
@@ -0,0 +1,157 @@
+// This file is part of MLDB. Copyright 2015 Datacratic. All rights reserved.
```
guyd: 2015 -> 2016

guyd:
Briefly reviewed +1

container_files/assets/www/doc/builtin/functions/SqlQueryFunction.md
jeremybarnes: I hear you for examples.

For the enumeration, I will auto-generate the table from the enum documentation.

container_files/assets/www/doc/builtin/functions/SqlQueryFunction.md
```diff
@@ -25,15 +28,59 @@ put them inside '"' characters.  For example,
...
+
+## Output format
+
+There are two possibilities for the output:
+
+1.  If the `output` field is set to `FIRST_ROW` (the default), then
+    it will outputs the `SELECT` expression applied to the first matching
+    row (ie, with an `OFFSET` of 0 and a `LIMIT` of 1).  If multiple
+    rows are matched, all but the first will be ignored.
+2.  If the `output` field is set to `NAMED_COLUMNS`, then the query
```
nicolaskruchten: I feel like we could already do this in SQL with something like transpose() no?

container_files/assets/www/doc/builtin/functions/SqlQueryFunction.md
```diff
@@ -25,15 +28,59 @@ put them inside '"' characters.  For example,
...
+    must return a two-column dataset with a `column` and a `value`
+    each column.  Each row  in the output will generate a single
+    column in the output row, with the column name equal to value of the
+    `column` column, and the column value equal to the value of the
+    `value` column.  In this case the `OFFSET` and `LIMIT` will be
+    respected.  This allows for more sophisticated rows to be returned
+    from queries, and is especially useful in conjunction with joins.
```
nicolaskruchten: I'm not sure I understand the use-case here... Can we see a real-world application?

jeremybarnes:
Reverted from master pending outcome of discussion.

container_files/assets/www/doc/builtin/functions/SqlQueryFunction.md
jeremybarnes: The application was:
- We have a 1000 element embedding with category classes across integers 0 to 999
- Category names are in a separate dataset
- We want to create a function that takes the 1,000 element embedding and transforms it into a single row with 5 columns: the name of the column being the category, the score being the value, and only the top 5 scores being represented
- Using this functionality, it could be represented as an sql.query which joins the embedding indexes on the category labels table, with an order by and limit.   Once that was done, the functionality in this PR was used to collapse the (multiple) output rows of the query down onto a single row for output.

Doing it another way is perfectly acceptable.  I hadn't thought of using transpose.

nicolaskruchten:
+1 to re-merge

PR #109, opened 2016-01-27
[MLDB-1260] modified the handling of exception to return a httpCode i…

guyd:
…n all cases + fix a few places where the wrong exception type was thrown.

### DIFF
--- a/testing/MLDB-951-run-on-creation.py
+++ b/testing/MLDB-951-run-on-creation.py
@@ -99,18 +99,21 @@ for row in transformed_rows:
     assert int(row['rowName']) + 1 == row['columns'][0][1], 'the transform was not applied correctly'
 
 # checking return code for invalid runs
-response =  mldb.perform('PUT', "/v1/procedures/transform_procedure", [], 
+res =  mldb.perform('PUT', "/v1/procedures/transform_procedure", [], 
                          {
                              "type": "transform",
                              "params": {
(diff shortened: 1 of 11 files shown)


testing/MLDB-951-run-on-creation.py
```diff
@@ -99,18 +99,21 @@ def load_test_dataset():
...
 
 # checking return code for invalid runs
-response =  mldb.perform('PUT', "/v1/procedures/transform_procedure", [], 
+res =  mldb.perform('PUT', "/v1/procedures/transform_procedure", [], 
                          {
                              "type": "transform",
                              "params": {
                                  "inputData": "select x + 1 as x from dataset1",
-                                 "outputDataset": { "id": "dataset2", "type":"beh" }, # try to create a non-mutable
+                                 "outputDataset": { "id": "dataset2", "type":"beh" }, # try to create a type that is not available
```
nicolaskruchten: that's an interesting change... in mldbpro, `beh` exists and this fails because it's not mutable. In mldb this fails because `beh` isn't available :)

guyd: I will change that to avoid the confusion.  After all, the only thing we want is to cause an error.

steadtler:
Looks good to me, +1

jeremybarnes:
Looks good to me too.

PR #110, opened 2016-01-28
MLDB-1315 row as table expression

jeremybarnes:
This allows rows to be used in table expressions, allowing rewriting and filtering of rows and gathering using joins.

### DIFF
--- a/container_files/assets/www/doc/builtin/sql/FromExpression.md
+++ b/container_files/assets/www/doc/builtin/sql/FromExpression.md
@@ -78,3 +78,53 @@ SELECT x.* FROM merge(dataset1, dataset2, dataset3) AS x
 ```
 
 See ![](%%doclink merged dataset) for more details.
+
+## Using rows as a dataset
+
+In some circumstances, it may be useful to use a row as a dataset,
+particularly when using the ![](%%doclink sql.query function) or a
+sub-select.  This can be done using the syntax
+
+```
+SELECT ... FROM row_dataset(expression) ...
+```
+
+When this construct is used, a dataset is constructed with one row
+for each column in the expression, with one column called `value`
+containing the value of the column, and one column called `column`
+with the column name.  For example, the following expression
+
+```
+row_dataset({x: 1, y:2, z: 'three'})
+```
+
+would yield the following dataset:
+
+```
+column    value
+x         1
+y         2
+z         "three"
+```
+
+As an example, the following `sql.query` object would strip
+out any numeric-valued columns and uppercase all names from a
+passed in row:
+
+```
+POST /v1/functions/row_transform {
+    type: 'sql.query',
+    params: {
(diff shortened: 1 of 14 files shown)


container_files/assets/www/doc/builtin/sql/FromExpression.md
```diff
@@ -78,3 +78,54 @@ SELECT x.* FROM merge(dataset1, dataset2, dataset3) AS x
...
+
+    }
+}
+
+SELECT row_transform({input: {x: 1, y: 2, z: "three"}})[output] AS *
+
+{ Z: "three" }
+```
+
+It can also be used to join a row against another table, in order to
```
nicolaskruchten: I'm not at all sure I understand the use-case here. Can you describe a real-world scenario?

container_files/assets/www/doc/builtin/sql/FromExpression.md
```diff
@@ -78,3 +78,54 @@ SELECT x.* FROM merge(dataset1, dataset2, dataset3) AS x
...
+
+As an example, the following `sql.query` object would strip
+out any numeric-valued columns and uppercase all names from a
+passed in row:
+
+```
+POST /v1/functions/row_transform {
+    type: 'sql.query',
+    params: {
+        query: 'SELECT upper(rowName()) AS column, value FROM ROW TABLE ($input) WHERE CAST (value AS NUMBER) IS NULL',
```
nicolaskruchten: I would prefer syntax that looks more like a function-call like `sample` or `transpose` which we already support. Perhaps `row_table(...)` ?

container_files/assets/www/doc/builtin/sql/FromExpression.md
nicolaskruchten: OK, I see the scenario in the JIRA issue.

container_files/assets/www/doc/builtin/sql/FromExpression.md
nicolaskruchten: The fact that a `ROW TABLE` has a `rowName()` function seems very counter-intuitive, which leads me to think that the naming is off.

nicolaskruchten:
-1 while we consider the DOD.

container_files/assets/www/doc/builtin/sql/FromExpression.md
jeremybarnes: All datasets have a rowName() function.  Would it make more sense that the returned dataset had a column called 'column'?  That way we wouldn't need to use the rowName() in the query.

nicolaskruchten:
Sorry about the multiple name changes here but I just realized that we don't use the word "table" in MLDB, we try to use "dataset" consistently. So we should probably change this to `row_dataset`.

nicolaskruchten:
+1 to merge after `row_table` -> `row_dataset`

PR #114, opened 2016-01-28
Mldb 1255 Fix calling dataset.rowName in N-way joins 

steadtler:
And also replace empty rowname entries with "null" in outer joins.

I wasnt sure is getOriginalRowName should be part of the dataset interface or the MatrixView interface - I picked the dataset interface because I thought MatrixView and ColumnIndex were mainly meant as pure interfaces, while getOriginalRowName should only be overriden by the JoinDataset. But Im open to suggestions.

I upgraded the PR to include the fixes to MLDB-1305 since no one had reviewed it yet.


testing/MLDB-1305_rowNames_join.py
```diff
@@ -0,0 +1,79 @@
+
+# This file is part of MLDB. Copyright 2015 Datacratic. All rights reserved.
```
guyd: 2015 -> 2016

testing/MLDB-1305_rowNames_join.py
```diff
@@ -0,0 +1,79 @@
...
+
+
+res = mldb.perform("GET", "/v1/query", [["q", """
+    SELECT dataset1.rowName(), dataset2.rowName()
+    FROM dataset1
+    JOIN dataset2 ON dataset2.rowName() = dataset1.rowName()
+    """]])
+
+mldb.log(res)
+assert res["statusCode"] != 404
```
guyd: This does not test much.  Can we check for the body as well?  And why 404?

types/string.cc
```diff
@@ -518,6 +518,32 @@ removePrefix(const Utf8String & prefix)
     return false;
 }
 
+bool
+Utf8String::
+removeSuffix(const Utf8String & suffix)
+{
+   if (suffix.empty())
```
guyd: indentation problem

testing/MLDB-1305_rowNames_join.py
steadtler: Francois wrote the test. I could check some of the output...

guyd:
good test coverage +1

PR #116, opened 2016-01-28
MLDB-1216 fetcher function

jeremybarnes:
Allows us to fetch URLs as part of a query and operate on what is behind them.

### DIFF
--- a/container_files/assets/www/doc/builtin/sql/ValueExpression.md
+++ b/container_files/assets/www/doc/builtin/sql/ValueExpression.md
@@ -339,6 +339,10 @@ Note that this syntax is not part of SQL, it is an MLDB extension.
   (or string) argument as a string.
 - `base64_decode(string)` returns a blob containing the decoding of the
   base-64 data provided in its argument.
+- `json_decode(string)` returns a row with the JSON decoding of the
+  string in the argument.
+- `extract_column(row)` extracts the given column from the row, keeping
+  only its latest value by timestamp.
 
 ### Numeric functions
 
--- a/container_files/assets/www/doc/builtin/sql/ValueExpression.md
+++ b/container_files/assets/www/doc/builtin/sql/ValueExpression.md
@@ -339,8 +339,9 @@ Note that this syntax is not part of SQL, it is an MLDB extension.
   (or string) argument as a string.
 - `base64_decode(string)` returns a blob containing the decoding of the
   base-64 data provided in its argument.
-- `json_decode(string)` returns a row with the JSON decoding of the
-  string in the argument.
+- `parse_json(string)` returns a row with the JSON decoding of the
+  string in the argument.  Nested arrays and objects will be parsed
+  recursively; no flattening is performed.
 - `extract_column(row)` extracts the given column from the row, keeping
   only its latest value by timestamp.
 
@@ -479,7 +480,7 @@ number of occurrences of those tokens within `str`. For example `tokenize('a b b
 The `unpack_json(str)` function will parse the string `str` as a JSON
 object and unpack it into multiple columns following the  algorithm
 outlined below. Note that JSON objects shown in the tables below are
-string reprensetations of the JSON.
+string representations of the JSON.
 
 Each `(key, value)` pair will be recorded as the column name and cell value respectively. The line `{"a": 5, "b": true}` is recorded as:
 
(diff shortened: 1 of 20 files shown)


nicolaskruchten:
is `json_decode` sufficiently different from `unpack_json` to warrant them coexisting within MLDB?

jeremybarnes:
json_decode performs a consistent, hierarchical parsing of JSON into an exact mirror of its structure.  This means that things aren't pre-sparsified for importation tasks, but it's also exactly what you would expect if doing something else with it.  unpack_json is more specialized, towards importing JSON with arrays in a manner which is pre-sparsified.

I would modify the implementation of unpack_json to be  more similar to the json_decode implementation, and possibly improve the naming.

Note that the function was added originally for a test-case, which could use unpack_json.

nicolaskruchten:
how does json_decode deal with JSON like: `{"a": [ {"b":1}, {"b":2} ] }` ?

jeremybarnes:
Since an expression value can be arbitrarily nested, it nests it.  If it's recorded into a dataset it will be flattened then.

nicolaskruchten:
So it would come out as 2 columns: `a.0.b` and `a.1.b` ?

jeremybarnes:
yes

nicolaskruchten:
How hard would it be to graft the behaviour of `unpack_json` onto `json_decode` as an option?

guyd:
What is the reason of having such function in MLDB?  After all, anyone using MLDB already performs REST calls so why do we need to support a partial implementation in MLDB?

nicolaskruchten:
It enables MLDB to scrape other websites and interact with/wrap other APIs.

nicolaskruchten:
Options for `json_decode`, in order of desirability:

1. unify with `unpack_json` with a second argument as a row with a single key: `flatten_arrays` defaulting to `true`, and when it's false use the `unpack_json` logic
1. don't commit `decode_json` for now and rewrite the test case to use `unpack_json`
1. commit `decode_json` and we will unify later

jeremybarnes:
OK, so I have unified the implementations, but kept the two names in to avoid breakage.

Note that there are two side-effects:
1.  bools now come out as integers in one-hot encodings
2.  row names for the JSON dataset are now integers, without the "row" prefix

container_files/assets/www/doc/builtin/sql/ValueExpression.md
```diff
@@ -472,48 +476,60 @@ number of occurrences of those tokens within `str`. For example `tokenize('a b b
 
 ### JSON unpacking <a name="unpack_json"></a>
 
-The `unpack_json(str)` function will parse the string `str` as a JSON object and unpack it into multiple columns following the  algorithm outlined below. Note that JSON objects shown in the tables below are string reprensetations of the JSONs.
+The `unpack_json(str)` function will parse the string `str` as a JSON
+object and unpack it into multiple columns following the  algorithm
+outlined below. Note that JSON objects shown in the tables below are
+string reprensetations of the JSON.
```
guyd: reprensetations - representations

guyd:
+1 beside a few typos

container_files/assets/www/doc/builtin/sql/ValueExpression.md
```diff
@@ -339,6 +339,10 @@ Note that this syntax is not part of SQL, it is an MLDB extension.
   (or string) argument as a string.
 - `base64_decode(string)` returns a blob containing the decoding of the
   base-64 data provided in its argument.
+- `json_decode(string)` returns a row with the JSON decoding of the
```
nicolaskruchten: this is now `parse_json` ?

container_files/assets/www/doc/builtin/sql/ValueExpression.md
```diff
@@ -472,48 +476,60 @@ number of occurrences of those tokens within `str`. For example `tokenize('a b b
 
 ### JSON unpacking <a name="unpack_json"></a>
 
-The `unpack_json(str)` function will parse the string `str` as a JSON object and unpack it into multiple columns following the  algorithm outlined below. Note that JSON objects shown in the tables below are string reprensetations of the JSONs.
+The `unpack_json(str)` function will parse the string `str` as a JSON
```
nicolaskruchten: so `unpack_json` is still there but uses the same underlying implementation as `parse_json` and so has changed behaviour? I'm confused.

jeremybarnes: Yes.  I ran out of time to unify the actual function calls, but the internal implementation is unified causing the changes I mentioned.

PR #117, opened 2016-01-29
[MLDB-1303] modified filter_stream to not fail when trying to memory …

guyd:
...memory map an empty file

### DIFF
--- a/vfs/filter_streams.cc
+++ b/vfs/filter_streams.cc
@@ -794,6 +794,8 @@ struct RegisterFileHandler {
                    const std::map<std::string, std::string> & options,
                    const OnUriHandlerException & onException)
     {
+        using namespace boost::iostreams;
+
         if (resource == "")
             resource = "/dev/null";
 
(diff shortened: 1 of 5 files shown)


vfs/filter_streams.cc
```diff
@@ -805,22 +807,13 @@ struct RegisterFileHandler {
...
-                //                        resource.c_str(), strerror(errno));
-                UriHandlerOptions options;
-                options.mapped = source.data();
-                options.mappedSize = source.size();
-                return UriHandler(buf.get(), buf, info, options);
-            }
-            else {
+            // MLDB-1303 mmap fails on empty files - force filebuf interface
+            // on empty files despite the mapped option
+            if (!options.count("mapped") || !Datacratic::getUriSize(uri)) {
```
jeremybarnes: We should use the size that's in the getUriObjectInfo, to avoid a second call to stat().

jeremybarnes:
Minor comment & +1

vfs/filter_streams.cc
guyd: Will do.

PR #118, opened 2016-01-29
MLDB-1329 sql query memory leak

jeremybarnes:
This fixes the memory usage explosion running the pooling function (or sql.query).  It also addresses other performance issues discovered along the way.

Note that the root cause wasn't completely determined (it looks like some kind of circular reference between shared pointers), but this addresses the problem.


guyd:
Is the commit 567f96b part of that PR?

guyd:
Reviewed except for commit 567f96b.  +1

jeremybarnes:
The extra commit made the test case much faster, which improved the iteration cycle.  It could be put into a separate PR but it's pretty low-risk.

guyd:
Briefly reviewed 567f96b as well. +1

jeremybarnes:
@nicolaskruchten @mailletf I'll let you merge this if and when you verify that it helps.

mailletf:
I'm happy to confirm this fixes the issue on the POC2 data! Thanks for looking into it

PR #119, opened 2016-01-29
MLDB-1314 allow user functions to accept output from other function d…

steadtler:
…irectly


sql/sql_expression_operations.cc
```diff
@@ -1369,8 +1369,13 @@ bindUserFunction(SqlBindingScope & context) const
          }
          else
          {
-            throw HttpReturnException(400, "User function " + functionName
-                                   + " expect a row argument ({ }), got " + args[0]->print() );
+            auto functionresult = std::dynamic_pointer_cast<FunctionCallWrapper>(args[0]);
```
guyd: Don't we need something similar for built-in functions?

guyd:
+1

sql/sql_expression_operations.cc
steadtler: Not for the moment no

testing/MLDB-724-time-arithmetic.js
```diff
@@ -359,4 +359,10 @@ assertEqual(query.json[0].columns[0][1]['ts'], "2015-01-03T00:00:00Z");
 assertEqual(query.json[0].columns[1][1]['ts'], "2015-01-01T00:00:00Z");
 assertEqual(query.json[0].columns[2][1]['ts'], "2015-01-03T00:00:00Z");
 
+//MLDB-1230
+
+query = mldb.get('/v1/datasets/test3/query',
+                      { select: 'CAST (INTERVAL \'1s\' as integer) as x'});
+plugin.log(query);
```
jeremybarnes: Shouldn't we assert on that?

sql/sql_expression_operations.cc
jeremybarnes: This smells bad.  The requirement is on the return type (that the expression should return a row), not on the structure of the expression tree.   Why can't we simply test for the output type of the expression rather than checking its actual node?

jeremybarnes:
I may be missing context, but this feels like a hack to me.  For example, user_fn1(identity(user_fn2(...))) should work just the same as user_fn1(user_fn2(...)) where identify is any expression or function that doesn't modify the row.  What am I missing?

steadtler:
@jeremybarnes, @nicolaskruchten 
Im just following the DoD. Before, we would accept ONLY rows that were **explicitly** expressed with curly braces { }. Now the DoD asked me to also accept the output of functions as long as the names match at binding, but not just any row.

nicolaskruchten:
I agree that there are cleaner/more general ways to do this, but right now the requirement is that we can chain user-defined functions. Realistically, there are no built-in functions which return rows which happen to be compatible with the inputs of user-defined functions right now, so covering the literal `{ }` case and the chaining case gets us most of what we need.

jeremybarnes:
OK.  I fear that we're creating a bit of a mess that we will need to clean up later, but how we prioritize expediency over technical debt is up to you.

testing/MLDB-724-time-arithmetic.js
steadtler: Oups! That doesnt belong in this branch.

PR #121, opened 2016-01-29
Cosmetic changes to graceful shutdown code

jraby:
- Use constant instead of magic number
 - Let the operator know mldb that the container is being shut down


nicolaskruchten:
+1

PR #123, opened 2016-01-29
mldb_base tweaks

jraby:
- Update mirrors url to use ec2
 - use lib{blas,lapack}3 instead of the gf transitional pkg
 - get rebuild_pycs.py through mldb/ symlink


nicolaskruchten:
+1

PR #124, opened 2016-01-29
Mldb 1332 integrate spdlog

guyd:
This is only adding the spdlog to our repo and build.  Integration with it will come as a separate PR.


steadtler:
+1, but beware of the risks of static-initialization order.

guyd:
@steadtler I know that there are no such order as static-initialization-oder!  Here it does not create any problems since there are no required order.

PR #125, opened 2016-01-29
Mldb 1335 no from segfault

guyd:
Added more coverage for the validation of the procedure configuration.  Namely, the from clause is now checked in any procedure supporting it and the test is more strict.  This is fixing a crash reported by Thomas earlier today.


FinchPowers:
+1

PR #126, opened 2016-01-29
changing default writeLevel of sparse.mutable MLDB-1339

nicolaskruchten:


jeremybarnes:
I would add a big warning to the documentation saying that the default setting will look like it's losing data.

nicolaskruchten:
Maybe I should rename the option consistencyLevel and the values
consistentAfterWrite vs consistentAfterCommit.

On Friday, 29 January 2016, Jeremy Barnes <notifications@github.com> wrote:

> I would add a big warning to the documentation saying that the default
> setting will look like it's losing data.
>
> —
> <https://github.com/mldbai/mldb/pull/126#issuecomment-176998262>.
>


-- 
http://datacratic.com/

nicolaskruchten:
@jeremybarnes I added the warning, and changed the option names to be more suggestive of their impact.

guyd:
+1

PR #127, opened 2016-02-01
MLDB-1307 FunctionExpressionContext is no longer readthrough

steadtler:
The only things that FunctionExpressionContext now makes available are:

-The input values
-Functions (from the server)
-Function stack depth for recursion detection


guyd:
This implementation seems a lot more aligned with the design and feel simpler.  I will let Jeremy give the OK.

jeremybarnes:
Looks good to me.  +1

PR #129, opened 2016-02-01
Mldbfb 331 sum excluding test

FinchPowers:


FinchPowers:
Bad query, should be
```
SELECT sum({* EXCLUDING (colB)}) AS * FROM example"
```
which works.

PR #131, opened 2016-02-01
Replaced custom unitted output with default one

FinchPowers:


nicolaskruchten:
+1 if this makes test output nicer.

PR #132, opened 2016-02-01
Mldb 1344 test enh

FinchPowers:


plugins/lang/python/python_loader.cc
```diff
@@ -613,6 +613,10 @@ class mldb_wrapper(object):
 
         def run_tests(self):
             import unittest
+            import StringIO
```
FinchPowers: Ignore this file. Part of PR #131

FinchPowers:
I added various test cases. I would update the DOD to
* remove manual
* remove unittest.expectedFailure

nicolaskruchten:
+1

PR #133, opened 2016-02-02
Mldb 1336 Better error message for function arguments

steadtler:
Note that the exception thrown should not specify the function name as it will be added when the exception is caught lower in the call stack.

### DIFF
--- a/sql/builtin_functions.cc
+++ b/sql/builtin_functions.cc
@@ -33,10 +33,10 @@ namespace {
     {
         if (number != expected)
         {
-            if (expected > 1)
-                throw HttpReturnException(400, "takes " + to_string(expected) + " arguments, got " + to_string(number));
+            if (expected != 1)
+                throw HttpReturnException(400, "expected " + to_string(expected) + " arguments, got " + to_string(number));
             else
-                throw HttpReturnException(400, "takes " + to_string(expected) + " argument, got " + to_string(number));
+                throw HttpReturnException(400, "expected " + to_string(expected) + " argument, got " + to_string(number));
         }
     }    
 }
@@ -1178,7 +1178,7 @@ BoundFunction token_extract(const std::vector<BoundSqlExpression> & args)
     // Comma separated list, first is row name, rest are row columns
 
     if (args.size() < 2)
-        throw HttpReturnException(400, "requires at least two argument");
+        throw HttpReturnException(400, "requires at least two arguments");
 
     if (args.size() > 3)
         throw HttpReturnException(400, "requires at most three arguments");
--- a/sql/builtin_functions.cc
+++ b/sql/builtin_functions.cc
@@ -28,6 +28,18 @@ namespace Datacratic {
 namespace MLDB {
 namespace Builtins {
 
+namespace {
+    void checkArgsSize(size_t number, size_t expected)
+    {
+        if (number != expected)
+        {
+            if (expected > 1)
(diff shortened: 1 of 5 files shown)


sql/builtin_functions.cc
```diff
@@ -28,6 +28,18 @@ namespace Datacratic {
 namespace MLDB {
 namespace Builtins {
 
+namespace {
+    void checkArgsSize(size_t number, size_t expected)
+    {
+        if (number != expected)
+        {
+            if (expected > 1)
```
guyd: expected != 1 so that we have _0 arguments_.

sql/builtin_functions.cc
```diff
@@ -1137,6 +1177,12 @@ BoundFunction token_extract(const std::vector<BoundSqlExpression> & args)
 {
     // Comma separated list, first is row name, rest are row columns
 
+    if (args.size() < 2)
+        throw HttpReturnException(400, "requires at least two argument");
```
guyd: argument -> arguments

guyd:
Lovely!  +1

testing/MLDB-1336-builtin-checks.py
```diff
@@ -0,0 +1,21 @@
...
+# This file is part of MLDB. Copyright 2016 Datacratic. All rights reserved.
+
+mldb = mldb_wrapper.wrap(mldb) # noqa
+
+import unittest
+
+class myTest(unittest.TestCase):
+
+    def test_sequence(self):
+        with self.assertRaisesRegexp(mldb_wrapper.ResponseException, "Executing builtin function exp: Can't convert value 'a' to double") as re:
```
nicolaskruchten: Message should be "Executing builtin function exp: Argument 1: Can't convert value 'a' to double"

nicolaskruchten: Actually, here is what would be ideal: "Executing builtin function exp: Argument 1: expected a double, got 'a' "

testing/MLDB-1336-builtin-checks.py
```diff
@@ -0,0 +1,21 @@
...
+            query = "SELECT exp('a')"
+            mldb.query(query)
+        with self.assertRaisesRegexp(mldb_wrapper.ResponseException, "Executing builtin function ln: Can't convert from row to double") as re:
+            query = "SELECT ln({1})"
+            mldb.query(query)
+        with self.assertRaisesRegexp(mldb_wrapper.ResponseException, "Binding builtin function sqrt: takes 1 argument, got 3") as re:
```
nicolaskruchten: should be: "Binding builtin function sqrt: expected 1 argument, got 3"

testing/MLDB-1336-builtin-checks.py
steadtler: Arg actually you were right, I can't do this without adding a try_catch to all builtins, or doing some complex registering of expected arguments types. Is it worth the effort at this point?

nicolaskruchten: It's not critical to identify which argument, no, but it is important to let the user know that the problem was with an argument. Would it make sense to change the "executing builtin function" to "arguments to function" or can that message cover any number of execution problems which just happen to include argument errors?

steadtler: Because its at execution, we wont know the problem is with an argument unless we put a try-catch within every builtin.

nicolaskruchten: All right.

nicolaskruchten:
+1

PR #135, opened 2016-02-02
python unittest assertQueryResylt

FinchPowers:


plugins/lang/python/python_loader.cc
```diff
@@ -634,6 +634,47 @@ class mldb_wrapper(object):
...
+        res = sorted(res, key=lambda(v): v['rowName'])
+        expected = sorted(expected, key=lambda(v): v['rowName'])
+        for res_row, expected_row in zip(res, expected):
+            self.assertEqual(res_row["rowName"], expected_row["rowName"])
+            res_columns = sorted(res_row["columns"])
+            expected_columns = sorted(expected_row["columns"])
+            self.assertEqual(res_columns, expected_columns)
+
+
+    def assertQueryResult(self, res, expected):
```
FinchPowers: This is the new function. Works both with `mldb.query` and `mldb.get("/v1/query", ...).json()`

testing/mldb_unit_test_test.py
```diff
@@ -0,0 +1,111 @@
...
+        res = mldb.query("SELECT * FROM ds")
+        expected = [["_rowName", "colC", "colA", "colD"],
+                    ["row2", None, 1, 1],
+                    ["row1", 1, None, 1]]
+        with self.assertRaises(AssertionError):
+            self.assertQueryResult(res, expected)
+
+    def test_object_valid(self):
+        res = mldb.get('/v1/query', q="SELECT * FROM ds").json()
+        expected = [{"rowName":"row1","columns":[["colA",1,"1970-01-01T00:00:00Z"],
```
FinchPowers: I can already think of a patch to accept int rather than dates so that `["colA",1,1]` instead of `["colA",1,"1970-01-01T00:00:00Z"]` would still work.

plugins/lang/python/python_loader.cc
nicolaskruchten: the json version is going to be very brittle, as we have many different output options: soa, aos, full, sparse etc... I would focus on the 'flat' version which works with format=table aka mldb.query

FinchPowers: For testing, the json version works with the default parameters so I like it. Plus, whenever we need to validate timestamps we kind of need it.

nicolaskruchten: OK. So this function will reorder both columns and rows? Maybe it should just reorder columns, and let the developer use order by?

FinchPowers: Sure I can do that too.

nicolaskruchten:
+1

PR #136, opened 2016-02-02
Added support for tensors to ExpressionValue (MLDB-1317)

jeremybarnes:
First cut at generalizing current embedding support across multiple dimensions and more data types.

Comments please with the goal of identifying a DOD and implementing it.

Note that this includes an expanded DOD to allow testing:
- Embedding literals are supported in multiple dimensions
- Unary and binary builting math functions can now be applied to embeddings or rows, which will do what you would expect
- Binary arithmetic operators can now be applied to embeddings or rows, which will also do what you would expect.


nicolaskruchten:
Not sure how to contribute here... Is the intent to be able to interact with tensors in queries? Are tensors and blobs different? Can we go from image-as-blob to tensor and back?

steadtler:
Nice! The tensors are going to be much useful. Unit tests upcoming for them? +1

sql/coord.cc
```diff
@@ -0,0 +1,385 @@
...
+    new (&str.str) Utf8String(std::move(other.getComplex()));
+}
+
+void
+Coord::
+initString(Utf8String str)
+{
+    ExcAssertEqual(strlen(str.rawData()), str.rawLength());
+    words[0] = words[1] = words[2] = words[3] = 0;
+    if (str.rawLength() <= 31) {
```
guyd: Shouldn't we have <= 30 and bytes + 2 so that we don't overwrite the simpleLen_ union member?

sql/coord.h
```diff
@@ -0,0 +1,262 @@
...
+    struct Str {
+        uint64_t md;
+        Utf8String str;
+        uint64_t savedHash;
+    };
+
+    union {
+        // The complex_ flag means we can't simply copy the words around;
+        // we need to do some more work.
+        struct { uint8_t complex_: 1; uint8_t simpleLen_:5; };
```
guyd: Why simpleLen_ is initialized to 5 here?

jeremybarnes: It's a number of bits of a bitfield value, not an initialization.

guyd: Oups!  I completely missed that!

sql/coord.cc
jeremybarnes: No, because they are bitfields they all fit into the first byte.

guyd: Yes.  I figured it out when my confusion about bitfields / initialization was cleared out.

guyd:
+1

steadtler:
Im ok with the latest changes +1, there is just some debug output left in it to comment.

guyd:
Briefly reviewed also.  +1

PR #138, opened 2016-02-02
MLDB-1313 delay resolution of wildcard ambiguity from parsing to binding

steadtler:
If you take SELECT x.*

instead of assuming it means all column from dataset x, we will wait until binding to see if it means that or all columns that starts with "x."

Add a new method to the SqlBindingScope interface to resolve table names in identifier.

### DIFF
--- a/server/dataset_context.h
+++ b/server/dataset_context.h
@@ -109,12 +109,16 @@ struct SqlExpressionDatasetContext: public SqlExpressionMldbContext {
     {
         return RowContext(row, params);
     }
+
+    virtual Utf8String 
+    doResolveTableName(const Utf8String & fullVariableName, Utf8String &tableName) const;
     
 protected:
 
     Utf8String removeTableName(const Utf8String & variableName) const;
     Utf8String removeQuotes(const Utf8String & variableName) const;
     Utf8String resolveTableName(const Utf8String& variableName) const;
+    Utf8String resolveTableName(const Utf8String& variableName, Utf8String& resolvedTableName) const;
 };
 
 
--- a/sql/sql_expression.cc
+++ b/sql/sql_expression.cc
@@ -394,7 +394,15 @@ SqlBindingScope::
 doGetTable(const Utf8String & tableName)
 {
     throw HttpReturnException(400, "Binding context " + ML::type_name(*this)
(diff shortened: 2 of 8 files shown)


server/dataset_context.h
```diff
@@ -109,12 +109,16 @@ struct SqlExpressionDatasetContext: public SqlExpressionMldbContext {
     {
         return RowContext(row, params);
     }
+
+    virtual Utf8String 
+    doResolveTableName(const Utf8String & fullVariableName, Utf8String &tableName) const;
```
guyd: It seems that the naming of the function is incorrect.  Are we trying to resolve a variable name or a table name?

sql/sql_expression.cc
```diff
@@ -2019,65 +2013,6 @@ parse(ML::Parse_Context & context, bool allowUtf8)
         }
     }
 
-    //cerr << "offset = " << context.get_offset() << endl;
-    /* Three possibilities:
-       1.  (tablename).(prefix)*
-       2.  expression (AS label)
-       3.  label : expression
-    */
-    if (!matched) {
```
guyd: What's happening with that segment of code?  It was never executed?

sql/sql_expression_operations.cc
```diff
@@ -2330,18 +2330,24 @@ WildcardExpression(Utf8String tableName,
...
       excluding(excluding)
 {
+
 }
 
 BoundSqlExpression
 WildcardExpression::
 bind(SqlBindingScope & context) const
 {
+    Utf8String simplifiedprefix = prefix;
```
guyd: simplifiedprefix -> simplifiedPrefix

server/dataset_context.h
steadtler: Well, both as it were. We have a variable name that may or may not contain a table name.

sql/sql_expression.cc
steadtler: That code was there to find the explicit dataset.* case before we handle the more generic case identifier*  Now we can't handle the explicit case at parsing time, so the code isnt needed anymore.

jeremybarnes:
I agree very strongly with that change.  I'm surprised it was such a small amount of code to change.  +1 from me.

PR #140, opened 2016-02-02
MLDBFB-335 when timestamp variable test

FinchPowers:


testing/MLDBFB-335_when_timestamp_variable_test.py
```diff
@@ -0,0 +1,47 @@
...
+
+    @unittest.expectedFailure
+    def test_it(self):
+        # t is indeed equal to the stored timestamp
+        t = self.__class__.t
+        res = mldb.query("SELECT time FROM timeDs")
+        self.assertEqual(res[1][1], t)
+
+        # but fetching via a value vs a variable containing the same value
+        # doesn't work
```
guyd: Not sure of what does not work here?

FinchPowers: `res_value` holds the results filtered by "t"
`res_variable` holds the full dataset as if the variable time was positive infinite. Since time == t, I expect those result sets to be equal.

nicolaskruchten:
+1

FinchPowers:
Already merged from e07dbc036a8908ced73c99f7479443c1b3278fba

PR #141, opened 2016-02-02
Mldbfb 192 row name as string test

FinchPowers:


testing/MLDBFB-192_row_name_as_string_test.py
```diff
@@ -0,0 +1,47 @@
...
+import unittest
+mldb = mldb_wrapper.wrap(mldb) # noqa
+
+class RowNameAsStringTest(MldbUnitTest):
+    query = "SELECT * FROM csv ORDER BY rowName()"
+
+    @classmethod
+    def setUpClass(cls):
+        tmp_file = tempfile.NamedTemporaryFile(
+            prefix=os.getcwd() + '/build/x86_64/tmp/')
```
guyd: This will keep adding new files in tmp as we rerun the test.  Consider using a fixed name containing the JIRA number.

testing/MLDBFB-335_when_timestamp_variable_test.py
```diff
@@ -0,0 +1,47 @@
...
+        })
+        ds.record_row('row1', [['time', cls.t, 0]])
+        ds.commit()
+
+    @unittest.expectedFailure
+    def test_it(self):
+        # t is indeed equal to the stored timestamp
+        t = self.__class__.t
+        res = mldb.query("SELECT time FROM timeDs")
+        self.assertEqual(res[1][1], t)
```
guyd: I probably missing one check-in, calling json() is now optional?

guyd:
+1

testing/MLDBFB-192_row_name_as_string_test.py
FinchPowers: Not at all! tempfile self destructs at program exit!

testing/MLDBFB-335_when_timestamp_variable_test.py
FinchPowers: `mldb.query` -> returns the array directly
`mldb.get('/v1/query', ...)` -> returns a response objects upon which you need to call `.json()` to get the body parsed as json.

testing/MLDBFB-192_row_name_as_string_test.py
```diff
@@ -0,0 +1,47 @@
...
+        self.assertQueryResult(res, [
+            ['_rowName', 'header'],
+            ['2', 'val1'],
+            ['3', 'val2']
+        ])
+
+    @unittest.expectedFailure
+    def test_object_result(self):
+        res = mldb.get('/v1/query', q=self.__class__.query).json()
+        # because of the now ts beign inserted we only check for row names
```
nicolaskruchten: where is the now ts being inserted?

nicolaskruchten:
+1

testing/MLDBFB-192_row_name_as_string_test.py
guyd: Great.  I thought that it only offered a unique filename and that deletion had to be done manually.

testing/MLDBFB-192_row_name_as_string_test.py
FinchPowers: When we load from a csv, the timstamps given to the cell values is "now".

nicolaskruchten: apparently it's the timestamp of the file, and it can be overridden with the 'timestamp' parameter, in case you want to refactor a bit: http://mldb.ai/doc/#builtin/datasets/TabularCsvDataset.md.html

PR #145, opened 2016-02-04
Mldb 1235 temporal agg

guyd:
The end goal of this PR was to implement the temporal_min function which operate on the time dimension to find the earliest occurrence of an event.  There was a fair amount of redesign required for this to work.

1. BoundSqlExpression was modified to accept the time filter. It was always set to GET_LATEST.  The filter is now accepted by all bound expressions.  Currently only `temporal_min` executes bound expressions (its argument) with a filter that is not the default.
2. Functions had to be changed to bind their arguments and execute/evaluate them.  As a result, all the built-in functions have changed signature and perform the extra work of evaluating their arguments.  This has the benefit of simplifying the binding in the FunctionCallWrapper and removing the need to rebind arguments in the ReadThroughBindingContext.
3. The PipelineExpressionScope has also being modified to not derive from a ReadThroughBindingContext.  The scope still owns an outer context to which it delegates most of the lock-up actions (getting functions, variables and datasets) but could no longer be a ReadThroughBindingContext because it prevented arguments from being evaluated with the same context and the function itself.


steadtler:
I dont mind the changes, Im just wondering if it would have been easier to allow BoundFunction to support two Exec signatures?

guyd:
Yes.  It would have been easier (I just thought of it too late!).  This would have minimized the changes in builtin_function.  We could certainly revisit and make that improvement.

guyd:
Regarding the performance impact, the changes seem to be neutral.  Here are some numbers with and without the changes:
query | without the change | with the change| 
------------ | ------------- | --------------|
select 1 | 164 223 | 151 441 |
select lower('WUP') | 135 498 | 139 590 |
select regexp_replace(lower('WUP WAP'), 'wap', 'wup') | 52 394 | 51 632 |
All differences seems to be in the margin of errors.

testing/MLDB-1235-temporal-aggregators.py
```diff
@@ -0,0 +1,64 @@
...
+        self.assertQueryResult(
+            mldb.query('select temporal_min(x) as t_min_x from dataset order by rowName()'),
+            [
+                ["_rowName", "t_min_x"],
+                ["row_0",  0 ],
+                ["row_1",  -1 ],
+                ["row_2",  -2 ],
+                ["row_3",  -3 ],
+                ["row_4",  -4 ]
+            ]
```
FinchPowers: Il serait préférable de tester le retour «complet», donc avoir les timestamps attachés aux valeurs pour valider que c'est bien yesterday dans tous les cas.

testing/MLDB-1235-temporal-aggregators.py
```diff
@@ -0,0 +1,64 @@
...
+        )
+
+
```
FinchPowers: Un autre test pour `temportal_min({*})`? C'est ce qui va m'intéresser. :)

guyd:
I've evaluated the idea proposed by Mathieu, that is, support two signatures in BoundFunction.  This brings other complications where several places in the code have to choose between the right signature to call.  So I did not pursue this approach.

However, I changed the builtin functions to return a _ValuedBoundFunction_ to lower the number of changes to these functions.  I am not convince that this is simpler in the end but please review and provide your comments.

guyd:
Still waiting for a +1 on that one.

PR #146, opened 2016-02-04
Mldb 1361 join rowname subselect

steadtler:
Fix issue when selecting column from join dataset with unmatched outer/left/right rows

### DIFF
--- a/testing/MLDB-1361_join_on_subselect.py
+++ b/testing/MLDB-1361_join_on_subselect.py
@@ -1,0 +1,59 @@
+#
+# MLDB-1361_join_on_subselect.py
+# Francois Maillet, 2016-02-04
+# This file is part of MLDB. Copyright 2016 Datacratic. All rights reserved.
+#
+
+import unittest
+
+mldb = mldb_wrapper.wrap(mldb) # noqa
+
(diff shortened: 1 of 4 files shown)


testing/MLDB-1361_join_on_subselect.py
```diff
@@ -0,0 +1,59 @@
...
+        
+        ds = mldb.create_dataset({ "id": "sub1", "type": "sparse.mutable" })
+        ds.record_row("row_a",[["warp", 8, 0]])
+        ds.commit()
+        
+        ds = mldb.create_dataset({ "id": "sub2", "type": "sparse.mutable" })
+        ds.record_row("row_b",[["warp", 9, 0]])
+        ds.commit()
+
+    def test_select_x_works(self):
```
guyd: What happen when two tests have the same name?

guyd:
+1

testing/MLDB-1361_join_on_subselect.py
steadtler: Apparently some pressured data scientist made a mistake with the test !

nicolaskruchten: the short answer is that one of the tests will not run

PR #147, opened 2016-02-04
Made SVD work OK in the case of real-valued columns with no variation in training (MLDB-687)

jeremybarnes:


steadtler:
I dont know the svd implementation that much but this seems ok +1

PR #148, opened 2016-02-04
Mldb 1361 join rowname subselect

steadtler:
-Fix issue with accessing columns from an outer Join
-Fix issue with using functions with arguments in Join expression

### DIFF
--- a/server/function_contexts.cc
+++ b/server/function_contexts.cc
@@ -283,7 +283,7 @@ doGetFunctionEntity(const Utf8String & functionName)
 }
 
 Utf8String 
-SqlBindingScope::
+FunctionExpressionContext::
 doResolveTableName(const Utf8String & fullVariableName, Utf8String &tableName) const
 {
     return fullVariableName;
--- a/testing/MLDB-1361_join_on_subselect.py
+++ b/testing/MLDB-1361_join_on_subselect.py
@@ -32,19 +32,62 @@ class SampleTest(MldbUnitTest):
         rez = mldb.get("/v1/query", q="""
 
             SELECT
-                text.txt, tbl1.warp, tbl2.warp
+                text.txt, sub1.warp, sub2.warp
             FROM text
             LEFT JOIN sub1 ON text.rowName() = regex_replace(sub1.rowName(), 'row_', '')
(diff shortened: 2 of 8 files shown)


server/function_contexts.cc
```diff
@@ -283,7 +283,7 @@ doGetFunctionEntity(const Utf8String & functionName)
 }
 
 Utf8String 
-SqlBindingScope::
```
guyd: The compiler did not complain!

testing/MLDB-1361_join_on_subselect.py
```diff
@@ -0,0 +1,155 @@
...
+        ds.record_row("row_a",[["warp", 8, 0]])
+        ds.commit()
+        
+        ds = mldb.create_dataset({ "id": "sub2", "type": "sparse.mutable" })
+        ds.record_row("row_b",[["warp", 9, 0]])
+        ds.commit()
+
+    def test_select_x_works(self):
+        # mldb.get asserts the result status_code is >= 200 and < 400
+        rez = mldb.get("/v1/query", q="""
```
guyd: assertQueryResult would make this a little simpler and would have a much better assert message.

testing/MLDB-1361_join_on_subselect.py
```diff
@@ -0,0 +1,155 @@
...
+        
+        ds = mldb.create_dataset({ "id": "sub1", "type": "sparse.mutable" })
+        ds.record_row("row_a",[["warp", 8, 0]])
+        ds.commit()
+        
+        ds = mldb.create_dataset({ "id": "sub2", "type": "sparse.mutable" })
+        ds.record_row("row_b",[["warp", 9, 0]])
+        ds.commit()
+
+    def test_select_x_works(self):
```
guyd: May be picking a more representation name would help also in finding what is failing.

guyd:
Not too sure about what the transform part does but otherwise the rest is fine.  +1

testing/MLDB-1361_join_on_subselect.py
steadtler: I was having issues with assertQueryResult and this was urgent. Will try again tomorrow

PR #151, opened 2016-02-05
Mldb 1344 aggregator in transform

steadtler:
Fix transform procedure not working when there are aggregators without a group by.

### DIFF
--- a/plugins/sql_functions.cc
+++ b/plugins/sql_functions.cc
@@ -535,7 +535,7 @@ run(const ProcedureRunConfig & run,
     bool skipEmptyRows = procedureConfig.skipEmptyRows;
 
     // Run it
-    if (procedureConfig.inputData.stm->groupBy.clauses.empty()) {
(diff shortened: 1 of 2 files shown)


plugins/sql_functions.cc
```diff
@@ -535,7 +535,7 @@ run(const ProcedureRunConfig & run,
     bool skipEmptyRows = procedureConfig.skipEmptyRows;
 
     // Run it
-    if (procedureConfig.inputData.stm->groupBy.clauses.empty()) {
+    if (procedureConfig.inputData.stm->groupBy.clauses.empty() && aggregators.empty()) {
```
FinchPowers: Quoi c'est tout!?! :p

FinchPowers:
+1

plugins/sql_functions.cc
steadtler: Heh.

PR #152, opened 2016-02-05
Reduced ExpressionValue useless padding (MLDB-1363)

jeremybarnes:
Removes some useless padding in the ExpressionValue class, which should reduce overall memory usage somewhat in queries.

Very trivial.


steadtler:
+1

PR #154, opened 2016-02-08
MLDB-1362 workaround for boost::date_time MT issue

steadtler:
Workaround for the issue that @mailletf encountered when calling date_part in MT

### DIFF
--- a/server/mldb_server.cc
+++ b/server/mldb_server.cc
@@ -113,9 +113,19 @@ init(std::string configurationPath,
 {
     auto server = std::make_shared<StandalonePeerServer>();
 
+    preInit();
     initServer(server);
     initRoutes();
     initCollections(configurationPath, staticFilesPath, staticDocPath, hideInternalEntities);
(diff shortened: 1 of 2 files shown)


server/mldb_server.cc
```diff
@@ -120,6 +121,15 @@ init(std::string configurationPath,
 
 void
 MldbServer::
+preInit()
+{
+    //Because of a multithread issue in boost, we need to call this to force boost::date_time to initialize in single thread
```
guyd: Possibly add the bug number for easy lookup.

guyd:
+1

server/mldb_server.cc
wsourdeau: Technically speaking, if multiple threads were instantiating an MldbServer simultaneously, the race condition would occur too. One way to guarantee that it does not occur is to call this during the initialisation of the program, possibly via the initialisation of a global variable.

PR #155, opened 2016-02-08
Mldb 1359 proc lastestrun

FinchPowers:
Pending on the UTF-8 in url decision.


server/procedure_collection.cc
```diff
@@ -99,6 +100,45 @@ initRoutes(RouteManager & manager)
...
+        Utf8String winningKey("");
+        for (const auto & key: keys) {
+            auto * currentProc = runs->getExistingEntry(key).get();
+            if (latestRunDate < currentProc->runStarted) {
+                latestRunDate = currentProc->runStarted;
+                winningKey = key;
+            }
+        }
+        redirect += "/runs/" + winningKey;
+        connection.sendRedirect(302, redirect.rawString());
```
FinchPowers: ^^ Pending on the utf-8 in url decision.

plugins/lang/python/python_loader.cc
```diff
@@ -567,11 +570,28 @@ class mldb_wrapper(object):
...
+                                             counter - 1)
+            return response
+
         def _perform(self, method, url, *args, **kwargs):
             raw_res = self._mldb.perform(method, url, *args, **kwargs)
             response = mldb_wrapper.Response(url, raw_res)
             if response.status_code < 200 or response.status_code >= 400:
                 raise mldb_wrapper.ResponseException(response)
+            if response.status_code >= 300:
+                return self._follow_redirect(response.headers['location'], 10)
```
FinchPowers: This is here because the auto redirect portion of the python wrapper wasn't done yet and the attached test covers it.

FinchPowers:
Ready for review.

plugins/lang/python/python_loader.cc
guyd: Consider putting a comment to explain the magic number _10_.

server/procedure_collection.cc
```diff
@@ -99,6 +100,45 @@ initRoutes(RouteManager & manager)
...
+
+  RestRequestRouter::OnProcessRequest getLatestRun =
+      [=] (RestConnection & connection, const RestRequest & req,
+           const RestRequestParsingContext & context)
+    {
+        // load current procedure
+        Utf8String redirect("/v1/procedures/");
+        auto collection = manager.getCollection(context);
+        auto key = manager.getKey(context);
+        auto procedure = static_cast<Procedure *>(
```
guyd: Do we need a static_cast here?  Could it be replaced with a dynamic_cast?

FinchPowers: @guyd static_cast is faster than dynamic_cast. 
static_cast is when you know what you are doing. If it fails, it crashes.
dynamic_cast is when you are not sure. If it fails, it retuns a nullptr.

FinchPowers: +1 for 404 not found :)

server/procedure_collection.cc
```diff
@@ -99,6 +100,45 @@ initRoutes(RouteManager & manager)
...
+
+        // load all runs
+        auto runs = procedure->runs.get();
+        auto keys = runs->getKeys();
+        if (keys.empty()) {
+            throw HttpReturnException(404, "not found");
+        }
+
+        // Find the latest run
+        Date latestRunDate = Date::negativeInfinity();
```
guyd: One thing to clarify, as implemented the latestrun is the one that started last not the one that ended last.  As a consequence, the latest run might still be running.  Is this ok with the rest of the code?

server/procedure_collection.cc
```diff
@@ -99,6 +100,45 @@ initRoutes(RouteManager & manager)
...
+        Utf8String winningKey("");
+        for (const auto & key: keys) {
+            auto * currentProc = runs->getExistingEntry(key).get();
+            if (latestRunDate < currentProc->runStarted) {
+                latestRunDate = currentProc->runStarted;
+                winningKey = key;
+            }
+        }
+        redirect += "/runs/" + encodeUriComponent(restEncode(winningKey));
+        connection.sendRedirect(302, redirect.rawString());
```
guyd: Wouldn't 307 be more accurate?

testing/MLDB-1359_procedure_latest_run.py
```diff
@@ -0,0 +1,59 @@
...
+# Mich, 2016-02-05
+# This file is part of MLDB. Copyright 2016 Datacratic. All rights reserved.
+#
+import time
+from dateutil import parser as date_parser
+
+mldb = mldb_wrapper.wrap(mldb) # noqa
+
+
+class ProcedureLatetRunTest(MldbUnitTest): # noqa
```
guyd: Can you had error case tests 
```
PUT /v1/procedures/proc/latestrun
```
and one for a proc that do not exist?

testing/MLDB-1359_procedure_latest_run.py
```diff
@@ -0,0 +1,59 @@
...
+                    'id' : 'dsOut'
+                },
+                'runOnCreation' : True
+            }
+        })
+
+        res = mldb.get(url + '/latestrun').json()
+        run_date = date_parser.parse(res['runStarted'])
+
+        time.sleep(0.01)
```
guyd: Why is this necessary?  The rest calls are sync unless specified otherwise.

testing/MLDB-1359_procedure_latest_run.py
```diff
@@ -0,0 +1,59 @@
...
+#
+import time
+from dateutil import parser as date_parser
+
+mldb = mldb_wrapper.wrap(mldb) # noqa
+
+
+class ProcedureLatetRunTest(MldbUnitTest): # noqa
+
+    def test_it(self):
```
guyd: I thought we were going away from monolithic tests.  Consider splitting this tests into two or three?

guyd:
+1 once the tests have been augmented with the two mentioned cases.

server/procedure_collection.cc
FinchPowers: @nicolaskruchten The code returns the latest based on "started on". Is it ok?

server/procedure_collection.cc
FinchPowers: J'y ai pensé, mais mon DOD spécifie 302. @nicolaskruchten plz confirm? :)

testing/MLDB-1359_procedure_latest_run.py
FinchPowers: I can, but it will 404 on the procedure itself.

testing/MLDB-1359_procedure_latest_run.py
FinchPowers: I want to make sure that whenf the system becomes lightning fast the auto generated id are still sortable by time.

server/procedure_collection.cc
nicolaskruchten: this is fine.

server/procedure_collection.cc
FinchPowers: Test added. It already works as is.

server/procedure_collection.cc
nicolaskruchten: sure, 307 works.

FinchPowers:
Pending master branch fix.

testing/MLDB-1359_procedure_latest_run.py
guyd: Ah right.  The validation is done before calling the latestrun endpoint.  This is why the code does not segfault in the static cast.  No need to add a test then.

testing/MLDB-1359_procedure_latest_run.py
guyd: Got it.

PR #156, opened 2016-02-08
MLDB-1350 validate row and column names in sparse dataset

steadtler:


guyd:
+1 for the empty rowname and columnname part.  We should create another JIRA for the numerical rownames.

PR #157, opened 2016-02-08
Mldb 1338 tokenize number

steadtler:
Allow 'value' argument to be of any atomic type


mailletf:
+1

sql/builtin_functions.cc
```diff
@@ -1676,8 +1676,11 @@ BoundFunction tokenize(const std::vector<BoundSqlExpression> & args)
                 while (it != bagOfWords.end()) {
                     if (check[4]) //values
                     {
+                        if (!values.isAtom())
+                          throw HttpReturnException(400, "requires 'value' argument be an atomic value, got type " + values.getTypeAsUtf8String());
```
guyd: atomic -> scalar (I believe this is what we have used elsewhere)

PR #158, opened 2016-02-08
Mldb 1365 py wrapper enh

FinchPowers:


plugins/lang/python/python_loader.cc
```diff
@@ -517,7 +517,10 @@ class mldb_wrapper(object):
     def wrap(mldb):
         return Wrapped(mldb)
 
-    class ResponseException(Exception):
+    class MldbBaseException(Exception):
```
FinchPowers: Will be applied to PR TooManyRedirectsException from https://github.com/mldbai/mldb/pull/155/files

plugins/lang/python/python_loader.cc
```diff
@@ -634,16 +637,62 @@ class mldb_wrapper(object):
...
-        if expected[0][0] != '_rowName':
-            raise Exception("The first row of expected needs to be the header")
+    import json
+
+    longMessage = True # Appends the user message to the normal message
+
+    class _AssertMldbRaisesContext(object):
+        """A context manager used to implement TestCase.assertRaises* methods.
+        Inspired from python unittests.
+        """
```
FinchPowers: This class was largely copied from python unittests and adapted.

plugins/lang/python/python_loader.cc
```diff
@@ -634,16 +637,62 @@ class mldb_wrapper(object):
...
+                            exc_value.response.status_code, self.status_code))
+            return True
+
+    def _get_base_msg(self, res, expected):
+        return '{line}Result: {res}{line}Expected: {expected}'.format(
+            line='\n' + '*' * 10 + '\n',
+            res=MldbUnitTest.json.dumps(res, indent=4),
+            expected=MldbUnitTest.json.dumps(expected, indent=4))
+
+    def assertTableResultEquals(self, res, expected, msg=""):
```
guyd: Pourquoi est-ce que l'on a besoin de deux asserts (Table et Full)? On ne pourrait pas comparer les deux jsons peu importe leur format.

FinchPowers: @nicolaskruchten trouvait que c'était trop _magique_ :dizzy:

nicolaskruchten: also, in table format there is no difference between "not part of the output" and "part of the output with null value"

nicolaskruchten:
+1

plugins/lang/python/python_loader.cc
guyd: Alors comment on test le output en format soa and aos?

FinchPowers: C'est quoi soa et aos?

guyd: On supporte plusieurs formats de output, table, full, sparse, soa et aos.  http://mldb.ai/doc/#builtin/sql/QueryAPI.md.html

FinchPowers: Il n'y a pas de _helper_ alors il faut les tester manuellement.

nicolaskruchten: basically yes. we have a special helper for format=table because that's the default output of mldb.query and we have a special helper for format=full because that's the most verbose one. Everything else can still be tested "the old way" but we can always add more helpers.

PR #161, opened 2016-02-09
MLDB-1367 id encoding / utf-8 support

FinchPowers:


rest/rest_request_params.cc
```diff
@@ -34,7 +35,14 @@ std::string restDecode(std::string str, std::string *)
...
 {
-    return std::move(str);
+    CURL *curl = curl_easy_init();
+    int outlength;
+    std::unique_ptr<char, void (*)(char *)> cres(
+        curl_easy_unescape(curl, str.rawData(), str.length(), &outlength),
+        [](char *p) { curl_free(p); });
+    Utf8String utf8(cres.get(), cres.get() + outlength);
+    curl_easy_cleanup(curl);
+    return std::move(utf8);
```
FinchPowers: 1. Je suppose que je devrais me faire un curl_easy_init que je partage à travers ce fichier plutôt que de le réinstantier/détruire à chaque fois? Est-ce qu'il y en a déjà un disponible que je devrais utiliser?
2. Si je réutilise le même, est-ce que quelqu'un sait si c'est thread safe?
3. J'ai laissé le std::move mais de ce que je comprend il ne sert à rien.

jeremybarnes: 1.  I would avoid using CURL if possible.
2.  I'm sure we're not looking at the right place here.  str is *already* a UTF-8 string, isn't it?  If we're sure it must be an ASCII encoded string, it should be a std::string; the fact it's already UTF-8 tells the user it's already been decoded.

1.  I think that curl_easy_init is already called elswhere; you should be able to grep it
2.  It's thread safe
3.  The std::move in the original version is useful, but not in your version, due to compiler optimizations for returning a value that's a local on the stack.

types/testing/decode_uri_test.cc
```diff
@@ -0,0 +1,105 @@
...
+#endif
+
+#if TEST_ALL
+BOOST_AUTO_TEST_CASE(test_utf8_repetitions)
+{
+    Utf8String in = "%C3%A9%C3%A9";
+    Utf8String expected = "éé";
+    BOOST_CHECK_EQUAL(Url::decodeUri(in), expected);
+}
+#endif
```
FinchPowers: Si on passe du input invalide, par exmpele `%C3`, ça throw `Invalid sequence within utf-8 string`. Est-ce correct?

types/url.cc
```diff
@@ -256,6 +257,73 @@ reconstitute(ML::DB::Store_Reader & store)
...
+            ++it; // over high
+            if (it == in.end() || !isxdigit(*it)) {
+                throw ML::Exception("Invalid uri"); // Todo better msg?
+            }
+            high = *it;
+            high -= high <= '9' ? '0' : (high <= 'F' ? 'A' : 'a') - 10;
+
+            ++it; // over low
+            if (it == in.end() || !isxdigit(*it)) {
+                throw ML::Exception("Invalid uri"); // Todo better msg?
```
FinchPowers: Est-ce que `Invalid uri` est suffisant comme message d'erreur?

nicolaskruchten: Badly encoded URI fragment: pata%%%ppp

types/testing/decode_uri_test.cc
FinchPowers: Nouvelle erreur.
`what:   Invalid encoding on uri fragment: %C3`

nicolaskruchten:
very good test coverage. +1

nicolaskruchten:
but out of curiosity: were we not able to re-use the existing query-string decoding code? Was that just because of the issue with "+" ?

FinchPowers:
I found no decoding code of any sort.

nicolaskruchten:
that's odd, because we already supported URL encoding in query-strings... might be worth checking to see if these can be combined.

nicolaskruchten:
i.e we could already send the query `select 'françois'` and it would decode and everything

types/url.cc
```diff
@@ -256,6 +257,85 @@ reconstitute(ML::DB::Store_Reader & store)
 }
 #endif
 
+/**
+ * Decodes uri encoded with Percent-encoding. It is meant to act like
+ * JavaScript decodeURI.
+ **/
+Utf8String
+Url::
+decodeUri(Utf8String in)
```
guyd: So we could not find a function that would do that?

FinchPowers: En fouillant pour encoder les uri, j'ai trouvé je crois dans la librairie google url. Les coquins ne nomment pas leurs affaires camel case donc c'est plus dur à trouver avec des greps.

PR #162, opened 2016-02-09
[MLDB-1231] removed auto-conversion of string literals into date

guyd:
I removed the code to parse strings into date.  This make the writing of the queries a little more difficult but this will be revisited later as part of MLDB-1357.

This is also fixing MLDB-1352.


steadtler:
Yay removing stuff + 1

PR #164, opened 2016-02-09
[MLDB-1351] transform run endpoint override config values

guyd:


steadtler:
+1

nicolaskruchten:
does this special code need to be present in all procedures in order for run-configuration-overriding to work?

guyd:
It does and it is as far I could see.

nicolaskruchten:
OK, so it was just missing for transform?

guyd:
I should have look deeper.  These are the procedures that are missing the call to merge run config with the config provided at creation:
- bucketize,
- word2vec,
- csv.export,
- xlsx.import,
- ranking, 
- create_function (python) and
- git

I will fix these as part of this JIRA unless we want to go a different path.

nicolaskruchten:
yes please :)

I thought it was automatic and that just this procedure was missing it somehow.

PR #165, opened 2016-02-09
Solve contention in stemming

steadtler:
Also add a couple of useful debug profiling macros I wrote while investigating.


guyd:
Interesting.  +1

PR #166, opened 2016-02-10
[MLDB-1351] run config override the config provided at creation in ma…

guyd:
…ny procedures

These were the procedures that were missing the call to merge the run config with the config provided at creation:

bucketize,
word2vec,
csv.export,
xlsx.import,
ranking,
create_function (python) and
git

I've changed all of these procedures so that runs override the config provided at creation except for the python procedure which does not have anything to override.


steadtler:
+1

PR #167, opened 2016-02-10
MLDB-1345 evaluate HAVING expressions

steadtler:
The HAVING was already part of the SQL statement and parsed and everything and even in the doc, so I only had to correctly evaluate it in the groupby query.

### DIFF
--- a/core/dataset.cc
+++ b/core/dataset.cc
@@ -551,8 +551,12 @@ queryStructured(const SelectExpression & select,
     std::mutex lock;
     std::vector<MatrixNamedRow> output;
 
+    if (!having.isConstantTrue() && groupBy.clauses.empty())
+        throw HttpReturnException(400, "HAVING expression requires a GROUP BY expression");
+
     std::vector< std::shared_ptr<SqlExpression> > aggregators = select.findAggregators();
-    
+    std::vector< std::shared_ptr<SqlExpression> > havingaggregators = having.findAggregators();
+
     // Do it ungrouped if possible
     if (groupBy.clauses.empty() && aggregators.empty()) {
         auto aggregator = [&] (NamedRowValue & row_,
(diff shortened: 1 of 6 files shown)


nicolaskruchten:
can you remove the 'manual' on the test case so we can see that it passes please? :)

core/dataset.cc
```diff
@@ -551,8 +551,12 @@ queryStructured(const SelectExpression & select,
     std::mutex lock;
     std::vector<MatrixNamedRow> output;
 
+    if (!having.isConstantTrue() && groupBy.clauses.empty())
+        throw HttpReturnException(400, "HAVING expression requires a GROUP BY expression");
```
jeremybarnes: Is that true?  What about something with an aggregator but no group by... is HAVING illegal there?

sql/sql_expression.cc
```diff
@@ -1740,6 +1747,46 @@ unimp(std::shared_ptr<SqlExpression> lhs,
...
+    std::vector<std::shared_ptr<SqlExpression> > output;
+    std::vector<std::shared_ptr<SqlExpression> > children = getChildren();
+
+    int index = 0;
+    while(index < children.size()) {
+        auto child = children[index];
+
+        bool foundAggregator = false;
+        if (child->getType() == "function") {
+            const FunctionCallWrapper * function = dynamic_cast<const FunctionCallWrapper *>(child.get());
```
jeremybarnes: Why is it called FunctionCallWrapper and not FunctionCallExpression?

jeremybarnes:
Looks OK to me.  +1 once Nic's comments are addressed.

core/dataset.cc
steadtler: According to the "internet" a having absolutely requires a group by. If you have aggregators but no group by, you get only 1 row anyway, so filtering that row has little use. Nic can overrule me here though...

sql/sql_expression.cc
steadtler: It was originally a wrapper around builtin function calls and user function calls back when the syntax and usage was more different. We could rename it.

PR #170, opened 2016-02-11
MLDB-1312 actually use quotechar parameter in CSV dataset and remove …

steadtler:
…text.line dataset


plugins/csv_dataset.cc
```diff
@@ -595,63 +597,84 @@ struct CsvDataset::Itl: public TabularDataStore {
         if (config.quoter.length() == 1) {
             quote = config.quoter[0];
         }
-        else {
+        else if (config.quoter.length() > 1) {
             throw HttpReturnException(400, "Quoter string must have one character");
         }
+        else if (config.delimiter.length() != 0)
```
nicolaskruchten: Ideally it would be possible for the quotechar to be empty even with a delimiter.

steadtler: Madness! Alright I'll see what I can do.

plugins/text_line_dataset.cc
```diff
@@ -1,297 +0,0 @@
-// This file is part of MLDB. Copyright 2015 Datacratic. All rights reserved.
-
-/** text_line_dataset.cc
```
nicolaskruchten: heads-up @mailletf this dataset is going away, so you will have to update the Sift notebooks if you rely on it. The replacement is text.csv.tabular with delimiter=quotechar=''.

Note #2: the text.csv.tabular dataset is queued up for being split up into a procedure and a dataset, so there's another change a'comin'.

plugins/csv_dataset.cc
```diff
@@ -580,78 +587,100 @@ struct CsvDataset::Itl: public TabularDataStore {
...
         if (config.quoter.length() == 1) {
             quote = config.quoter[0];
+            hasQuoteChar = true;
         }
-        else {
+        else if (config.quoter.length() > 1) {
             throw HttpReturnException(400, "Quoter string must have one character");
         }
 
+        const bool isTextLine = config.quoter.length() == 0 && config.delimiter.length() == 0;
```
FinchPowers: `config.quoter.length() == 0` pourait pas être `config.quoter.empty()` ?
Même chose pour `config.delimiter`.

plugins/csv_dataset.cc
```diff
@@ -580,78 +587,100 @@ struct CsvDataset::Itl: public TabularDataStore {
...
+                                           header.c_str(), header.length(), 1, 0);
+                
+                vector<string> fields = ML::expect_csv_row(pcontext, -1, separator);
 
-            switch (encoding) {
-            case ASCII:
-                for (auto & f: fields)
+                switch (encoding) {
+                case ASCII:
+                    for (auto & f: fields)
```
FinchPowers: Constness - Je ne suis pas certain, mais est-ce que ça pourrait être `const auto & f: fields` ?

FinchPowers:
Commentaires mineurs à vérifier.
+1

PR #171, opened 2016-02-11
Fixed up issues with row dataset literal (MLDB-1374)

jeremybarnes:
Adds extra test cases and fixes up issues with the row_dataset() function and the NAMED_COLUMNS mode of sql.query (which is basically its inverse).


steadtler:
Looks fine +1

PR #172, opened 2016-02-11
Fix incorrect number of row to treat per bucket

steadtler:
It now works like the other cases. Fixes segfault in some case where the last bucket would start out of bound.


guyd:
+1

PR #175, opened 2016-02-11
Further fixes to row_dataset (MLDB-1374)

jeremybarnes:
Fixes calling functions within row_dataset()


steadtler:
ok +1

sql/execution_pipeline.h
```diff
@@ -137,6 +137,11 @@ struct PipelineExpressionScope:
     virtual ColumnFunction
     doGetColumnFunction(const Utf8String & functionName);
 
+    // Function used to get a function.  Default throws an exception that the
```
guyd: This is probably the comment on SqlBindingScope.  Does not really apply here.

PR #177, opened 2016-02-11
Allow memory to be returned to OS (MLDBFB-329)

jeremybarnes:
Adds an option (enabled by default) to return memory to the OS by unmapping unused page ranges from tcmalloc.

### DIFF
--- a/core/dataset.cc
+++ b/core/dataset.cc
@@ -18,6 +18,7 @@
 #include "mldb/server/dataset_context.h"
 #include "mldb/server/per_thread_accumulator.h"
 #include "mldb/jml/utils/worker_task.h"
+#include "mldb/jml/utils/environment.h"
 #include "mldb/types/any_impl.h"
 #include "mldb/http/http_exception.h"
 #include "mldb/rest/rest_request_router.h"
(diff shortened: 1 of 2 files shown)


core/dataset.cc
```diff
@@ -328,9 +347,30 @@ Dataset(MldbServer * server)
...
 
+ML::Env_Option<int> RETURN_OS_MEMORY("RETURN_OS_MEMORY", 1);
+ML::Env_Option<int> PRINT_OS_MEMORY("PRINT_OS_MEMORY", 0);
+
+
 Dataset::
 ~Dataset()
 {
+    // MLDBFB-329
+    // Once a dataset is deleted, try to free its memory from the system
```
steadtler: Its going to free all unused pages, right? Not only those from the dataset?

jeremybarnes: yes

jeremybarnes:
@nicolaskruchten can we get a perspective on whether we should merge this, and if not what a DOD would look like to merge something similar?

nicolaskruchten:
Are there any downsides to this PR?

jeremybarnes:
The possible downside is that when freeing a dataset, it may cause extra work which could slow the system down if you're simultaneously predicting.

nicolaskruchten:
If that is the only downside then I vote to merge. +1

PR #180, opened 2016-02-15
MLDB-1382 upgrade python deps

jraby:
- All python deps are upgraded to their latest version.
- A few tweaks were required to match new behaviour.


nicolaskruchten:
+1

PR #181, opened 2016-02-16
MLDB-1162 html plugin

jeremybarnes:
This is a hangover from the pre-open sourced MLDB version.

Not to be merged until there is a clear DOD.

FYI @nicolaskruchten


nicolaskruchten:
So you want me to spec out some HTML parsing functions?

guyd:
Can we split the commits of the external libraries and those of the plugins so that we can review the relevant code more easily.

jeremybarnes:
Commits have been split into external and internal.

html/html_function.h
```diff
@@ -0,0 +1,48 @@
+/** tokensplit.h                                      -*- C++ -*-
```
guyd: Wrong header

html/html_function.h
```diff
@@ -0,0 +1,48 @@
...
+    Function to parse strings for tokens and insert separators
+*/
+
+#pragma once
+
+#include "mldb/server/function.h"
+#include "mldb/types/value_description_fwd.h"
+#include "mldb/types/optional.h"
+
+namespace Datacratic {
```
guyd: We should remove that namespace.

FinchPowers:
This PR is big enough to choke my browser. The part I saw had no test coverage of the new parser. Is that correct?

PR #183, opened 2016-02-16
[MLDB-1240] Replace the Favorite Recipe demo with an all-MLDB one, th…

simlmx:
Here are the main changes this new version of the Favorite Recipe demo:
- use the cosine distance throughout (before there was a mix of euclidean and cosine, whichever was the default in different places)
- not using the t-SNE anymore (it wasn't used)
- loading the recipe names in MLDB instead of just in python
- using the stemmer/filter_words/tf-idf to have score for words in a cluster
- 8 clusters instead of 20 (I though 20 made too many, and there were a bunch of redundant clusters)

What could be improved:
The word Chicken, because it is so much present, has a big TF score, and end up top in many word clouds. This didn't happen in the last version, because the score for words was a bit different. Maybe there is something that can be improved here. Also the words "i", "ii", etc. end up appearing in the word clouds.


guyd:
Look briefly at the notebook.  Text.line is no longer available.  See MLDB-1312.

nicolaskruchten:
I will fix up the text importation when I fix up all the demos :)

PR #185, opened 2016-02-17
MLDB-1354 merge unpack_json into parse_json

steadtler:
NK will update the doc afterwards.


mailletf:
Since this is breaking production deployments, I vote for a brief period where the old function name still works and is marked as deprecated, so there can be a smooth switch over.

nicolaskruchten:
DOD updated.

nicolaskruchten:
@steadtler can you add the `unpack_json` code back in plz? I'll add the documentation commit to this branch.

nicolaskruchten:
+1

PR #186, opened 2016-02-18
[MLDB-1380] returning a specific error message when the cpu does not …

guyd:
…support sse 4.2


server/test_arch.cc
```diff
@@ -0,0 +1,12 @@
+#include "mldb/arch/simd.h"
```
guyd: This file will not be committed.

server/mldb_server.cc
```diff
@@ -187,26 +188,49 @@ initRoutes()
...
-                                         true),
-                  RestParamDefault<bool>("rowNames",
-                                         "Do we include row names in output",
-                                         true),
-                  RestParamDefault<bool>("rowHashes",
-                                         "Do we include row hashes in output",
-                                         false));
+
+   // MLDB-1380 - make sure that the CPU support the minimal instruction sets
+    if (ML::has_sse42()) {
```
jeremybarnes: Consider splitting this out into a separate function like supportsSystemRequirements()

guyd: You mean the if condition or the whole if statement?

jeremybarnes: the condition.  +1 once that's done

PR #187, opened 2016-02-18
Fixed error message on loading file:// uri that doesn't exist (MLDB-1395)

jeremybarnes:
Very simple improvement to error message.


testing/MLDB-1395-error-message-file-doesnt-exist.js
```diff
@@ -0,0 +1,22 @@
...
+    mldb.log(expr, 'IS NOT EQUAL TO', val);
+
+    throw "Assertion failure";
+}
+
+var resp = mldb.post("/v1/datasets", { type: "text.csv.tabular", params: { dataFileUrl: "file://thisfiledoesnotexist" } });
+
+mldb.log(resp);
+
+assertEqual(resp.json.error, "Opening file ./thisfiledoesnotexist: failed opening file: No such file or directory");
```
guyd: The assertRaisesRegexp in python unit test does this check nicely.  See for example, MLDB-1336-builtin-checks.py.

guyd:
+1

testing/MLDB-1395-error-message-file-doesnt-exist.js
jeremybarnes: I would like to add some unit testing functions to the JS side in order to reduce boilerplate, but that's out of scope of this DOD :)

PR #188, opened 2016-02-18
Fixed type names not being found by demangled lookup (MLDB-1392)

jeremybarnes:
Fixes documentation error messages.  Looks like the root cause was that the typeName field of ValueDescription was inadvertently changed from the demangled to the non-demangled form.

Also adds a missing documentation block to the InputQuery, which is enabled by this change.


nicolaskruchten:
+1

PR #192, opened 2016-02-18
Use lock-free work stealing thread pool (MLDB-1173)

jeremybarnes:
This replaces the venerable ML::Worker_Task with a modern lock-free work stealing pool, and improves the management of thread groups and work grouping.

In particular, groups of work should simply create a new thread pool, which will by default forward jobs to be run on the global or parent thread pool.  This will also allow for an individual thread pool to limit its parallelism, for cases where we know that otherwise there will be too much lock contention.

It allows an order of magnitude more fine-grained work, and should allow us to deal with some of the places where we've artificially limited parallelism to avoid lock contention on the worker threads.

We also replace ML::run_in_parallel... with Datacratic::parallelMap...

This is a first release, which simply replaces the old functionality with the new.  None of the potential new features are used, yet.

It also removes 700 lines of code as compared to the old version.


guyd:
Briefly reviewed.  +700!

PR #195, opened 2016-02-21
Fixed termination in CSV loading exception (MLDB-1404)

jeremybarnes:
Removes the exit with std::terminate when an error occurs in CSV parsing from a worker thread.


nicolaskruchten:
@steadtler let's make sure this changes makes it into your import.text PR!

guyd:
+1 once the failing test is passing.

jeremybarnes:
The failing testcase has nothing to do with this PR; it's due to rate limiting of an external file that one of the other tests uses.

PR #198, opened 2016-02-22
Mldb 1235 temporal agg

guyd:
This is a first step into supporting the time dimension in more built-in functions and SQL expression in general.  In this PR:
- explicitly force GET_LATEST in place where getting many value for a column does not make sense
(WHERE clause, arithmetic with columns, ORDER BY clause, etc..).  To minimize the amount of changes, there are places where I've used GET_LATEST knowing that at some point we will want to support other variable filter
- modified ExpressionValue::getFiltered to return the columns in their _natural_ order
- fixed an inconsistency in min_timestamp
- remove some copying of ExpressionValues


builtin/joined_dataset.cc
```diff
@@ -195,7 +195,7 @@ struct JoinedDataset::Itl
         auto boundConstant = condition.constantWhere->bind(context);
         SqlRowScope scope;
         ExpressionValue storage;
-        if (!boundConstant(scope, storage).isTrue())
+        if (!boundConstant(scope, storage, GET_LATEST).isTrue())
```
steadtler: Maybe use a default value, since GET_LATEST is the default behavior?

server/bound_queries.cc
```diff
@@ -1242,7 +1242,7 @@ struct GroupContext: public SqlExpressionDatasetContext {
                     auto & row = static_cast<const RowContext &>(context);
              
                     const ExpressionValue * result
-                        = searchRow(row.output.columns, columnName, filter, storage);
+                        = searchRow(row.output.columns, columnName, filter == GET_ALL ? GET_LATEST : filter, storage);
```
steadtler: maybe a comment or a //todo in those places :)

steadtler:
a couple of remarks, but the code looks good.

server/bound_queries.cc
guyd: Actually this likely no longer necessary.  Let me check.

builtin/joined_dataset.cc
guyd: I agree that this is annoying for this review but I felt that not having a default was helpful.  Now one has to make it explicit when executing a bound expression that it should return one value or all values.   Said differently, I feel that GET_LATEST is not a better default than GET_ALL (it just appears more often).

steadtler:
+1

PR #199, opened 2016-02-22
Added test-case for plugin implemented in multiple libraries (MLDB-1398)

jeremybarnes:
This test-case was implemented to reproduce an issue in plugin loading.  In the end, the issue wasn't reproducible, but we should keep the test-case to ensure there isn't a regression.

This only changes tests.


jeremybarnes:
Again, the test failure is spurious.

jeremybarnes:
I also added documentation on how to avoid plugin loading errors.

steadtler:
+1

PR #200, opened 2016-02-22
Mldb 1163 split csv

steadtler:
move CSV loading code from dataset to procedure.
csv dataset replaced by generic tabular dataset.

Note that almost all changes is code moving around, there isnt a lot of new code.


nicolaskruchten:
@steadtler  were you able to incorporate the changes from #195 ?

steadtler:
In progress

plugins/importtext_procedure.cc
```diff
@@ -0,0 +1,1154 @@
+/** importextprocedure.cc
```
jeremybarnes: import_text_procedure?

jeremybarnes: Also, this only works for importing tabular files, not sparse ones.  So maybe the filename should mention tabular.

plugins/importtext_procedure.cc
```diff
@@ -0,0 +1,1154 @@
...
+    addField("named", &ImportTextConfig::named,
+             "Row name expression for output dataset. Note that each row "
+             "must have a unique name.",
+             SqlExpression::parse("lineNumber()"));
+    addField("timestamp", &ImportTextConfig::timestamp,
+             "Expression for row timestamp.",
+             SqlExpression::parse("fileTimestamp()"));
+
+    addParent<ProcedureConfig>();    
+    onUnknownField = [] (ImportTextConfig * config,
```
jeremybarnes: This can probably be removed now.  It was added to help with migration, but by definition everything that was using this is being migrated.

jeremybarnes:
This should be rebased/squashed before merging.  +1

plugins/importtext_procedure.cc
nicolaskruchten: let's leave the name as-is: we will be generalizing this further :)

nicolaskruchten:
Let's get this squash-merged in! :)

steadtler:
squashed into another PR

PR #202, opened 2016-02-23
Compile with clang, gcc5 (MLDB-1405)

jeremybarnes:
Fixes the warnings that come from the other compilers and make MLDB compile with any of them.  It found some consequential bugs including those in MLDB-1409, MLDB-1410 and that in binding_contexts.cc.


steadtler:
+1

PR #204, opened 2016-02-23
utf8 decoding fix

FinchPowers:


types/url.cc
```diff
@@ -273,21 +273,20 @@ decodeUri(Utf8String in)
     char buffer[output.length() * 4 + 1]; // prepare for the worse, 4 char + \0
     ssize_t index = 0;
     for (ssize_t i = 0; i < output.length(); ++i) {
-        char16 c = data[i];
+        char32_t c = data[i];
```
jeremybarnes: This is still going to be 16 bits as initialized here, of course.  You're putting it to 32 bits  to allow arithmetic later on?

jeremybarnes:
It looks good to me.  You'll need to undo my hack for MLDB-1405 to merge.  +1

types/url.cc
FinchPowers: Exactly

PR #211, opened 2016-02-25
[MLDB-1389] added a test to ensure there is no crash when training wi…

guyd:
…th no features


guyd:
This is only adding a new test in hope we can catch the crash mentioned in MLDB-1389. 

Merging right away.

PR #212, opened 2016-02-25
Mldb 1399 exp proc testset valid

guyd:
Added the ability to chain several configuration field validators and added a specialization of the existing validators for optional type.

Note that the validator code is growing more complex.  I will think of a refactoring to simplify it.


plugins/sql_config_validator.h
```diff
@@ -98,6 +111,14 @@ template<typename FieldType> struct NoGroupByHaving
     }
 };
 
+template<typename FieldType> struct NoGroupByHaving<Optional<FieldType> >
```
mailletf: This is more of a c++ question I guess but can't the Optional be templated away in the other NoGroupByHaving at line 99? Meaning I could declare a NoGroupByHaving<Optional<bla>> and the FieldType would take the type Optional<bla> ?

mailletf:
+1

plugins/sql_config_validator.h
guyd: Not sure if I fully understand the question.  The second template is a partial template specialization.  It is preferred whenever we instantiate the template NoGroupByHaving with an optional type.  This allows for testing if the optional is set and if so do the rest of the validation.

guyd: As mentioned, this is getting too complex for no reason.  I plan to revisit it once we I've some bandwith.

PR #213, opened 2016-02-25
MLDB-1369 MLDB-1419 routes

jeremybarnes:
Adds the ability for functions to have custom routes, and for plugins to host their own static content.

Concrete use-case is to allow the Tensorflow plugin to host a graph that can be used to explore a loaded model, as well as an eventual Tensorboard integration.


steadtler:
Code looks alright, but I have no idea of the impacts. +1, maybe Guy can look at it?

PR #214, opened 2016-02-25
Mldb 1417 reverted changes to readthroughbindingcontext

guyd:
You can ignore the first four commits.  They were from another branch I was about to merge in master.  

Basically, I've returned the BoundFunction signature to what they were a while back.  The changes are temporary and mainly to unblock Francois.  Therefore, much of the existing code was placed in #if #endif clauses.  This means that I had to disable temporal_earliest/min (sorry Mich).

Also, the rest of the temporal changes cannot be merged until we find a way to support all use cases.


jeremybarnes:
Would it be possible to rebase just the required changes?

+1 for the rest.  Clearly we will need some design discussion for how to re-enable the aggregators.

guyd:
@jeremybarnes I will rebase and merge into MLDB-1417-tensorflow-example.  Do you want me to merge also MLDB-1417-tensorflow-example into master or this is still work in progress?

jeremybarnes:
I think those changes are OK to merge as well.  If the tests pass, we're fine.  We should just note that the dynamic_cast calls may cause performance problems, and we should probably turn them back into static_casts when we're done debugging.

guyd:
The performance impact is tracked under MLDB-1421.

guyd:
First few commit were already merged in master.  The last commit was pushed directly to MLDB-1417-tensorflow-example.  The merge to master is tracked under https://github.com/mldbai/mldb/pull/215.

PR #215, opened 2016-02-25
Mldb 1417 tensorflow example

guyd:


guyd:
Got the green light from Jeremy.  Merging into master.

PR #217, opened 2016-02-27
MLDB-1423 thread pool improvements

jeremybarnes:
Fixes various issues in the new ThreadPool and adds the ability to specify the concurrency level.

Closes MLDB-1423, MLDB-1424, MLDB-1425

### DIFF
--- a/base/parallel.cc
+++ b/base/parallel.cc
@@ -1,0 +1,122 @@
+/** parallel.cc                                                    -*- C++ -*-
+    Jeremy Barnes, 5 February 2016
+    Copyright (c) 2016 Datacratic Inc.  All rights reserved.
+
+    This file is part of MLDB. Copyright 2015 Datacratic. All rights reserved.
+*/
+
+#include "parallel.h"
+#include "mldb/compiler/compiler.h"
+#include "mldb/base/exc_assert.h"
+#include "thread_pool.h"
+#include <atomic>
+#include <mutex>
+
+namespace Datacratic {
+
+void parallelMap(size_t first, size_t last,
+                 const std::function<void (size_t)> & doWork,
+                 int occupancyLimit)
+{
+    ExcAssertGreaterEqual(last, first);
+    ExcAssertLess((last - first), 1ULL << 31);
+
+    std::atomic<int> has_exc(0);
+    std::atomic<size_t> index(first);
+    std::mutex exc_mutex;
+    std::exception_ptr exc;
+
+    // This creates a thread pool that runs jobs on the default thread pool
+    ThreadPool tp;
+
+    if (occupancyLimit == -1)
+        occupancyLimit = numCpus();
+    if (occupancyLimit > (last - first))
+        occupancyLimit = (last - first);
+
+    auto worker = [&] ()
+        {
+            while (!has_exc.load(std::memory_order_relaxed)) {
+                size_t myindex = index.fetch_add(1);
+                if (myindex >= last)
+                    return;
+                try {
+                    doWork(myindex);
+                } JML_CATCH_ALL {
+                    if (has_exc.fetch_add(1) == 0) {
+                        ExcAssert(!exc);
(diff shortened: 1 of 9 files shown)


base/parallel.cc
```diff
@@ -0,0 +1,122 @@
...
+    std::atomic<size_t> index(first);
+    std::mutex exc_mutex;
+    std::exception_ptr exc;
+
+    // This creates a thread pool that runs jobs on the default thread pool
+    ThreadPool tp;
+
+    if (occupancyLimit == -1)
+        occupancyLimit = numCpus();
+    if (occupancyLimit > (last - first))
```
FinchPowers: Why not an else?

base/parallel.cc
```diff
@@ -0,0 +1,122 @@
...
+            while (!has_exc.load(std::memory_order_relaxed)) {
+                size_t myindex = index.fetch_add(1);
+                if (myindex >= last)
+                    return;
+                try {
+                    doWork(myindex);
+                } JML_CATCH_ALL {
+                    if (has_exc.fetch_add(1) == 0) {
+                        ExcAssert(!exc);
+                        exc = std::current_exception();
```
FinchPowers: What happens if there are multiple exceptions?

base/parallel.cc
```diff
@@ -0,0 +1,122 @@
...
+    std::atomic<size_t> index(first);
+    std::mutex exc_mutex;
+    std::exception_ptr exc;
+
+    // This creates a thread pool that runs jobs on the default thread pool
+    ThreadPool tp;
+    
+    if (occupancyLimit == -1)
+        occupancyLimit = numCpus();
+    if (occupancyLimit > (last - first + chunkSize - 1) / chunkSize)
```
FinchPowers: else?

base/thread_pool.cc
```diff
@@ -370,24 +374,37 @@ struct ThreadPool::Itl: public std::enable_shared_from_this<ThreadPool::Itl> {
...
-            int n = (entry.workerNum + i) % entry.queues->size();
+        auto stealFrom = [&] (int n)
+            {
+                const std::shared_ptr<ThreadQueue<ThreadJob> > & q
+                    = entry.queues->at(n);
+                
+                ThreadJob * job;
+                while ((job = q->steal())) {
+                    entry.lastFound = n;
+                    //++jobsStolen;
```
FinchPowers: Dead line?

base/parallel.cc
```diff
@@ -0,0 +1,122 @@
+/** parallel.cc                                                    -*- C++ -*-
+    Jeremy Barnes, 5 February 2016
+    Copyright (c) 2016 Datacratic Inc.  All rights reserved.
+
+    This file is part of MLDB. Copyright 2015 Datacratic. All rights reserved.
```
FinchPowers: Don't we want to keep a single Copyright line for new files?

FinchPowers:
Minor comments. +1

base/parallel.cc
jeremybarnes: Because if occupancyLimit is -1, numCpus() is 32 but (last - first) is 5, we still want only 5 jobs to run, not 32.

base/parallel.cc
jeremybarnes: The first one is captured, and the rest are caught and silently ignored.  I guess it would be possible to make it capture all exceptions, except that the number caught would vary because no new jobs are started once the first has been detected.

base/parallel.cc
jeremybarnes: Same as above... it's on purpose.

base/parallel.cc
jeremybarnes: It's OK.

base/thread_pool.cc
jeremybarnes: yes, true.  Nice catch; I will fix.

PR #219, opened 2016-03-01
Mldb 1420 filter in function

guyd:
Added a filter parameter to bound function structure to let function decide how to filter their values.  This approach is less encapsulated as it delegates to the callers (e.g FunctionCallWrapper) the work of filtering in the right-way the function arguments before passing them but it has the advantage of being a lot less intrusive (the signature of the BoundFunction exec is kept as-is).

### DIFF
--- a/testing/MLDB-1235-temporal-aggregators.py
+++ b/testing/MLDB-1235-temporal-aggregators.py
@@ -45,10 +45,10 @@ class TemporalTest(MldbUnitTest):
             ]
         )
 
-    def test_temporal_min_returns_first_event(self):
-        #temporal min on one column
+    def test_temporal_earliest_returns_first_event(self):
+        #temporal earliest on one column
         resp = mldb.get('/v1/query',
-                        q = 'select temporal_min(x) as t_min_x from dataset order by rowName()',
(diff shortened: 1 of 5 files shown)


testing/MLDB-1235-temporal-aggregators.py
```diff
@@ -45,10 +45,10 @@ def test_min_returns_last_event(self):
             ]
```
FinchPowers: Le test ici ne devrait-il pas retourner -2?

testing/MLDB-1235-temporal-aggregators.py
```diff
@@ -78,10 +78,10 @@ def test_temporal_min_returns_first_event(self):
...
-    def test_temporal_min_on_rows(self):
-        #temporal min on one column
+    def test_temporal_earliest_on_rows(self):
+        #temporal earliest on one column
         resp = mldb.get('/v1/query',
-                        q = 'select temporal_min({*}) as * from dataset order by rowName()',
+                        q = 'select temporal_earliest({*}) as * from dataset order by rowName()',
                         format = 'full').json()
 
         self.assertFullResultEquals(resp,
```
FinchPowers: Les dates dans le test ci-bas devraient utiliser la variable `TemporalTest.before`.

PR #220, opened 2016-03-01
Mldb 1428 text sparse output

steadtler:
Allow text.import procedure to import text with a select expression with an open schema.

### DIFF
--- a/plugins/importtext_procedure.cc
+++ b/plugins/importtext_procedure.cc
@@ -805,6 +805,8 @@ struct ImportTextProcedureWorkInstance
     //we might not know the number of output columns
 		//const size_t numberOutputColumns = columnNames.size();
 
+    const bool outputColumnNames = !areOutputColumnNamesKnown;
+
 		mutex lineMutex;
 
     PerThreadAccumulator< std::vector<std::pair<RowName, std::vector<std::tuple<ColumnName, CellValue, Date> > > > > accum;
@@ -822,13 +824,13 @@ struct ImportTextProcedureWorkInstance
           {
               if (names)
                 rowvalues.push_back( make_tuple(names[i], std::move(vals[i]), rowTs) );
-              else //identity select
-          	    rowvalues.push_back( make_tuple(inputColumnNames[i], std::move(vals[i]), rowTs) );
+              else
+                rowvalues.push_back( make_tuple(knownColumnNames[i], std::move(vals[i]), rowTs) );
           }
 
           rows.push_back( { rowName, std::move(rowvalues) } );
 
-          if (rows.size() == 1000)
+          if (rows.size() == ROWS_PER_CHUNK)
           {
           		{
           			std::unique_lock<std::mutex> guard(lineMutex);
@@ -840,7 +842,14 @@ struct ImportTextProcedureWorkInstance
           ++totalRows;
 	    };
 
-	    loadTextData(dataset, stream, config, scope, /*numberOutputColumns,*/ onLine);
+	    loadTextData(dataset, stream, config, scope, outputColumnNames, onLine);
+
+      for (int i = 0; i < accum.threads.size(); ++i)
+      {
+          auto & rows = *(accum.threads.at(i).get());
+          if (!rows.empty() > 0)
+            dataset->recordRows(rows);
+      };
 
 	    this->rowCount = totalRows;
 	}
@@ -852,8 +861,10 @@ struct ImportTextProcedureWorkInstance
 						 const ImportTextConfig& config,
 						 SqlCsvScope& scope)
 	{
-    if (areOutputColumnNamesKnown)
-		  dataset->initialize(knownColumnNames, columnIndex);
+      const bool outputColumnNames = !areOutputColumnNamesKnown;
+
+      if (areOutputColumnNamesKnown)
+		     dataset->initialize(knownColumnNames, columnIndex);
 
 	    auto createPayload = [=] ()
 	        {
@@ -944,9 +955,9 @@ struct ImportTextProcedureWorkInstance
 	            }
 	        };
 
-		loadTextData(dataset, stream, config, scope, /*numberOutputColumns,*/ onLine);
+		  loadTextData(dataset, stream, config, scope, outputColumnNames, onLine);
 
-		 // Accumulate the partial chunks, too, at the end
+		  // Accumulate the partial chunks, too, at the end
 	    std::mutex doneChunksLock;
 
 	    auto doLeftoverChunk = [&] (int threadNum)
@@ -986,6 +997,7 @@ struct ImportTextProcedureWorkInstance
 						 Datacratic::filter_istream& stream, 
 						 const ImportTextConfig& config,
 						 SqlCsvScope& scope,
+             bool outputColumnNames,
 						 const std::function<void (int, int64_t , RowName , Date , CellValue *, ColumnName * , int)> & processLine)
 	{	
 		  std::mutex lineMutex;
@@ -1081,6 +1093,7 @@ struct ImportTextProcedureWorkInstance
 	            
 	            //cerr << jsonEncodeStr(vector<CellValue>(values, values + numberOutputColumns)) << endl;
 	            
+              ExcAssert(!(isIdentitySelect && outputColumnNames));
 
 	            if (isIdentitySelect) {
 	                // If it's a select *, we don't really need to run the
@@ -1101,8 +1114,12 @@ struct ImportTextProcedureWorkInstance
                   // here.  Find another way to allocate it on the
                   // stack.
                   // CellValue valuesOut[numberOutputColumns];
+
                   vector<CellValue> valuesOut(selectRow.size());
-                  vector<ColumnName> namesOut(selectRow.size());
+                  vector<ColumnName> namesOut;
+
+                  if (outputColumnNames)
+                    namesOut.resize(selectRow.size());
 
 	                if (&selectOutput == &selectStorage) {
 	                    // We can destructively work with it
@@ -1114,7 +1131,8 @@ struct ImportTextProcedureWorkInstance
                           if (!std::get<1>(rItem).isAtom())
                               throw HttpReturnException(400, "select expression must return atomic values in text import procedure");
 	                        valuesOut[i] = std::move(std::get<1>(rItem).stealAtom());
-                          namesOut[i] =  std::move(std::get<0>(rItem));
+                          if (outputColumnNames)
+                            namesOut[i] =  std::move(std::get<0>(rItem));
 	                    }
 	                    
 	                }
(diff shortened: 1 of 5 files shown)


plugins/importtext_procedure.cc
```diff
@@ -96,7 +96,7 @@ ImportTextConfigDescription::ImportTextConfigDescription()
             }
             else {
                 context.exception("Unknown field '" + context.fieldName()
-                                  + " parsing CSV dataset configuration");
+                                  + " parsing import text configuration");
```
guyd: May be import.text since this is type of the procedure.

plugins/importtext_procedure.cc
```diff
@@ -697,11 +698,12 @@ struct ImportTextProcedureWorkInstance
 	        }             
 	    }
 	    
-	    // Early check for duplicate column names
+	    // Early check for duplicate column names in input
+      ML::Lightweight_Hash<ColumnHash, int> inputColumnIndex;
```
guyd: Indentation problem.

plugins/importtext_procedure.cc
```diff
@@ -594,6 +595,7 @@ struct ImportTextProcedureWorkInstance
     // happened.
     vector<ColumnName> inputColumnNames;
     bool isTextLine;
+    bool areOutputColumnNamesKnown;
```
guyd: Initialized to true I guess.

plugins/importtext_procedure.cc
```diff
@@ -730,29 +732,31 @@ struct ImportTextProcedureWorkInstance
...
 	                (400,
-	                 "CSV dataset select expression cannot have row-valued columns.",
+	                 "Import select expression cannot have row-valued columns.",
 	                 "select", config.select,
 	                 "selectOutputInfo", selectBound.info,
 	                 "columnName", col.columnName);
+
+          ColumnHash ch(col.columnName);
+          if (!columnIndex.insert(make_pair(ch, i)).second)
+                  throw HttpReturnException(400, "Duplicate column name in select expression",
```
guyd: Duplicate -> Duplicated

guyd: Adding the duplicated name would be nice for the user.

plugins/importtext_procedure.cc
guyd: There are two hash structures and one vector structure to keep the column names.  I don't have a complete view of the code but that seem too many.  Can we remove one or two?

steadtler: The first is for the input columns, and the second one for the output columns.

plugins/importtext_procedure.cc
```diff
@@ -797,39 +797,46 @@ struct ImportTextProcedureWorkInstance
 		          SqlCsvScope& scope)
 	{
 
-		const size_t numberOutputColumns = columnNames.size();
-
+    const bool outputColumnNames = !areOutputColumnNamesKnown;
```
guyd: There seems to also be an indentation problem here making the diff look bigger than it is actually.

plugins/importtext_procedure.cc
```diff
@@ -859,21 +865,60 @@ struct ImportTextProcedureWorkInstance
...
+                  if (!areOutputColumnNamesKnown) {
+
+                    //we need the first line to initialize
+                    ExcAssert(names != nullptr);
+                    for (unsigned i = 0;  i < numVals;  ++i) {
+                        const ColumnName & c = names[i];
+                        ColumnHash ch(c);
+                        knownColumnNames.push_back(c);
+                        if (!columnIndex.insert(make_pair(ch, i)).second)
+                            throw HttpReturnException(400, "Duplicate column name in CSV file",
```
guyd: Added back one instance of CSV file.  Was it the intent?

plugins/importtext_procedure.cc
```diff
@@ -859,21 +865,60 @@ struct ImportTextProcedureWorkInstance
...
+
+                    if (numVals != knownColumnNames.size())
+                      throw HttpReturnException(400, "Variable number of columns while importing text to tabular dataset");
+
+                    std::vector<CellValue> orderedValues(knownColumnNames.size());
+                    for (int i = 0; i < numVals; ++i) {
+
+                       auto iter = columnIndex.find(names[i]);
+                       if (iter == columnIndex.end())
+                          throw HttpReturnException(400, "Inconsistent column names while importing text to tabular dataset");
```
guyd: Here also it would be great if we could give a hint to the user.

guyd:
Can you update the PR with the indentation problems fixed?  It is difficult to review the core of the changes because of the noise.

plugins/importtext_procedure.cc
```diff
@@ -301,7 +300,7 @@ Encoding parseEncoding(const std::string & encodingStr)
     }
     else if (encodingStr == "latin1" || encodingStr == "iso8859-1")
         encoding = LATIN1;
-    else throw HttpReturnException(400, "Unknown encoding for CSV parser",
+    else throw HttpReturnException(400, "Unknown encoding for import text parser",
```
guyd: I guess we can change that one to import.text as well.

plugins/importtext_procedure.cc
```diff
@@ -797,39 +795,45 @@ struct ImportTextProcedureWorkInstance
...
-		const size_t numberOutputColumns = columnNames.size();
+            const bool outputColumnNames = !areOutputColumnNamesKnown;
+	    mutex lineMutex;
 
-		mutex lineMutex;
+            PerThreadAccumulator< std::vector<std::pair<RowName, std::vector<std::tuple<ColumnName, CellValue, Date> > > > > accum;
+            std::atomic<size_t> totalRows;
 
-        PerThreadAccumulator< std::vector<std::pair<RowName, std::vector<std::tuple<ColumnName, CellValue, Date> > > > > accum;
+	    auto onLine = [&] (int chunkNum, int64_t actualLineNum, RowName rowName, Date rowTs, CellValue * vals, ColumnName * names, int numberOutputColumns) {
```
guyd: Can we use a vector instead of a raw pointer?

plugins/importtext_procedure.cc
```diff
@@ -859,54 +862,93 @@ struct ImportTextProcedureWorkInstance
 
 	    mutex lineMutex;
 
-		auto onLine = [&] (int chunkNum, int64_t actualLineNum, RowName rowName, Date rowTs, CellValue * vals)
-	        {
-	        	TabularDatasetChunk & threadAccum = accum.get();
+	    auto onLine = [&] (int chunkNum, int64_t actualLineNum, RowName rowName, Date rowTs, CellValue * vals, ColumnName * names, int numVals) {
```
guyd: Can we also consider using a vector instead of a raw pointer here?

plugins/importtext_procedure.cc
```diff
@@ -859,54 +862,93 @@ struct ImportTextProcedureWorkInstance
...
+	            //size_t before JML_UNUSED = threadAccum.memusage();
+	            threadAccum.freeze();
+	            //size_t after JML_UNUSED = threadAccum.memusage();
+	            TabularDatasetChunk newChunk(numVals, ROWS_PER_CHUNK);
+	            std::unique_lock<std::mutex> guard(lineMutex);
+	            doneChunks.emplace_back(std::move(newChunk));
+	            doneChunks.back().swap(threadAccum);
+	            ExcAssertEqual(threadAccum.rowCount(), 0);
 
 	#if 0
```
guyd: While you are _intimate_ with the code, consider removing this commented code if it no longer apply.

plugins/importtext_procedure.cc
```diff
@@ -971,117 +1012,138 @@ struct ImportTextProcedureWorkInstance
...
-	                    // Need to copy things
-	                    const auto & selectRow = selectOutput.getRow();
-	                    ExcAssertEqual(selectRow.size(), numberOutputColumns);
-	                    for (unsigned i = 0;  i < selectRow.size();  ++i)
-	                        valuesOut[i] = std::get<1>(selectRow[i]).getAtom();
-	                }
-	                
-	                processLine(chunkNum, actualLineNum, std::move(rowName), rowTs, &valuesOut[0]);
+	                // Need to copy things
+	                //ExcAssertEqual(selectRow.size(), numberOutputColumns);
```
guyd: Why was this assert commented out?  If it no longer apply I would remove it.

plugins/importtext_procedure.cc
```diff
@@ -971,117 +1012,138 @@ struct ImportTextProcedureWorkInstance
...
+	                for (unsigned i = 0;  i < selectRow.size();  ++i) {
+                            auto& rItem = selectRow[i];
+                            if (!std::get<1>(rItem).isAtom())
+                                throw HttpReturnException(400, "select expression must return atomic values in text import procedure");
```
guyd: test import --> import.text

testing/MLDB-1428-text-sparse-output.py
```diff
@@ -0,0 +1,103 @@
...
+# MLDB-1428-text-sparse-output.py
+# Mathieu Bolduc, 2016-02-29
+# This file is part of MLDB. Copyright 2016 Datacratic. All rights reserved.
+#
+
+mldb = mldb_wrapper.wrap(mldb) # noqa
+
+
+import unittest
+class HavingTest(unittest.TestCase):
```
guyd: HavingTest --> InportTextTest

testing/MLDB-1428-text-sparse-output.py
```diff
@@ -0,0 +1,103 @@
...
+                "select": "*",
+                'outputDataset': {'id': 'iris', 'type': 'sparse.mutable'},
+                 'headers': ["a", "b", "c", "d", "label"],
+                'runOnCreation': True
+            } 
+        })
+
+        res = mldb.query('SELECT * FROM iris limit 1')
+
+        expected = [["_rowName",
```
guyd: Since new lines are not considered in comparison, I usually write it this way
```
["_rowName", a",  "b",  "c", "d", "label"],
 [ "97",  5.7,  2.9,  4.2, 1.3, "Iris-versicolor"]
```
which I find easier to read and reduce the length of the tests.

testing/MLDB-1428-text-sparse-output.py
```diff
@@ -0,0 +1,103 @@
...
+                    ]]
+
+        self.assertEqual(res, expected)
+
+    def test_sparse_unknown_cols(self):
+
+        res = mldb.put('/v1/procedures/import_reddit', {
+            "type": "import.text",
+            "params": {
+                "dataFileUrl": "http://files.figshare.com/1310438/reddit_user_posting_behavior.csv.gz",
```
guyd: Can we create a test that do not depend on an online dataset?  Some spurious failures have been due to datasets failing to download.

plugins/importtext_procedure.cc
steadtler: Its memory profiling code that will be useful.

plugins/importtext_procedure.cc
steadtler: I think Jeremy's intent was to skip 2 memory load per line.

testing/MLDB-1428-text-sparse-output.py
steadtler: But we dont want large datasets on the repository either. This one should be fairly reliable.

guyd: That was part of my suggestion.  Why do we need a large dataset here?  It seems that a 5-lines dataset would provide equally good test coverage.

guyd:
+1 - added few minor comments and suggestions but there is nothing blocking

testing/MLDB-1428-text-sparse-output.py
steadtler: The logic is dependant on blocks of thousands of lines.

PR #221, opened 2016-03-01
MLDBFB-318_misleading_error_msg.py

FinchPowers:


testing/MLDBFB-318_misleading_error_msg.py
```diff
@@ -0,0 +1,22 @@
...
+mldb = mldb_wrapper.wrap(mldb) # noqa
+
+
+class MisleadingErrorMessageTest(MldbUnitTest): # noqa
+
+    @unittest.expectedFailure
+    def test_misleading_error_message(self):
+        msg = "Function eat doesn't exist"
+        with self.assertMldbRaises(status_code=400, expected_regexp=msg):
+            mldb.get('/v1/query', q="SELECT eat('food')")
```
nicolaskruchten: this tests `eat(scalar)` when `eat` doesn't exist. Can we get cases for `eat()` and `eat(row)` when `eat` doesn't exist, as well as all three when `eat` does exist?

PR #225, opened 2016-03-01
HTTP_BASE_URL

jraby:
Lots of tweaks to be able to run the container at an alternative base url.


container_files/assets/www/doc/builtin/functions/Kmeans.md
```diff
@@ -49,7 +49,7 @@ Undecided | 0.5 | 0.5
 
 Which represents the following situation:
 
-![Sliced Dataset](/doc/builtin/img/KMeansFunction.svg)
+![Sliced Dataset](../img/KMeansFunction.svg)
```
nicolaskruchten: this is inconsistent with the change above, in terms of their relative positions. why `../img` in one case but `img/` in the other?

nicolaskruchten:
+1

PR #226, opened 2016-03-01
Mldb 1435 http base url doc prefix

FinchPowers:
@jeremybarnes Is it enough?
Is there a place where I can save the static variable rather than call secure_getenv over and over again?

### DIFF
--- a/server/mldb_doc_server.cc
+++ b/server/mldb_doc_server.cc
@@ -7,6 +7,11 @@
     Server for MLDB documentation.
 */
 
+#include <boost/program_options/cmdline.hpp>
+#include <boost/program_options/options_description.hpp>
+#include <boost/program_options/positional_options.hpp>
+#include <boost/program_options/parsers.hpp>
+#include <boost/program_options/variables_map.hpp>
 #include "mldb/server/mldb_server.h"
 #include "mldb/server/plugin_collection.h"
 #include "mldb/http/http_rest_proxy.h"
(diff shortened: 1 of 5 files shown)


server/mldb_server.cc
```diff
@@ -498,9 +498,11 @@ getPackageDocumentationPath(const Package & package) const
     // always be provided by the plugin "pro", but this is not
     // by any means guaranteed.
 
+    auto * rawBaseUrl = secure_getenv("HTTP_BASE_URL");
```
jeremybarnes: This should be passed as a command line option and put as a member variable of MldbServer, not magically brought in from the environment IMO.  As for the performance aspects, I don't think they are worrisome, especially if it's stored in a local variable.

FinchPowers:
@jeremybarnes @jraby Redone with the flag instead of the environment variable.

server/mldb_server.cc
```diff
@@ -499,8 +499,8 @@ getPackageDocumentationPath(const Package & package) const
     // by any means guaranteed.
 
     if (package.packageName() == "builtin")
-        return "/doc/builtin/";
-    return "/v1/plugins/" + package.packageName() + "/doc/";
+        return httpBaseUrl + "/doc/builtin/";
+    return httpBaseUrl + "/v1/plugins/" + package.packageName() + "/doc/";
```
guyd: Since httpBaseUrl is a user input I would use the safer boost::filesystem operator/.

nicolaskruchten: does this make sense in the context of a URL?

guyd: I think it does on linux since url and path are more or less the same thing.  If we don't use boost::filesystem then we can simply trim trailing slashes.  That's the case I was trying to prevent.

FinchPowers:
Inadequate.

server/mldb_server.cc
jeremybarnes: Personally, I don't think we should use the boost::filesystem for URLs, though I'd be curious to understand the advantages of b::f::operator/

server/mldb_doc_server.cc
```diff
@@ -18,9 +23,31 @@ using namespace Datacratic::MLDB;
...
+        cerr << all_opt << endl;
+        return 0;
+    }
+
+    notify(vm);
+
     vector<string> pluginDirectory = { "file://build/x86_64/mldb_plugins" };
 
     MldbServer server;
+    server.httpBaseUrl = httpBaseUrl;
```
jeremybarnes: Ideally we would make the init() take an httpBaseUrl rather than reaching inside the object.

PR #227, opened 2016-03-01
Fixed handling of PyTuple and PyList in converter

mailletf:
Lists and tuples were handled in the same if, which caused an exception in an actual prod plugin under certain scenarios. This fixes that by dealing with both cases seperately


plugins/lang/python/python_converters.cc
```diff
@@ -230,7 +230,15 @@ construct_recur(PyObject * pyObj)
     //else if(PyDateTime_Check(pyObj)) {
     //    throw ML::Exception("do datetime!!");
     //}
-    else if(PyTuple_Check(pyObj) || PyList_Check(pyObj)) {
+    else if(PyTuple_Check(pyObj)) {
```
guyd: The function could be turned into a function template to avoid the duplication.

guyd:
+1 once the tests pass

mailletf:
@guyd J'ai modifié le code pour suivre ta suggestion de template. Peux-tu rejeter un coup d’œil svp?

guyd:
+1  I will a look at what is happening with the build.

PR #228, opened 2016-03-02
MLDB-1429 allow record rows in tabular dataset

steadtler:


nicolaskruchten:
Out of curiosity, does this implementation allow recording/committing, and then recording/committing again? or is it 'sealed' once committed the first time?

steadtler:
Only 1 commit. I didnt think the Tabular dataset supported multiple commits?

plugins/experiment_procedure.cc
```diff
@@ -407,7 +407,7 @@ run(const ProcedureRunConfig & run,
         if(runProcConf.outputAccuracyDataset) {
             PolyConfigT<Dataset> outputPC;
             outputPC.id = ML::format("%s_results_%d", runProcConf.experimentName, (int)progress);
-            outputPC.type = "sparse.mutable";
```
guyd: DoD says that accuracy (classifier.test) should also have this default.

plugins/importtext_procedure.cc
```diff
@@ -697,9 +697,9 @@ struct ImportTextProcedureWorkInstance
 	        }             
 	    }
 	    
-	    // Early check for duplicate column names in input
-            ML::Lightweight_Hash<ColumnHash, int> inputColumnIndex;
-	    for (unsigned i = 0;  i < inputColumnNames.size();  ++i) {
+        // Early check for duplicate column names in input
```
guyd: Indentation issue?

plugins/tabular_dataset.cc
```diff
@@ -349,6 +353,57 @@ struct TabularDataset::TabularDataStore: public ColumnIndex, public MatrixView {
...
+            mutable_chunks.emplace_back(columnNames.size(), TABULAR_DATASET_DEFAULT_ROWS_PER_CHUNK);
+        }
+
+        std::vector<CellValue> orderedVals(columnNames.size());
+        Date ts;
+        for (unsigned i = 0;  i < vals.size();  ++i) {
+            const ColumnName & c = std::get<0>(vals[i]);
+            auto iter = columnIndex.find(c);
+            if (iter == columnIndex.end())
+                throw HttpReturnException(400, "New column name while recording tow in tabular dataset", "columnName", c.toString());
```
guyd: tow -> row

plugins/tabular_dataset.cc
```diff
@@ -349,6 +353,57 @@ struct TabularDataset::TabularDataStore: public ColumnIndex, public MatrixView {
...
+
+            orderedVals[iter->second] = std::get<1>(vals[i]);
+
+            ts = (i == 0) ? std::get<2>(vals[i]) : std::max(ts, std::get<2>(vals[i]));
```
guyd: We can probably initialize date to -INFINITY so that the line becomes
```
ts = std::max(ts, std::get<2>(vals[i]));
```

plugins/tabular_dataset.cc
```diff
@@ -349,6 +353,57 @@ struct TabularDataset::TabularDataStore: public ColumnIndex, public MatrixView {
...
+        }
+
+        mutable_chunks[0].add(rowName, ts, orderedVals.data());
```
guyd: Is it true that for tabular dataset each column has the same timestamp for a given row?  Isn't it this too limiting?

guyd:
Minor comments.  My only concern or misunderstanding is that it seems that tabular dataset force each columns to have the same timestamp for a given row.

plugins/tabular_dataset.cc
nicolaskruchten: Yes and yes. We can address that in a future iteration, as this is the way the old text.csv.tabular dataset used to work also.

guyd:
+1

plugins/tabular_dataset.cc
steadtler: @nicolaskruchten ?

plugins/experiment_procedure.cc
steadtler: But the output dataset in classifier is an optional, do it doesnt have a default... @nicolaskruchten ?

nicolaskruchten: it's possible to specify a default type while still keeping it optional. @guyd can you point to an example of this plz?

plugins/tabular_dataset.cc
nicolaskruchten: I'm comfortable with the old behaviour

plugins/experiment_procedure.cc
guyd: https://github.com/mldbai/mldb/blob/master/plugins/accuracy.cc#L51

nicolaskruchten:
I'm fine with tabular datasets being 'sealed' once committed, I just wanted to make sure I understood the limitations today. +1 from me if it's +1 from Guy.

PR #230, opened 2016-03-02
[MLDB-1357] turned to_timestamp and at function into operators

guyd:
This is the new syntax for at() and to_timestamp() functions.  The new syntax is in fact interpreted as a builtin function call.

Comments on the syntax are welcomed.   I will update the test to the new syntax once I get the green light on that.


steadtler:
ok +1

PR #234, opened 2016-03-03
MLDB-1435 http_base_url for doc

FinchPowers:
Pas mal plus vite quand on me dit où coder! :)


server/static_content_macro.cc
```diff
@@ -200,6 +200,16 @@ writeText(const Utf8String & text)
     hoedown_escape_html(output, (uint8_t *)text.rawData(), text.rawLength(), 0);
 }
 
+Utf8String&
+MacroContext::
+patchUrl(Utf8String & url) const
```
FinchPowers: @nicolaskruchten / naming dept. est-ce que ce nom de fonction est convenable?

nicolaskruchten: I would call it `prefixUrl` personally

server/static_content_macro.cc
```diff
@@ -216,7 +226,7 @@ writeInternalLink(Utf8String url,
     }
 
     writeHtml("<a href=\"");
-    writeText(url);
+    writeText(patchUrl(url));
```
nicolaskruchten: can you add this method to the `handleLink` function we were initially looking at as well plz?

nicolaskruchten: and the `nbLink` macro and the links generated by the `renderType` function?

guyd:
The DoD in MDLB-1435 talks also about resources created by MLDB.  This PR does not cover that.  Again, I might not understand completely the goal of that story.

server/static_content_macro.h
```diff
@@ -54,6 +54,7 @@ struct MacroContext: public MacroData {
     void writeInternalLink(Utf8String url,
                            const Utf8String & anchorText,
                            bool followInternalRedirect);
+    Utf8String& patchUrl(Utf8String & url) const;
```
guyd: I would hide this function in an anonymous namespace to reduce the API surface.

server/mldb_server.h
```diff
@@ -131,6 +131,7 @@ struct MldbServer: public ServicePeer, public EventRecorder {
     std::string getCacheDirectory() const;
 
     std::string httpBoundAddress;
+    std::string httpBaseUrl;
```
guyd: I would not keep this variable in the server unless we need it in the future.  I would pass it to the init function of the server and pass it to the static content handler.

nicolaskruchten: This will likely be useful later on in other places IMO

guyd:
+1

nicolaskruchten:
@guyd DoD clarified: https://datacratic.atlassian.net/browse/MLDB-1435

guyd:
Re-read the DoD.  I feel we should review the current PR with the goal of making documentation url relative whenever possible.  Also, this PR, as it is, does not cover most of the DoD.

nicolaskruchten:
@guyd for the documentation, it will be difficult to make URLs reliably relative without parsing the path, so inserting the base url is an attractive solution.

guyd:
@nicolaskruchten I don't know well that part of the code but I don't see why any parsing would be required since we are appending a prefix and not inserting in an URI.  Anyway, if it is simpler to have an absolute path let's do it.

nicolaskruchten:
We will also need this in the `getTypeName(...)` function in `static_content_macro.cc` and in the big template block in the `getStaticRouteHandler(...)` function in `static_content_handler.cc`

FinchPowers:
@nicolaskruchten Done.

server/mldb_server.h
jeremybarnes: Yeah, we need it to implement getPath() of the MldbServer.

server/mldb_server.cc
```diff
@@ -72,10 +72,11 @@ MldbServer::
 MldbServer(const std::string & serviceName,
            const std::string & etcdUri,
            const std::string & etcdPath,
-           bool enableAccessLog)
```
jeremybarnes: Consider moving enableAccessLog and httpBaseUrl into the init method. Otherwise I think we should remove the default values.

FinchPowers:
Verbal +1

server/mldb_server.cc
```diff
@@ -517,6 +519,31 @@ getCacheDirectory() const
     return cacheDirectory_;
 }
 
+Utf8String
+MldbServer::
+prefixUrl(Utf8String url) const
```
jeremybarnes: Should take a const Utf8String &, not make a copy

server/mldb_server.cc
```diff
@@ -517,6 +519,31 @@ getCacheDirectory() const
...
+{
+    if (url.startsWith("/")) {
+        return httpBaseUrl + url;
+    }
+    return url;
+}
+
+string
+MldbServer::
+prefixUrl(string url) const
```
jeremybarnes: We should not need this override.

server/mldb_server.cc
```diff
@@ -517,6 +519,31 @@ getCacheDirectory() const
...
+{
+    Utf8String str(url);
+    return prefixUrl(str).rawString();
+}
+
+string
+MldbServer::
+prefixUrl(const char* url) const
```
jeremybarnes: Or this one.

server/mldb_server.h
```diff
@@ -131,6 +132,11 @@ struct MldbServer: public ServicePeer, public EventRecorder {
...
+
+    Utf8String prefixUrl(Utf8String url) const;
```
jeremybarnes: These should all return a Utf8String.

server/mldb_server.h
jeremybarnes: httpBaseUrl should be a Utf8String.

server/static_content_macro.cc
```diff
@@ -200,6 +203,13 @@ writeText(const Utf8String & text)
     hoedown_escape_html(output, (uint8_t *)text.rawData(), text.rawLength(), 0);
 }
 
+Utf8String
+MacroContext::
+prefixUrl(Utf8String url) const
+{
+    return macroData->server->prefixUrl(url);
```
jeremybarnes: Are we sure we're initializing the server field before using it?

server/mldb_server.cc
FinchPowers: Removing this yields errors such as
```
./mldb/server/static_content_handler.cc:264:24: error: no match for ‘operator+=’ (operand types are ‘std::string {aka std::basic_string<char>}’ and ‘Datacratic::Utf8Strin
g’)
```

server/mldb_server.cc
FinchPowers: And in turn removing this yields errors such as
```
./mldb/server/static_content_handler.cc:264:105: error: call of overloaded ‘prefixUrl(const char [25])’ is ambiguous
                 result += "<link rel='stylesheet' href='" + server->prefixUrl("/resources/css/prism.css") + "'>\n";
```

server/static_content_macro.cc
FinchPowers: Yes. It is initialized in the ctor as "" by default.

FinchPowers: No. Clearly this is not the case since this is the cause of the segfault.

PR #235, opened 2016-03-04
move public html resources

jraby:
- mv container_files/assets/www container_files/public_html
  - version.json is now in public_html/resources
  - removed runtime copy of version.json
  - path/url fixups


nicolaskruchten:
I would do this slightly differently:

* container_files/assets/www/resources -> container_files/public_html (not container_files/public_html/resources)
* container_files/assets/www/doc -> container_files/doc (not container_files/public_html/doc)

jraby:
I did it like this because it required less changes to the makefiles and nginx config and because the layout would be the same in the container.

jraby:
verbal +1 from @nicolaskruchten

PR #237, opened 2016-03-07
Mldb 1342 added call to logger to ClassifierProcedure

guyd:
This is an early stage of the logger integration to see what you think:
- all entities will have a link to its class logger and logger output is specific for each class (here ClassifierProcedure) and use the type name for lookup
- cerr are changed to info level in most cases

The output looks like this:
```
Datacratic::MLDB::ClassifierProcedure [2016-03-07T13:16:19.901--5:00] info initialized feature space in elapsed: [0.00s cpu, 0.9277 mticks, 0.00s wall]
Datacratic::MLDB::ClassifierProcedure [2016-03-07T13:16:19.901--5:00] info extracted feature vectors in elapsed: [0.00s cpu, 1.3139 mticks, 0.00s wall]
Datacratic::MLDB::ClassifierProcedure [2016-03-07T13:16:19.902--5:00] info merged feature vectors in elapsed: [0.00s cpu, 0.3506 mticks, 0.00s wall]
Datacratic::MLDB::ClassifierProcedure [2016-03-07T13:16:19.902--5:00] info added feature vectors in elapsed: [0.00s cpu, 0.1196 mticks, 0.00s wall]
Datacratic::MLDB::ClassifierProcedure [2016-03-07T13:16:19.902--5:00] info indexed training data in elapsed: [0.00s cpu, 0.0110 mticks, 0.00s wall]
Datacratic::MLDB::ClassifierProcedure [2016-03-07T13:16:19.902--5:00] info Training with 6 features
Datacratic::MLDB::ClassifierProcedure [2016-03-07T13:16:19.902--5:00] info factorTrue = 0.141421
Datacratic::MLDB::ClassifierProcedure [2016-03-07T13:16:19.902--5:00] info factorFalse = 0.1
```

The configuration (not implemented yet) would be part of a general MLDB config.  The log section would look like this
```
[logging]
# specifies the default logging handler
handler=Console
# specifies the default logging level
level=WARNING
# specifies a logging level for class ClassifierProcedure
Datacratic::MLDB::ClassifierProcedure.level=ALL
```

### DIFF
--- a/core/function.h
+++ b/core/function.h
@@ -563,7 +563,9 @@ registerFunctionType(const Package & package,
                                     PolyConfig config,
                                     const std::function<bool (const Json::Value)> & onProgress)
                                 {
-                                    return new FunctionT(FunctionT::getOwner(server), config, onProgress);
+                                    auto function = new FunctionT(FunctionT::getOwner(server), config, onProgress);
+                                    function->logger = MLDB::getMldbLog<FunctionT>();
+                                    return function;
                                 },
                                 makeInternalDocRedirect(package, docRoute),
                                 customRoute,
--- a/core/mldb_entity.h
+++ b/core/mldb_entity.h
@@ -10,6 +10,7 @@
 #pragma once
 
 #include "mldb/rest/poly_entity.h"
+#include "mldb/utils/log_fwd.h"
 
 namespace Datacratic {
 
@@ -64,6 +65,7 @@ struct MldbEntity: public PolyEntity {
     }
 
     static constexpr const char * INTERNAL_ENTITY  = "INTERNAL_ENTITY";
+    std::shared_ptr<spdlog::logger> logger;
 };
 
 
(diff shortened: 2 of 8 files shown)


nicolaskruchten:
At first glance, it seems like the `Datacratic::MLDB::` leader is a bit redundant... Could we keep just the classname without namespaces?

guyd:
We can output the class name only.  Alternatively, we could output a shorten version like Java loggers do
```
D::M::ClassifierProcedure
```
to limit the chances of confusion when logging from classes with the same name but in different namespace.

nicolaskruchten:
I would favour class names only until we notice that name clashes are getting confusing. Also, there appears to be an extra `-` between the time and the timezone offset.

guyd:
@nicolaskruchten The extra ```-``` was abug in spdlog that I reported earlier and that they have fixed.  It is on my todo list to integrate with the latest spdlog later this week.

PR #238, opened 2016-03-07
[MLDB-1437] handling of all redirect codes in file downloader

guyd:


guyd:
@jeremybarnes You are the one who touched that file the most in the last couple of month.  Can you take a quick look?

mailletf:
+1

PR #239, opened 2016-03-08
Mldb 1241 enron demo

simlmx:
First version of the Enron Spam Filtering demo. Only a simple bag of word which leads to awesome AUC. We then say "yes but" and propose to use a generalization of the F-score as a more interesting score for our spam filter. The results then don't appear so good but... to be continued!


simlmx:
Closed because outdated. Will make two new PR in time, one with the v1 of the demo, and one with v2 (I might just reopen this one at that point).

PR #245, opened 2016-03-11
Mldb 1443 cls test coverage

mailletf:
This fixes a couple of things. Apologies for mixing up things. A commit-by-commit review is probably easier.

- Changed Classifier_Impl inferface so that the label passed in explain function is now a float so we can deal with regression tasks.
- Added check for a null feature-set when we try to use the classifier function with no columns matching (ie doing a select on columns where none exist in the dataset)
- Binary stats track weighted counts as well as unweighted counts to fix the accuracy plugin when doing a weighted training. the code incremented counters assuming it got actual example counts, but when doing a weighted training, examples were represented by floats possibly smaller than 1 so counters would not move in the expected manner

### DIFF
--- a/container_files/public_html/doc/builtin/functions/ClassifierApply.md
+++ b/container_files/public_html/doc/builtin/functions/ClassifierApply.md
@@ -31,8 +31,11 @@ These functions output a single value where the name depends on the classifier m
 
 ## Status
 
-The status of a Classifier Apply function will return a JSON representation of the
-model parameters of the trained classifier, to allow introspection.
+To allow introspection into a trained model, the following routes of a Classifier function will 
+return status information:
+
+* `/status`: overview of the trained model
+* `/details`: parameters of the trained model, such as weights
 
 ## Examples
 
--- a/ml/jml/boosted_stumps.cc
+++ b/ml/jml/boosted_stumps.cc
@@ -773,10 +773,13 @@ struct Results_Explain {
 Explanation
 Boosted_Stumps::
 explain(const Feature_Set & feature_set,
-        int label,
+        float label,
         double weight,
         PredictionContext * context) const
 {
+    if(label >= bias.size())
+        throw ML::Exception("Boosted Stumps do not support explain for regression tasks");
+
     Explanation result(feature_space(), weight); 
 
     // Get the bias for the label
--- a/ml/jml/boosted_stumps.cc
+++ b/ml/jml/boosted_stumps.cc
@@ -777,9 +777,6 @@ explain(const Feature_Set & feature_set,
         double weight,
         PredictionContext * context) const
 {
-    if(label >= bias.size())
-        throw ML::Exception("Boosted Stumps do not support explain for regression tasks");
-
     Explanation result(feature_space(), weight); 
 
     // Get the bias for the label
--- a/ml/jml/decision_tree.cc
+++ b/ml/jml/decision_tree.cc
@@ -421,7 +421,7 @@ print_recursive(int level, const Tree::Ptr & ptr,
 Explanation
 Decision_Tree::
 explain(const Feature_Set & feature_set,
-        int label,
+        float label,
         double weight,
         PredictionContext * context) const
 {
@@ -436,7 +436,7 @@ void
 Decision_Tree::
 explain_recursive(Explanation & explanation,
                   const Feature_Set & feature_set,
-                  int label,
+                  float label,
                   double weight,
                   const Tree::Ptr & ptr,
                   const Tree::Node * parent) const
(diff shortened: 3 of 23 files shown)


steadtler:
Looks fine to me +1

ml/jml/decision_tree.cc
```diff
@@ -444,6 +444,10 @@ explain_recursive(Explanation & explanation,
     StandardGetFeatures get_features(feature_set);
     int nl = label_count();
 
+    if(nl < 2)
+        throw ML::Exception("Decision_Tree does not support explain "
```
guyd: If this is outputted to users, I would not refer to the class name.

ml/jml/feature_set.h
```diff
@@ -365,6 +365,9 @@ class Mutable_Feature_Set : public Feature_Set {
     virtual std::tuple<const Feature *, const float *, int, int, size_t>
     get_data(bool need_sorted = false) const
     {
+        if(features.size() == 0)
+            throw ML::Exception("features_type is empty!");
```
guyd: Same here.

ml/jml/glz_classifier.cc
```diff
@@ -363,12 +363,25 @@ explain(const Feature_Set & feature_set,
...
+        label_idx = label;
+    }
+    if(label_idx >= weights.size()) {
+        throw ML::Exception("label bigger than weight vector");
+    }
+
     for (unsigned j = 0;  j < features.size();  ++j) {
         result.feature_weights[features[j].feature]
-            += weight * weights[label][j] * decoded[j];
+            += weight * weights[label_idx][j] * decoded[j];
```
guyd: What's the effect of operator[] on 0.5f?

plugins/classifier.cc
```diff
@@ -878,6 +878,11 @@ apply(const FunctionApplier & applier_,
         }
     }
     else {
+        if(!fset) {
+            throw ML::Exception("Feature_Set is null! Are you giving only null features to "
```
guyd: I would not refer to an internal structure in a user error.

testing/MLDB-256_accuracy_accepts_all_cls_modes.py
```diff
@@ -69,6 +69,124 @@ def setUpClass(self):
...
+                    },
+                    "kfold": 2,
+                    "mode": "boolean",
+                    "outputAccuracyDataset": True,
+                    "runOnCreation": True
+                }
+            })
+    
+        mldb.log(str(re.exception))
+        self.assertTrue("Feature_Set is null! Are you giving only null features to  the classifier function?" in str(re.exception))
```
guyd: This check can be done directly with assertRaisesRegex.

container_files/public_html/doc/builtin/functions/ClassifierApply.md
```diff
@@ -31,8 +31,11 @@ These functions output a single value where the name depends on the classifier m
 
 ## Status
 
-The status of a Classifier Apply function will return a JSON representation of the
-model parameters of the trained classifier, to allow introspection.
+To allow introspection into a trained model, the following routes of a Classifier function will 
+return status information:
+
+* `/status`: overview of the trained model
+* `/details`: parameters of the trained model, such as weights
```
nicolaskruchten: are there examples or tests for these routes?

ml/jml/boosted_stumps.cc
```diff
@@ -773,7 +773,7 @@ struct Results_Explain {
 Explanation
 Boosted_Stumps::
 explain(const Feature_Set & feature_set,
-        int label,
+        float label,
```
jeremybarnes: There is a class Label in ml/jml/label.h that is designed for this usage.

ml/jml/feature_set.h
jeremybarnes: I'm not sure that this should be an error.  In a sparse classifier, it's certainly possible that rows would have no data.

ml/jml/glz_classifier.cc
jeremybarnes: Use the Label class; it should be clearer that it's a float for regression and an int for classification.

container_files/public_html/doc/builtin/functions/ClassifierApply.md
mailletf: Not that I know of

ml/jml/feature_set.h
mailletf: I put that in because 2 lines lower, it tries to access the first element of features and so we're vulnerable to a segfault without that check

guyd:
I am not to clear about the usage of label but this seems clearer.  +1

PR #247, opened 2016-03-14
Don't depend on phony target for plugin tests (MLDB-1456)

jeremybarnes:
Stops the tests running over and over when they depend on a plugin.

Verified manually that a change to the plugin still causes the tests to be rerun.


nicolaskruchten:
+1

PR #248, opened 2016-03-14
MLDB-1447: set _mldb's shell to /bin/bash

jraby:


nicolaskruchten:
I tried exactly that and it didn't work for me :(

It works for you through Jupyter??

nicolaskruchten:
+1

PR #249, opened 2016-03-14
[MLDB-1342] set unnecessary boot message to debug and removed other m…

guyd:
…isleading messages


rest/http_rest_service.cc
```diff
@@ -89,8 +89,7 @@ HttpRestConnection::
 sendErrorResponse(int responseCode, const Json::Value & error)
 {
     using namespace std;
-    cerr << "sent error response " << responseCode << " " << error
-         << endl;
+    cerr << "sent error response " << responseCode << " " << error << endl;
```
nicolaskruchten: this is still cerr?

guyd: I targeted messages that we shown at boot time.  This one can be done next time I touch that file!  :-)

nicolaskruchten:
Looking good! Can we get similar treatment for the credentials daemon and the bit in the mldb_server that connects to it?

guyd:
Yes.  I am on it.

guyd:
looking into failed test

guyd:
I've discussed, added an option to mute the output in credential daemon exec.

nicolaskruchten:
+1

PR #251, opened 2016-03-15
MLDB-1454 remove table name in WHERE for function

steadtler:
The WHERE optimizations are done by the dataset but the dataset doesn't know its alias which is context-dependant so we had to pass the alias to the dataset.

### DIFF
--- a/core/dataset.cc
+++ b/core/dataset.cc
@@ -1228,7 +1228,7 @@ queryBasic(const SqlBindingScope & scope,
            bool allowParallel) const
 {
     // 1.  Get the rows that match the where clause
-    auto rowGenerator = generateRowsWhere(scope, "", where, 0 /* offset */, -1 /* limit */);
+    auto rowGenerator = generateRowsWhere(scope, "" /*alias*/ , where, 0 /* offset */, -1 /* limit */);
 
     // 2.  Find all the variables needed by the orderBy
     // Remove any constants from the order by clauses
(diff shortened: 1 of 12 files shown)


core/dataset.cc
```diff
@@ -1224,7 +1228,7 @@ queryBasic(const SqlBindingScope & scope,
            bool allowParallel) const
 {
     // 1.  Get the rows that match the where clause
-    auto rowGenerator = generateRowsWhere(scope, where, 0 /* offset */, -1 /* limit */);
+    auto rowGenerator = generateRowsWhere(scope, "", where, 0 /* offset */, -1 /* limit */);
```
guyd: Consider adding /* alias */ so that it is easier to know what this empty string is referring to.

testing/MLDB-1305_rowNames_join.py
```diff
@@ -26,6 +26,83 @@
...
+
+expected = [["_rowName","col1"],["row1","a"]]
+
+res = mldb.perform("GET", "/v1/query", [["q", """
+    SELECT *
+    FROM dataset1
+    WHERE dataset1.col1 = 'a'
+    """],['format', 'table']])
+
+assert res["statusCode"] != 404
```
guyd: Why don't you use python unittest?  I thought we would be adopting it for new tests.

guyd:
+1 with minor comments

PR #253, opened 2016-03-15
MLDB-1464 tensorflow python

jeremybarnes:
This builds and exposes Tensorflow's Python wrappers, allowing for Tensorflow graphs to be trained and manipulated and for Tensorflow examples to be run.


jeremybarnes:
Test failure is due to lack of swig on the Jenkins machine.  There are two possibilities:
1.  Check in the artifacts that swig generates
2.  Install swig as a build dependency

@jraby @nicolaskruchten thoughts?

tensorflow/doc/Tensorflow.md
```diff
@@ -1 +1,30 @@
...
+MLDB incorporates the
+[Tensorflow deep learning framework](https://www.tensorflow.org/).
+This allows MLDB to apply and train deep learning models.
+
+The following functionality is currently exposed in MLDB:
+
+* Application of serialized Tensorflow graphs as a function
+  (see the [Tensorflow Graph Function](TensorflowGraph.md)
+* Access to the full set of Python wrappers for tensorflow.  This
+  allows the examples in Tensorflow to be applied unmodified from within
```
nicolaskruchten: from within a notebook?

jeremybarnes: Hmmm, you're right.  We would need to set the PYTHONPATH for ipython to make this accessible from a notebook.

PR #255, opened 2016-03-15
Mldbfb 411 clearer expression val error msg

mailletf:


sql/expression_value.h
```diff
@@ -807,6 +807,31 @@ struct ExpressionValue {
 
     Type type_;
 
+    static inline std::string print(Type t) {
+        switch (t)  {
+        case NONE:      return "empty";
+        case ATOM:      return "atomic value";
```
guyd: Although _atom_ is still present in the documentation, I thought we were using _scalar_ now.  @nicolaskruchten ?

sql/expression_value.h
```diff
@@ -807,6 +807,31 @@ struct ExpressionValue {
...
+        case ROW:       return "row";
+        case STRUCT:    return "struct";
+        case EMBEDDING: return "embedding";
+        default:
+            throw ML::Exception("Unknown ExpressionValue type: " + t);
```
guyd: If this is targeted to user, I would not mention the name of the class here.

sql/expression_value.h
```diff
@@ -807,6 +807,31 @@ struct ExpressionValue {
...
+        }
+    }
+
+    void inline assertType(Type requested, const std::string & details="") const {
+        if(requested != type_) {
+            std::string msg = "Cannot convert ExpressionValue of type "
```
guyd: and here

guyd:
+1 with minor comments

sql/expression_value.h
nicolaskruchten: atomic value is ok for now, we still haven't fully standardized :)

PR #257, opened 2016-03-16
[MLDB-1477] fixed temporal aggregators problem with expression of dif…

guyd:
…ferent type

The logic to detect the expression type was incorrect, leading to errors when dealing with dataset containing a column.


FinchPowers:
+1

PR #258, opened 2016-03-16
MLDB-1452 introducing the LIKE operator

steadtler:


sql/sql_utils.cc
```diff
@@ -38,6 +43,64 @@ std::string escapeSql(const char * str)
...
+            case ('.'): {
+                regExFilter += "[.]";
+                break;
+            }
+            case ('|'): {
+                regExFilter += "[|]";
+                break;
+            }
+            case ('('): {
+                regExFilter += "[()]";
```
nicolaskruchten: i don't know what this case statement does but shouldn't this be `[(]` instead of the current `[()]` ?

sql/sql_expression.cc
```diff
@@ -1076,14 +1076,11 @@ const SqlExpression::Operator operators[] = {
     { "!<", false,       SqlExpression::compar, 4, "Not less than" },
     { ">", false,        SqlExpression::compar, 4, "Greater than" },
     { "<", false,        SqlExpression::compar, 4, "Less than" },
-    { "NOT", true,       SqlExpression::booln,  5, "Boolean not" },
+    { "NOT", true,       SqlExpression::booln,  5, "Unary not" },
     { "AND", false,      SqlExpression::booln,  6, "Boolean and" },
     { "OR", false,       SqlExpression::booln,  7, "Boolean or" },
     { "ALL", true,       SqlExpression::unimp,  7, "All true" },
     { "ANY", true,       SqlExpression::unimp,  7, "Any true" },
-    { "BETWEEN", false,  SqlExpression::unimp,  7, "Between operator" },
```
nicolaskruchten: did the BETWEEN and IN line go somewhere else or... ?

steadtler: They are actually implemented now, though not as explicit binary operators, so I took them out of the unimplemented list :)

sql/sql_utils.cc
steadtler: Well spotted! It passed the test because of how [] works in reggex :) Fixed.

sql/sql_expression.cc
nicolaskruchten: Gotcha, I didn't catch the `unimp` part ;)

nicolaskruchten:
+1, nice tests!

sql/sql_expression_operations.cc
```diff
@@ -2851,6 +2851,105 @@ getChildren() const
...
+                     const VariableFilter & filter) -> const ExpressionValue &
+        {
+            ExpressionValue vstorage, fstorage;
+
+            const ExpressionValue & value = boundLeft(rowScope, vstorage, filter);
+
+            if (!value.isString())
+                throw HttpReturnException(400, "LIKE expression expected its left hand value to be a string, got " + value.getTypeAsUtf8String());
+
+            const ExpressionValue & filterEV = boundRight(rowScope, fstorage, filter);
```
guyd: The filter is constant expression and do not depend on the row.  All the validation and the regex compilation should ideally be done at binding not at execution.

sql/sql_utils.cc
```diff
@@ -38,6 +43,65 @@ std::string escapeSql(const char * str)
...
+                break;
+            }  
+            default: {
+                regExFilter += filterChar;
+            }
+        }
+    }
+
+    //std::cerr << "sql filter regex filter: " << regExFilter << std::endl;
+    boost::u32regex regex = boost::make_u32regex(regExFilter.rawData());
```
guyd: This part, the regex compilation, could be expensive.  I would do it only once at binding time.

sql/sql_utils.cc
```diff
@@ -38,6 +43,65 @@ std::string escapeSql(const char * str)
...
 }
 
+bool matchSqlFilter(const Utf8String& valueString, const Utf8String& filterString)
+{
+    if (filterString.empty() || valueString.empty())
+        return false;
+
+    Utf8String regExFilter;
+
+    for (const auto& filterChar : filterString) {
```
guyd: If I understand correctly, the purpose here is to transform a like expression into a regex.  I believe that this does not handle the escape mechanism in like expression.  If I read the doc correctly, ```[]``` are used as escape.  Here we transform them into character to match.  Also, there are several special character in regex (e.g. $, ^) that are not special in like expression that would need to be escaped as well.

testing/MLDB-1452-like-operator.py
```diff
@@ -0,0 +1,160 @@
...
+        res = mldb.query('''
+            select x
+            from sample3
+            where x LIKE '%(__)%'
+        ''')
+
+        expected = [["_rowName","x"],["e","eg(ro)te"]]
+        self.assertEqual(res, expected)
+
+    def test_like_number(self):
```
guyd: I think we need to add test case for like expression with escape
```
%[%]%
```
which match any string with a ```%``` character.

guyd:
I might have misunderstood the code but I feel there are a few cases missing.

testing/MLDB-1452-like-operator.py
steadtler: square brackets arent part of LIKE... this filter means any string with an open bracket character before a close bracket character.

sql/sql_utils.cc
steadtler: See below, [] are not part of like. I'll check for the $ and ^, not sure which standard boost is running.

sql/sql_utils.cc
steadtler: The filter is not necessarily be constant, and regardless, ASAIK the regex object is not multithread safe :(

sql/sql_expression_operations.cc
steadtler: The filter is not necessarily constant. You could build it or take in in a dataset.

testing/MLDB-1452-like-operator.py
guyd: Then how do we say any string with ```%```?

nicolaskruchten: We can't at the moment.

sql/sql_utils.cc
guyd: I thought it was and this seems to be confirmed here:
http://www.boost.org/doc/libs/1_34_1/libs/regex/doc/thread_safety.html.

In any case, this could be part of an optimization round and does not need to be done now I feel.

testing/MLDB-1452-like-operator.py
nicolaskruchten: (or rather, we use `regex_match()` :) )

guyd: I see.

sql/sql_expression_operations.cc
guyd: I would add a test case for a non-constant filter then.

guyd:
+1 once it is confirmed that reserved regex characters are handled correctly.

PR #259, opened 2016-03-16
Quick fix for Accuracy procedure thread-safety issue

steadtler:
recordRows is NOT guaranteed to be multithread-safe and so calls to it must be properly protected.


guyd:
+1

PR #263, opened 2016-03-17
Mldb 1415 neighbors func

mailletf:
- Adds nearest.neighbour function type
- Removes /nearestNeighbour and /neighbour endpoints to embedding dataset

WARNING: There is currently a problem in the way we deal with column ordering of rows vs embeddings. I've created the issue MLDB-1486 to address this. There are a few places in the following commits where I've had to disable sorting of columns, which ave been marked with the above ticket number, to have things work. One test is marked as failing and should be fixed. We may of may not want to solve the underlying issue before merging this in.


plugins/embedding.cc
```diff
@@ -897,6 +879,34 @@ struct EmbeddingDataset::Itl
...
+        auto dist = [&] (int item) -> float
+        {
+            float result = repr->dist(item, coord);
+            ExcAssert(isfinite(result));
+            return result;
+        };
+
+        auto neighbours = repr->vpTree->search(dist, numNeighbours, maxDistance);
+
+        cerr << "neighbours = " << jsonEncode(neighbours) << endl;
```
steadtler: Comment out?

plugins/embedding.cc
```diff
@@ -1213,20 +1165,169 @@ getKnownColumnInfos(const std::vector<ColumnName> & columnNames) const
...
 }
 
-RestRequestMatchResult
-EmbeddingDataset::
-handleRequest(RestConnection & connection,
-              const RestRequest & request,
-              RestRequestParsingContext & context) const
+
+/*****************************************************************************/
+/* NEAREST NEIGHBOUR FUNCTION                                                */
```
steadtler: We should move it to its own file.

plugins/embedding.cc
```diff
@@ -1213,20 +1165,169 @@ getKnownColumnInfos(const std::vector<ColumnName> & columnNames) const
...
+    }
+
+    RowValue rtnRow;
+    for(auto & neighbor : neighbors) {
+        // std::tuple<Coord, CellValue, Date>
+        rtnRow.push_back(make_tuple(get<0>(neighbor), CellValue(get<2>(neighbor)), d));
+    }
+
+    output.set("neighbors", rtnRow);
+    return output;
```
steadtler: return std::move(output) ?

jeremybarnes:
It probably makes sense to merge this now, even though it could be better tested.  It will have a pretty major effect on performance, but guarantees the correct results in nearly all circumstances.  Things that are still not done:

1.  normalize() function should fall back on more generic functionality not re-implement everything
2.  specialization of the ExpressionValueInfo functions for when we have embeddings, etc as currently they will be very slow
3.  more test cases in expression_value_test of different paths through the code
4.  optimization of the binary operators between embeddings and scalars or embeddings to use the existing code
5.  porting of the code in pro
6.  {0: 1, 1: 2} + [1, 2] should work and return {0: 2, 1: 4} from duality of row/embedding representation
7.  [1, 2] + { 0: 1, 1: 2} should work and return [ 2, 4] from duality of row/embedding representation
8.  one nearest neighbours test case doesn't work
9.  the embedding dataset should return indexes for its columns, since they are always in the same place
10. same for the tabular dataset, once it's merged
11.  same for variable and wildcard lookups in datasets
12.  tests of the info returned from getCompatibleDoubleEmbedding
13.  test of vector_sum({0: 1, 1: 2}, [1, 2]) which should work as per above

Note that all of these issues existed previously.

nicolaskruchten:
What should I do with the laundry-list of "still not done" here...? Are these performance optimizations or breakages?

jeremybarnes:
- 5 and 8 need to be addressed before merging
- The work for MLDB-1486 should address 6, 7, 13
- 2, 4, 9, 10 and 11 are optimizations which should be handled in a follow-up once MLDB-1486 is done
- 3, 12 are testcases which are missing
- 1 is technical debt reduction.

We should also make the embedding code in analytics.cc fall back on standard functionality.

nicolaskruchten:
OK. Can we independently merge the branch from MLDB-1486 ? Is it far from being ready?

jeremybarnes:
The tip of the MLDB-1486 branch is this one, which has all of the low-risk part and passes tests apart from the one indicated.  It's still a fair way from being ready.  I would de-risk by doing the release in two parts: this one first, and the deeper MLDB-1486 changes once they are ready.

mailletf:
@jeremybarnes I've tried to fix the failing test in MLDB-665_nearest_neighbours.py but it feels like there is still a problem. I've made a simpler failing one in the same file called "test_num_neighbours_parameter" that maybe would help

jeremybarnes:
This branch is subsumed by the MLDB-1486 branch; I'm closing it.  The TODOs have been collected and added to JIRA.

PR #264, opened 2016-03-17
Fixed git import issues (MLDB-1487)

jeremybarnes:
- Commit is properly called, to avoid empty dataset issues
- We accept messages with ISO-8859-1 encodings, and other encodings that are unknown simply cause the commit message to be skipped.
- Note that the test is not enabled, since there is a separate error caused by problems with expression scope nesting.


nicolaskruchten:
+1

PR #265, opened 2016-03-18
[MLDB-1485] forced the shell variables on the mldb_runner exec

guyd:


nicolaskruchten:
as per @jraby , let's switch to using export instead of the backslashes.

PR #266, opened 2016-03-18
[MLDB-1232] tfidf output rows not columns

guyd:
The tfidf output has changed format.  It is now contain one row per term with the document count in the _count_ column.


plugins/tfidf.cc
```diff
@@ -238,15 +241,39 @@ run(const ProcedureRunConfig & run,
...
 
         Date applyDate = Date::now();
+        ColumnName columnName("count");
 
-        std::vector<std::tuple<ColumnName, CellValue, Date> > row;
-        row.reserve(dfs.size());
-
-        for (auto& df : dfs) {
-            row.emplace_back(ColumnName(df.first), df.second/*(float)count*/, applyDate);
+#if 0
```
steadtler: Should be cleaned before merging, you'll have it saved in github regardless if you want to use it later.

plugins/tfidf.cc
```diff
@@ -25,6 +25,7 @@
 #include "mldb/vfs/filter_streams.h"
 #include "mldb/vfs/fs_utils.h"
 #include "mldb/plugins/sql_config_validator.h"
+#include <cmath>
```
steadtler: Only necessary to the commented out code?

steadtler:
Simple enough, but we should avoid dead code as much as possible. +1 after cleaning.

plugins/tfidf.cc
guyd: Indeed

plugins/tfidf.cc
guyd: I will.  It may be put back as early as this week but, as pointed out, this is already in github!

plugins/tfidf.cc
```diff
@@ -37,7 +37,8 @@ serialize(ML::DB::Store_Writer & store,
     std::string name = "tfidf";
     int version = 0;
     store << name << version;
-    store << corpusSize;
+    store << corpusSize; // number of documents in corpus
+    store << dfs.size(); // number of terms in corpus
```
guyd: Here I change what we store in the model and *should* ideally change the version.  However, since the tf-idf has not been used yet in prod, I did not do so.

PR #268, opened 2016-03-21
Mldb 1342 log config

guyd:
- Updated spdlog to fix the double "-" sign for timezone
- Modified the config interface so that lookup function accept a default value.  This makes the calling code simpler.  Updated the tests accordingly.
- The logger now read the config to determine which log level to use


steadtler:
+1

PR #269, opened 2016-03-21
MLDB-1360 sparse mutable multithreaded insert

jeremybarnes:
This adds a test-case and fixes the multithreading issues for the sparse mutable dataset.

It also removes mutexes in the import procedures, and fixes a few other small-ish bottlenecks.  As part of doing that, the indentation of the import text producedure was fixed.

Effect is that bid request importation goes from 2,000 bid requests per second to 20,000.


steadtler:
+1

PR #272, opened 2016-03-21
Mldb 1488 explain regression

mailletf:


ml/jml/decision_tree.h
```diff
@@ -120,7 +120,8 @@ class Decision_Tree : public Classifier_Impl {
                            const ML::Label & label,
                            double weight,
                            const Tree::Ptr & ptr,
-                           const Tree::Node * parent) const;
+                           const Tree::Node * parent,
+                           std::function<float (const Tree::Base&)>& get_pred) const;
```
jeremybarnes: should be a const reference to the function, or ideally a simple function pointer

jeremybarnes:
I'm not sure that this is how we want to do it.

For binary classification, we are trying to push it towards either positive or negative infinity, and so any contribution can be accumulated, with only the sign affecting the end result.

For regression, we want to hit a given number right on.  So the label is important.

For example, let's say that the label is 0.5, and the prediction at the root of the node is 0.4.  If the prediction at the child node was 0.5, then we're 0.1 closer to the right answer, and so we should get a positive contribution of 0.1 to the feature.  If the prediction at the child node was 0.6, then we're still 0.1 from the right answer, and so the contribution for this feature should be zero.  If the prediction at the child node was 0.7, then we've moved from 0.1 to 0.2 distant from the correct answer, and so we should get a -0.2 contribution for this feature.  I don't see this logic anywhere.

Note that we also need to transform the labels based upon the output functions of the classifier.  So if we pass through a logistic function on the output, we need to apply the inverse logistic function to the output label when passing to the internal logic.

mailletf:
I think there were actually two different ways of looking at it for regression, the one which is in the PR currently and what you said. Regression is different than classification since for classification we can use the sign as an indicator of helpful VS hurtful, and the number itself as an indicator of importance, while in regression I don't think we can.

What is implemented right now shows how we got to the prediction. So adding up the scores give us the prediction. The number gives an indication of importance, while the sign just says weather it pushed up or down; but that doesn't mean helpful/hurtful.

What you're saying if essentially giving out explain scores showing helpfulness and hurtfulness, where the sign says weather the feature pushed us towards or further away from the target. It intuitively makes sense in the case of the decision tree, where we apply cuts in sequence, and so we can look at where we are with regards to the target at each cut. If have a version of that where adding up the explain scores gives the residual. But I find it harder to see with the glz, where there is no concept of order of application. The equivalent I see would be something like distributing the residual to all features proportionally to their respective contribution (input val * glz weight).

But so what we loose in the 2nd way is our ability to differentiate between features that do nothing and those that are helping perfectly. Could the most important feature get a score of 0 since it makes us be spot on right away? So it seems like the two approaches may be complementary and maybe something we would want to do it have a different behaviour depending on if the user provides a label when calling the function. 

Since we need this to send out a notebook, I suggest I merge it as is, since it is "working". I'll open a second branch with what I see as the other way of doing it and we can iterate there while unblocking us right now. I'll merge at lunch if no ones objects cc @jeremybarnes

PR #275, opened 2016-03-22
[MDLB-1504] replaced a call to toString by one to toUtf8String in Row…

guyd:
…Name constructor


nicolaskruchten:
+1 once the tests pass

PR #277, opened 2016-03-22
Mldb 1384 where in outer join

steadtler:
…tcolumn.

getcolumn, getcolumnvalues and the like all assume an existing column and will throw if it doesnt exist, so if the column is user-provided, we need to check that its a valid column so that we can return 0 rows instead of a 404/500 http.

### DIFF
--- a/core/dataset.cc
+++ b/core/dataset.cc
@@ -723,22 +723,32 @@ executeFilteredColumnExpression(const Dataset & dataset,
                                 const ColumnName & columnName,
                                 const Filter & filter)
 {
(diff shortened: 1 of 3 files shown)


guyd:
I'm ok with the code but I am not sure why MLDB-1384 is a good thing.  If the expression in the where close cannot be evaluated, it evaluates to false.  I am afraid that this could hide user typos that would be hard to debug.  Any convincing argument?

guyd:
After discussion with Mathieu I am convinced!  +1

PR #278, opened 2016-03-22
MLDB-1508 json import improvements

jeremybarnes:
This removes many of the bottlenecks and inefficiencies in JSON importing.  It moves the speed up from parsing around 16,000 complex records per second to around 23,000.

### DIFF
--- a/sql/cell_value.cc
+++ b/sql/cell_value.cc
@@ -1053,6 +1053,115 @@ CellValue::printInterval() const
 
 }
 
+void
+CellValue::
+extractStructuredJson(JsonPrintingContext & context) const
+{
+    switch (cellType()) {
+    case CellValue::EMPTY:
+        context.writeNull();
+        return;
+    case CellValue::INTEGER:
+        if (isInt64())
+            context.writeLong(toInt());
+        else
+            context.writeUnsignedLongLong(toUInt());
+        return;
+    case CellValue::FLOAT:
+        {
+            double floatVal = toDouble();
+            if (std::isnan(floatVal))
+                {
+                    Json::Value v;
+                    v["num"] = std::signbit(floatVal) ? "-NaN" : "NaN";
+                    context.writeJson(v);
+                }
+            else if (std::isinf(floatVal))
+                {
+                    Json::Value v;
+                    v["num"] = std::signbit(floatVal) ? "-Inf" : "Inf";
+                    context.writeJson(v);
+                }
+            else
+                {
+                    context.writeDouble(toDouble());
+                }
+                
+            return;
+        }           
+    case CellValue::UTF8_STRING:
+    case CellValue::ASCII_STRING:
+        switch (type) {
+        case ST_UTF8_SHORT_STRING:
+            context.writeStringUtf8((const char *)shortString, (size_t)strLength);
+            return;
+        case ST_ASCII_SHORT_STRING:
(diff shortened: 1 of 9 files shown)


nicolaskruchten:
This is using `import.text` with `parse_json()` ?

jeremybarnes:
No it's import.json

nicolaskruchten:
oh... I was planning on retiring that procedure and recommending that we use import.text with parse_json() as * in the select parameter.

nicolaskruchten:
Doesn't import.json basically just do that under the hood anyway?

jeremybarnes:
Not really.  Import JSON has optimizations which will be harder to achieve with generic text import.  I agree with the goal though... In the end JSON is just text.

jeremybarnes:
In any case, this PR doesn't modify the importers directly.  It improves the efficiency of the underlying mechanisms.

nicolaskruchten:
Ok. I assumed that import.text would be faster because memory mapped..?
Does import.json do that trick too?

On Tuesday, 22 March 2016, Jeremy Barnes <notifications@github.com> wrote:

> In any case, this PR doesn't modify the importers directly. It improves
> the efficiency of the underlying mechanisms.
>
> —
> <https://github.com/mldbai/mldb/pull/278#issuecomment-200052199>
>


-- 
http://datacratic.com/

jeremybarnes:
Yes, that part of the machinery is common.  Although in the tests I'm running the file is compressed and so not memory mapped.

nicolaskruchten:
Ok. And what is the performance loss from import.json to import.text?

On Tuesday, 22 March 2016, Jeremy Barnes <notifications@github.com> wrote:

> Yes, that part of the machinery is common. Although in the tests I'm
> running the file is compressed and so not memory mapped.
>
> —
> <https://github.com/mldbai/mldb/pull/278#issuecomment-200083294>
>


-- 
http://datacratic.com/

jeremybarnes:
We can test it.  It may not be much.  There are other performance improvements (sharing column names and other schema elements) that would be harder to do in the more generic context though not impossible.

Alternatively we could keep two implementations buy switch them based upon matching a pattern in the select expression.  That way at least it becomes an implementation detail.

sql/cell_value.cc
```diff
@@ -1015,6 +1053,115 @@ CellValue::printInterval() const
...
+                context.newArrayElement();
+                context.writeInt(*p++);
+            }
+        }
+        context.endArray();
+        context.endObject();
+        return;
+    }
+    default:
+        throw HttpReturnException(400, "unknown cell type");
```
steadtler: Code 500?

sql/coord.cc
```diff
@@ -52,7 +52,15 @@ Coord(const char * str, size_t len)
 Coord::
 Coord(uint64_t i)
 {
-    initString(std::to_string(i));
+    // NOTE: std::to_string is SLOOOOW in GCC 4.8
```
steadtler: maybe share with the code in cellvalue.cc using a template?

steadtler:
Looks good, just a couple of suggestions but otherwise +1

sql/cell_value.cc
jeremybarnes: I'm leaving this one.  The code was simply moved.  There are lots of other places in this file that should probably be a 500; we can address them separately.

PR #280, opened 2016-03-23
Row names

steadtler:
MLDB-1505: 
default rowname for queries without from is 'result'
queries without from can use 'NAMED'
MLDB-1381
better error message when providing illegal row name

### DIFF
--- a/builtin/joined_dataset.cc
+++ b/builtin/joined_dataset.cc
@@ -648,7 +648,7 @@ struct JoinedDataset::Itl
         auto it = columnIndex.find(columnName);
 
         if (it == columnIndex.end())
-            throw HttpReturnException(400, "Joined dataset did not find column ",
+            throw HttpReturnException(500, "Joined dataset did not find column ",
                                       "columnName", columnName);
         
         auto doGetColumn = [&] (const Dataset & dataset,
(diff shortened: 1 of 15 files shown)


builtin/joined_dataset.cc
```diff
@@ -648,7 +648,7 @@ struct JoinedDataset::Itl
         auto it = columnIndex.find(columnName);
 
         if (it == columnIndex.end())
-            throw HttpReturnException(500, "Joined dataset did not find column with given hash",
+            throw HttpReturnException(400, "Joined dataset did not find column ",
```
guyd: I wonder if this should not stay a 500.  `getColumn` is called in many contests where some might not have been triggered directly by a user input.  Can we catch that error later down the stack and convert it to a 400 when it is clear that the user input is the source of the problem?

sql/sql_expression.cc
```diff
@@ -3619,20 +3619,27 @@ SelectStatement::parse(ML::Parse_Context& context, bool acceptUtf8)
...
 
+    bool needDefaultRowName = false;
     if (matchKeyword(context, "NAMED ")) {
         statement.rowName = SqlExpression::parse(context, 10, acceptUtf8);
         skip_whitespace(context);
     }
     else {
-        statement.rowName = SqlExpression::parse("rowName()");
+        //default name is "rowName()" if there is a dataset and "result" if there is none
+        //wo we need to check if there is a FROM first.
```
guyd: --wo--

guyd:
+1

PR #281, opened 2016-03-23
Remove rule for $(INC)/%: $(HOME)/local/include/%

jraby:


jeremybarnes:
I would also remove the MLDB_LOCAL_HEADER_FILES variable

jeremybarnes:
+1 once tests pass

PR #282, opened 2016-03-24
MLDB-1513 fix issues with tabular dataset and row generation in queries

steadtler:
-Fix a off-by one error in tabular dataset

-Made clear that generateRowsWhere must return an exact set

-`where` expression no longer evaluated outside of generateRowsWhere when generateRowsWhere is called. Should have a positive performance impact on large, non-group-by queries.

-Dont allow generateRowsWhere to cast other values to string when comparing to rowName(). This is because we evaluated rowName() to a string type when we evaluated SQL, so if we allow generateRowsWhere to cast ints to rownames in an optimized path, it would give a different result than the non-optimized path.

-Currently all row names are stored as strings in Coord and all rowName() evaluated to a string.

### DIFF
--- a/core/dataset.cc
+++ b/core/dataset.cc
@@ -838,7 +838,12 @@ generateRownameIsConstant(const Dataset & dataset,
             },
             "generate single row matching rowName()"};
 }
-    
+
+/*
+    Must return the *exact* set of rows or a stream that will do the same
+    because the where expression will not be evaluated outside of this method
+    if this method is called.
+*/    
 GenerateRowsWhereFunction
 Dataset::
 generateRowsWhere(const SqlBindingScope & scope,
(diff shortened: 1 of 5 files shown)


testing/MLDB-1428-text-sparse-output.py
```diff
@@ -84,4 +84,39 @@ def test_sparse_unknown_cols(self):
                     ["471242",1]]
         self.assertEqual(res, expected)
 
+    # MLDB-1513
+    def test_missing_rows(self):
```
mailletf: could those 2 test cases maybe be combined into one with a: for datasetType in ["tabular", "sparse.mutable"]: ?

then it would be clear that we're trying to control for identical behavior for two dataset types

mailletf:
mostly looked at the test. so +0.75 :P

core/dataset.cc
```diff
@@ -968,6 +973,11 @@ generateRowsWhere(const SqlBindingScope & scope,
 
                             for (auto& c : inExpression->tuple->clauses) {
                                 ExpressionValue v = c->constantValue();
+
+                                //casting other types to string will give a different result than non-optimized path.
+                                if (!v.isString())
```
guyd: This means that row names are always interpreted as string and they will never match a IN expression that is not a string, correct?

guyd: I would add a test with numerical row names in beh and IN expression.

steadtler: Effectively we are breaking this, because it was working *only* in some optimized case and not in the generic case - giving inconsistent result.

guyd:
+1

PR #283, opened 2016-03-24
[MLDB-1451] serving index.html in directories if it exists

guyd:
- changed the nginx config to serve all documents under doc/ and resources/
- changed the behaviour of the static file handler to look for an index.html if the uri points to a folder
- fixed a couple path building to make them more robust when ill-formed user input is provided


guyd:
@nicolaskruchten I believed I covered all the cases and my tests in the docker reveal no problem.  However, when using the mldb_doc_server it seems that using /doc will brake the frame display in index.html because the browser history is not replaced correctly in the javascript.  I can add better check there but I'd liked to get your input first as whether this a problem.

guyd:
After further thinking I realized that my test with the container were mostly likely incorrect since some of the pages were likely serve with nginx and not mldb.  The problem affecting the mldb_doc_server will  likely be problematic.

nicolaskruchten:
Capturing HipChat feedback: /doc should return a 3xx to /doc/ and the problem will go away I think.

guyd:
@jraby Can you please review the changes to nginx?

container_files/mldb_nginx_site.conf
```diff
@@ -87,12 +87,20 @@ server {
             return 301 $scheme://$http_host{{HTTP_BASE_URL}}/doc/;
         }
 
-        location {{HTTP_BASE_URL}}/doc/ {
-            # send documentation render requests to mldb (.md.html)
-            rewrite ^{{HTTP_BASE_URL}}(/?.*) $1 break;
-            try_files $uri $uri/ @mldb;
+        location = {{HTTP_BASE_URL}}/resources {
```
jraby: Not sure if this is required. That rule was in place so that people hitting `/doc` would get redirected correctly to `/doc/` with the correct port number in the URL.
But I don't think there is (or should be) any link pointing to `/resources` directly.

jraby:
nginx part looks ok to me, modulo the `/resource` rewrite rule which is possibly not needed.

container_files/mldb_nginx_site.conf
guyd: Indeed, I added it because I was manually testing redirect but I did not see a link to this URL.

guyd:
@nicolaskruchten I've reworked the code.  Redirect is done by mdlb except for uri `.../doc` and `.../resources` which is done by nginx (doing it in mldb required more changes than expected).  The mldb_doc_server works fine now including the interactive REST doc (not that it really matter!).

container_files/mldb_nginx_site.conf
```diff
@@ -87,12 +87,20 @@ server {
...
+            return 301 $scheme://$http_host{{HTTP_BASE_URL}}/resources/;
         }
 
+        # Documentation routes
+        location ~ {{HTTP_BASE_URL}}/(doc|resources)/ {
```
jraby: We need to add a `auth_request off;` at least for the `resources` route, otherwise the licensing page breaks.

nicolaskruchten:
In the docker, yes, but what about the raw `mldb_runner` or `mldb_doc_server` ?

guyd:
It returns a 404 because that route is not registered in mldb.

nicolaskruchten:
Ok, that's good. Any objections to getting rid of the doc_server
executable, given that mldb_runner does the same thing?

On Tuesday, 29 March 2016, Guy Dumais <notifications@github.com> wrote:

> It returns a 404 because that route is not registered in mldb.
>
> —
> You are receiving this because you were mentioned.
> <https://github.com/mldbai/mldb/pull/283#issuecomment-203086496>
>


-- 
http://datacratic.com/

guyd:
I've certainly have no objections to remove the mldb_doc_server.

nicolaskruchten:
OK, let's remove the doc_server as part of this PR then please, and +1 from me :)

nicolaskruchten:
@jraby once this is merged, you should be clear to work on MLDB-1446

container_files/mldb_nginx_site.conf
guyd: I removed the redirect from `/resources` to `/resources`.

container_files/mldb_nginx_site.conf
guyd: I removed the auth in `/doc` and kept the auth to off for `/resources`

guyd:
I removed the mldb_doc_server as agreed.

PR #284, opened 2016-03-25
[MLDB-1490] added validation of select clauses in presence of aggrega…

guyd:
Added validation of the select parse tree in presence of group by or aggregator.  Output a very specific error message when such case occur.


steadtler:
+1

PR #285, opened 2016-03-26
Made RowStream destructor virtual (MLDB-1520) (TRIVIAL)

jeremybarnes:
Trivial addition of a virtual destructor, to allow subclasses to actually free their resources.


guyd:
+1

PR #286, opened 2016-03-29
Mldb 15xx perf improvements

jeremybarnes:
A bunch of minor SQL performance improvements.

Includes an interface improvement and minor bug fix for forEachLineBlock.

### DIFF
--- a/sql/expression_value.cc
+++ b/sql/expression_value.cc
@@ -2286,7 +2286,13 @@ getFiltered(const VariableFilter & filter) const
     if (filter == GET_ALL)
         return *row_;
     
-    std::function<bool(const ExpressionValue&, const ExpressionValue&)> filterFn = [](const ExpressionValue& left, const ExpressionValue& right){return false;};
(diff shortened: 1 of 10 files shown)


sql/expression_value.cc
```diff
@@ -2022,22 +2022,27 @@ appendToRowDestructive(ColumnName & columnName, RowValue & row)
...
+                    return true;
+                };
+            
+            forEachColumnDestructiveT(onSubexpr);
+        }
+        else {
+            auto onSubexpr = [&] (ColumnName & innerColumnName,
+                                  ExpressionValue & val)
+                {
+                    ColumnName newColumnName = columnName + innerColumnName;
```
guyd: Is the period gone?

jeremybarnes: No, the Coord::operator + puts the period in by itself.

guyd:
There is a fair amount of code duplication in expression_value but we can fix it later.

+1

PR #287, opened 2016-03-30
MLDB-1507 bid request import

jeremybarnes:
This improves the importation of JSON-structured data like bid requests.  It has the following two main effects:

1.  Bid request import throughput goes from ~15,000 per second to ~75,000 per second
2.  Memory usage of bid requests is roughly 2x the gzipped file size, rather than roughly 50x the gzipped file size

It works by adding the ability for the Tabular dataset to have an open schema, and adding a Recorder interface which can be requested by something that's bulk inserting and can be specialized when the schema (or part of the schema) is known.


container_files/public_html/doc/builtin/datasets/TabularDataset.md
```diff
@@ -1,7 +1,44 @@
...
+The tabular dataset has the following limitations:
+
+- Each row has a single timestamp; it is not possible for different values
+  in the row to have separate timestamps
+- Each column can only have a single value within a single row
+- The dataset will work well up to tens of thousands of columns, but for
+  extremely sparse data it will not be efficient due to a per-column
+  overhead.  It's better to use a sparse dataset for these situations.
+- It may only be committed once, and will not be queryable until it is
+  committed the first time.  As a reusult, this dataset type is mostly
```
steadtler: reusult

core/dataset.cc
```diff
@@ -405,7 +405,57 @@ getColumnBuckets(const ColumnName & column,
 
 
 /*****************************************************************************/
-/* DATASET                                                                    */
+/* DATASET RECORDER                                                          */
+/*****************************************************************************/
+
+// This is here to allow future extension without breaking the ABI
```
steadtler: ABI -> API ? Unless there is an acronym I dont know

core/recorder.cc
```diff
@@ -0,0 +1,93 @@
...
+    Copyright (c) 2016 Datacratic Inc.  All rights reserved.
+*/
+
+#include "recorder.h"
+
+namespace Datacratic {
+namespace MLDB {
+
+
+/*****************************************************************************/
```
steadtler: All the arrays in Recorder are passed by copy, is that on purpose?

core/recorder.h
```diff
@@ -0,0 +1,127 @@
...
+        Extra allows for unexpected columns to be recorded, for when the
+        dataset contains some fixed columns but also may have dynamic
+        columns.
+    */
+    virtual
+    std::function<void (RowName rowName,
+                        Date timestamp,
+                        CellValue * vals,
+                        size_t numVals,
+                        std::vector<std::pair<ColumnName, CellValue> > extra)>
```
steadtler: does extra also need  to be passed by copy?

jeremybarnes: Yes, since we will destroy it we need to pass by copy.  The caller needs to std::move the argument if it's destroyable or pass it directly (which will invoke the copy constructor) if not.  Note that the values and rowname will also be destroyed, so it's consistent over the interface.

steadtler:
Looks good to me +1 but someone else should definitely look at it.

guyd:
I will take a quick look at it.

core/recorder.h
```diff
@@ -0,0 +1,127 @@
...
+    specializeRecordTabular(const std::vector<ColumnName> & columns);
```
guyd: This is weird.  The base class contains a specialization for a specific dataset.  It looks like the abstraction is not complete.

plugins/column_types.cc
```diff
@@ -0,0 +1,121 @@
...
+/*****************************************************************************/
+
+/** This is an accumulator that keeps statistics on the types of values that
+    a column could have.  It's useful for knowing how to treat a column in
+    an algorithm.
+*/
+
+ColumnTypes::   
+ColumnTypes()
+    : numNulls(false), numIntegers(false),
```
guyd: If these are counters as they seem to be, I would initialize with 0 not false.

plugins/column_types.cc
```diff
@@ -0,0 +1,121 @@
...
+ColumnTypes::   
+getExpressionValueInfo() const
+{
+    if (!numNulls && !numReals && !numStrings && !numOther) {
+        // Integers only
+        if (minNegativeInteger == 0) {
+            // All positive
+            return std::make_shared<Uint64ValueInfo>();
+        }
+        else if (maxPositiveInteger <= (1ULL << 63)) {
```
guyd: shouldn't that check be also done in the if above?

plugins/column_types.cc
```diff
@@ -0,0 +1,121 @@
...
+            && minNegativeInteger > -(1LL << 53)) {
+            return std::make_shared<Float64ValueInfo>();
+        }
+        // Doubles would lose precision.  It's an atom.
+        return std::make_shared<AtomValueInfo>();
+    }
+    else if (!numNulls && !numIntegers && !numReals && !numOther) {
+        return std::make_shared<Utf8StringValueInfo>();
+    }
+    else {
```
guyd: blobs do not seem to be supported here

plugins/frozen_column.cc
```diff
@@ -0,0 +1,297 @@
...
+    }
+
+    virtual bool forEachDistinctValue(std::function<bool (const CellValue &, size_t)> fn) const
+    {
+        if (hasNulls) {
+            if (!fn(CellValue(), columnTypes.numNulls))
+                return false;
+        }
+        for (auto & v: table) {
+            if (!fn(v, 1 /* todo: real count */))
```
guyd: Should this be completed before committing?

plugins/frozen_column.cc
```diff
@@ -0,0 +1,297 @@
...
+
+        return result;
+    }
+
+    virtual bool forEachDistinctValue(std::function<bool (const CellValue &, size_t)> fn) const
+    {
+        if (!fn(CellValue(), columnTypes.numNulls))
+            return false;
+        for (auto & v: table) {
+            if (!fn(v, 1 /* todo: real count */))
```
guyd: Same here, should we complete that before committing?

plugins/importtext_procedure.cc
```diff
@@ -1039,113 +881,83 @@ struct ImportTextProcedureWorkInstance
...
-	                
-	        auto row = scope.bindRow(&values[0], ts, actualLineNum, 0 /* todo: chunk ofs */);
+	        auto row = scope.bindRow(&values[0], ts, actualLineNum,
+                                         0 /* todo: chunk ofs */);
 
 	        // If it doesn't match the where, don't add it 
 	        if (!isWhereTrue) {
 	            ExpressionValue storage;
-	            if (!whereBound(row, storage, GET_LATEST).isTrue())
+	            if (!whereBound(row, storage, GET_ALL).isTrue())
```
guyd: I am surprised that we require `GET_ALL` here.

plugins/importtext_procedure.cc
```diff
@@ -1039,113 +881,83 @@ struct ImportTextProcedureWorkInstance
...
 	                return true;
 	        }
 	            
 	        // Get the timestamp for the row
 	        Date rowTs = ts;
 	        ExpressionValue tsStorage;
-	        rowTs = timestampBound(row, tsStorage, GET_LATEST).coerceToTimestamp().toTimestamp();
+	        rowTs = timestampBound(row, tsStorage, GET_ALL)
```
guyd: same here and below.

plugins/tabular_dataset.cc
```diff
@@ -20,6 +31,65 @@ namespace Datacratic {
...
+
+namespace {
+
+// Note: in GCC 4.9+, we can use the std::atomic_xxx overloads for
+// std::shared_ptr.  Once the Concurrency TR is available, we can
+// replace with those classes.  For the moment we use a simple,
+// spinlock protected implementation that is a lowest common
+// denominator.
+template<typename T>
+struct atomic_shared_ptr {
```
guyd: Consider moving this code to a separate header.

plugins/tabular_dataset.cc
```diff
@@ -294,7 +474,7 @@ struct TabularDataset::TabularDataStore: public ColumnIndex, public MatrixView {
     {
         auto result = tryLookupRow(rowName);
         if (result.first == -1)
-            throw HttpReturnException(400, "Row not found in CSV dataset");
+            throw HttpReturnException(400, "Row not found in tabular dataset");
```
guyd: Consider adding the row name for these 400 error messages.

plugins/tabular_dataset.cc
```diff
@@ -425,83 +637,513 @@ struct TabularDataset::TabularDataStore: public ColumnIndex, public MatrixView {
         for (unsigned i = 0;  i < chunks.size();  ++i) {
             for (unsigned j = 0;  j < chunks[i].rowNames.size();  ++j) {
                 if (!rowIndex.insert({ chunks[i].rowNames[j], { i, j } }).second)
-                    throw HttpReturnException(400, "Duplicate row name in CSV dataset",
+                    throw HttpReturnException(400, "Duplicate row name in text dataset",
```
guyd: text -> tabular

testing/MLDB-749-csv-dataset.js
```diff
@@ -106,7 +106,7 @@ mldb.put("/v1/procedures/csv_proc", csv_conf)
 res = mldb.put("/v1/procedures/csv_proc/runs/0", {})
 mldb.log(res);
 assertEqual(res['responseCode'], 400);
-assertEqual(res['json']['error'], "Duplicate row name in CSV dataset");
+assertEqual(res['json']['error'], "Duplicate row name in text dataset");
```
guyd: text -> tabular

guyd:
+1
I know it is often difficult to plan this ahead but when we know that some code will be moved, it would be better to move the code first and request a review after of the modification after the move.  I realized that I spent time reviewing old code as if it was newly added.

core/recorder.h
jeremybarnes: It's not a specialization for a specific dataset.  It's a specialization for a particular data layout.  Any dataset that can record could improve performance by implementing a specialization, as it allows us to avoid manipulating column names.

plugins/column_types.cc
jeremybarnes: They are counters, and should be initialized with false.  Their values are important later on when we look for optimization opportunities in how we store the column's data.

plugins/frozen_column.cc
jeremybarnes: Yes, we should remove the count field.

plugins/importtext_procedure.cc
jeremybarnes: We don't require GET_ALL.  However, in the context, GET_ALL is far, far faster than GET_LATEST, and it's not possible for a row to have more than one timestamp.  I have difficulty imagining how this could cause a failure.

jeremybarnes:
@guyd thanks for the review.  Apologies for the moved code; on the other hand it's probably not such a bad thing to have an extra set of eyes on it.

plugins/importtext_procedure.cc
guyd: `GET_ALL` is certainly faster.  We can certainly think of an obscure where condition that would evaluate to true with `GET_ALL` but not with `GET_LATEST` (or vice-versa).  I would say let's wait until this obscure condition becomes real and problematic.

plugins/tabular_dataset.cc
```diff
@@ -607,25 +1230,70 @@ commit()
...
 };
 
-DECLARE_STRUCTURE_DESCRIPTION(TabularDatasetConfig);
 DEFINE_STRUCTURE_DESCRIPTION(TabularDatasetConfig);
 
 TabularDatasetConfigDescription::
 TabularDatasetConfigDescription()
 {
     nullAccepted = true;
+
```
nicolaskruchten: I would consider adding a 'columns' array as a parameter as well, which would in effect default to the column set of the first row.

I assume that if the second row has fewer columns than the first, 'null' is inferred for the missing columns?

container_files/public_html/doc/builtin/datasets/TabularDataset.md
```diff
@@ -1,7 +1,44 @@
 # Tabular Dataset
 
-The Tabular Dataset is used to represent dense datasets with more rows than columns. It is ideal for storing text files such as Comma-Separated Values (CSV) files.
+The Tabular Dataset is used to represent dense datasets with more rows than
+columns. It is ideal for storing text files such as Comma-Separated Values
+(CSV) files.
+
+The dataset will learn the available columns from the first data that is
```
nicolaskruchten: this is no longer strictly true with the `unknownColumns` parameter

plugins/tabular_dataset.cc
jeremybarnes: Yes, the missing columns will have nulls inferred.

The issue with that parameter is that it will be confusing for a user with respect to the importer function and the select expression.  We're giving them 3 different places to write column names, without any clear way to know where to put them.

nicolaskruchten: That's true. On balance, I would still favour adding it, but i this works without adding it now, let's just go with it and we can add it later if it becomes important. It's not a useful feature in the context of import.text or import.json but possibly important in the context of creating a dataset and inserting rows one at a time. If you don't want to use unknownColumns=add and you can't guarantee that the first row recorded will be representative of the rest, this will be necessarily.

core/recorder.cc
jeremybarnes: Yes.  They are copied on the Destructive version, as this allows the calling code to construct the objects which can then be moved into place by the implementation, and gives the possibility for the calling code to move them directly if it can.

The non-Destructive versions take references.

jeremybarnes:
This should be ready to go, if someone wants to give it a final +1.

guyd:
Reviewed rapidly the additional commits. +1

PR #289, opened 2016-03-30
MLDBFB-440 test issue 400 on ds wo cols

FinchPowers:


testing/MLDBFB-440_error_on_ds_wo_cols.py
```diff
@@ -0,0 +1,24 @@
...
+class TestErrorOnDsWoCols(MldbUnitTest):  # noqa
+
+    @unittest.expectedFailure
+    def test_it(self):
+        ds = mldb.create_dataset({'id' : 'noColDs', 'type' : 'sparse.mutable'})
+        ds.record_row('row1', [])
+        ds.commit()
+
+        res = mldb.query("SELECT sum({*}) FROM noColDs")
+        self.assertEqual(res[1][1], 0)
```
nicolaskruchten: not sure I agree. the sum of nothing is not 0, it's nothing...

FinchPowers: It can be updated to null, nan, or whatever you want. It currently assert fails and returns with a 400.

testing/MLDBFB-440_error_on_ds_wo_cols.py
```diff
@@ -0,0 +1,24 @@
...
+import unittest
+
+mldb = mldb_wrapper.wrap(mldb)  # noqa
+
+class TestErrorOnDsWoCols(MldbUnitTest):  # noqa
+
+    @unittest.expectedFailure
+    def test_it(self):
+        ds = mldb.create_dataset({'id' : 'noColDs', 'type' : 'sparse.mutable'})
+        ds.record_row('row1', [])
```
nicolaskruchten: @FinchPowers @jeremybarnes does it make sense that this call succeeds?

FinchPowers: My rationale is: I can apply a transform over a dataset in such a way that some of the rows no longer have columns.
Hence, we should be able to create empty rows directly.

nicolaskruchten: ok.

PR #290, opened 2016-03-30
MLDBFB-441 distinct_timestamps bad ts value kept test case

FinchPowers:


FinchPowers:
On hold. I could fix test_MLDB_1509 instead.

testing/MLDB-724-time-arithmetic.py
```diff
@@ -794,6 +794,24 @@ def test_MLDB_1509(self):
...
+        ds.commit()
+        res = mldb.get('/v1/query',
+                       q="SELECT distinct_timestamps(colA) FROM ds")
+        mldb.log(res)
+        self.assertFullResultEquals(res.json(), [{
+            "rowName": "row1",
+            "columns": [
+                ["distinct_timestamps(colA).0",
+                 {"ts":"1970-01-01T00:00:00Z"},
+                 "1970-01-01T00:00:00Z"],  # <--- MLDB returns the 5th second
```
guyd: This is expected.  We returned the list of known distinct timestamp at the latest timestamp.

nicolaskruchten: Yes, this is expected. The result of the `distinct_timestamps()` function is an embedding (an array) of timestamps (those are the values) and the timestamps of the timestamps are all set to the latest timestamp.

timestamp timestamp timestamp

testing/MLDB-724-time-arithmetic.py
```diff
@@ -604,21 +613,21 @@ def test_MLDB_1509(self):
                            {
                                "ts": "2015-01-01T00:00:00Z"
                            },
-                           "2015-01-03T00:00:00Z"
+                           "2015-01-01T00:00:00Z"
```
FinchPowers: Otherwise we lose the power of the "WHEN" operator.

FinchPowers:
@nicolaskruchten Good to go now.

testing/MLDB-724-time-arithmetic.py
nicolaskruchten: can you elaborate?

FinchPowers: Edit: We lose the *some of the power* of the "WHEN" operator in the particular use case that interrests me.

With the current implementation, that cell/value is kept with a `WHEN value_timestamp() > to_timestamp('2015-01-02T00:00:00Z')`. I want it to be discarded.

I'm investigating alternatives. I can probably push the distinct_timestamps usage later in the pipeline at the cost of  performance.

PR #292, opened 2016-04-01
Join sort optims

steadtler:
Added a new algo to perform a quicksort in multithread.

On fmaillet's test case of a left join on 41M+ rows (on one side, less on the other), with this PR we save:

About 60 seconds by sorting in parallel in join_dataset
About 173 seconds (3 minutes!) by sorting in parallel in QueryBasic
About 20 seconds by not sorting in TabularDataset::getrownames

In total we cut the time necessary to perform this large outer join by more than half. There may be a larger refactoring to make it much faster but those were the low-hanging apples.

### DIFF
--- a/core/dataset.h
+++ b/core/dataset.h
@@ -60,7 +60,7 @@ typedef EntityType<Dataset> DatasetType;
 struct MatrixView {
     virtual ~MatrixView();
 
-    /*
+    /**
     Return a list of all rownames.
     Rownames are always unique.
 
@@ -437,7 +437,8 @@ struct Dataset: public MldbEntity {
         because the where expression will not be evaluated outside of this method
         if this method is called.
 
-        Ordering can be arbitrary but need to be deterministic
+        Ordering can be arbitrary but needs to be deterministic, and there must not
+        be duplicated rows.
     */    
 
     virtual GenerateRowsWhereFunction
--- a/core/dataset.h
+++ b/core/dataset.h
@@ -60,6 +60,18 @@ typedef EntityType<Dataset> DatasetType;
 struct MatrixView {
     virtual ~MatrixView();
 
+    /*
+    Return a list of all rownames.
+    Rownames are always unique.
+
+    The sorting criteria is the same as with RowStream:
+
+    Row names can be returned in an arbitrary order as long as it is deterministic.
+    I.e. Calling getRowNames several times on the same (unchanged) dataset should return rownames
+    in the same (arbitrary) order.
+
+    The ordering needs to be preserved regardless of start and limit.
+    */
     virtual std::vector<RowName>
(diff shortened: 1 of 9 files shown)


plugins/tabular_dataset.cc
```diff
@@ -257,13 +257,6 @@ struct TabularDataset::TabularDataStore: public ColumnIndex, public MatrixView {
             result.insert(result.end(), c.rowNames.begin(), c.rowNames.end());
         }
 
-        std::sort(result.begin(), result.end(),
```
jeremybarnes: This one makes me nervous.

jeremybarnes:
It all looks good, but I wonder about not sorting the row names.  If we don't here, we shouldn't anywhere and we should then not expect anything to be sorted in the rest of the code.

steadtler:
Its my next story to remove it everywhere else. I can wait until that story is done if you want.

jeremybarnes:
I'm OK to merge it, so long as you document in dataset.h what the requirement on the method is, and also in the places where it's called.  You should also document the ordering requirement for the RowStream.

steadtler:
@jeremybarnes can you review the updated code documentation?

guyd:
I have a related question, do all datasets have a natural row ordering (e.g. the order in which recordRow[s] was called)?  If so, what is that natural order and do `MatrixView::getRowNames` and `ColumnIndex::getRowNames` return rows in that natural order.

steadtler:
@guy 

Yes, they should all have a natural ordering of sorts. getrownames can return the rows in any order as long as successive calls will return them in the same order.

guyd:
So the order is dataset-dependent but consistent for a given dataset.

core/dataset.h
```diff
@@ -60,6 +60,18 @@ typedef EntityType<Dataset> DatasetType;
 struct MatrixView {
     virtual ~MatrixView();
 
+    /*
```
jeremybarnes: /** for Doxygen

core/dataset.h
```diff
@@ -420,6 +436,8 @@ struct Dataset: public MldbEntity {
         Must return the *exact* set of rows or a stream that will do the same
         because the where expression will not be evaluated outside of this method
         if this method is called.
+
+        Ordering can be arbitrary but need to be deterministic
```
jeremybarnes: needs to be, missing fullstop (minor)

jeremybarnes: Also, we should say that they must not be duplicated.

jeremybarnes:
I had some minor comments on the doc.  My +1 still stands once you've lokoed at them.

In the future, we should look at having each dataset be able to expose its natural row order.  This can lead to all kinds of optimizations, especially for joining of two datasets that have the same natural order.

PR #297, opened 2016-04-05
[MLDB-1556] fixed the parsing in CellValue to ensure that we have a n…

guyd:
…ull terminated string

### DIFF
--- a/sql/cell_value.cc
+++ b/sql/cell_value.cc
@@ -199,14 +199,22 @@ CellValue
 CellValue::
 parse(const char * s_, size_t len, StringCharacteristics characteristics)
 {
-    std::string cstring(s_, len); // ensure the buffer is null terminated
-    const char * s = cstring.c_str();
+    static constexpr size_t NUMERICAL_BUFFER = 64;
+
+    // if the string is longer than 64 characters it can't realistically
+    // be a numerical value
+    if (len > NUMERICAL_BUFFER)
+        return CellValue(s_, len, characteristics);
 
     if (len == 0)
         return CellValue();
 
+    // this ensures that our buffer is null terminated as required below
+    char s[NUMERICAL_BUFFER + 1] = {0};
+    memcpy(s, s_, len);
+
     // First try as an int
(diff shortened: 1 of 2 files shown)


steadtler:
+1

sql/cell_value.cc
```diff
@@ -199,14 +199,22 @@ CellValue
...
+    // if the string is longer than 64 characters it can't realistically
+    // be a numerical value
+    if (len > NUMERICAL_BUFFER)
+        return CellValue(s_, len, characteristics);
 
     if (len == 0)
         return CellValue();
 
+    // this ensures that our buffer is null terminated as required below
+    char s[NUMERICAL_BUFFER + 1] = {0};
```
jeremybarnes: Instead of initializing to zero then memcpy, do the memcpy and then set the final byte to zero

sql/cell_value.cc
```diff
@@ -199,14 +199,22 @@ CellValue
 CellValue::
 parse(const char * s_, size_t len, StringCharacteristics characteristics)
 {
-    std::string cstring(s_, len); // ensure the buffer is null terminated
-    const char * s = cstring.c_str();
+    static constexpr size_t NUMERICAL_BUFFER = 64;
+
+    // if the string is longer than 64 characters it can't realistically
+    // be a numerical value
+    if (len > NUMERICAL_BUFFER)
```
jeremybarnes: I guess it's OK, though we should clearly document in the natural_cast documentation that numbers must have a length < 64, and put an issue (soon!) to fix it.  For example, 0.0000000000000000000000000000000000000000000000000000000000000000000000000000000023942190 is a valid number that could easily be produced by a bad format string.

jeremybarnes:
+1 with some comments above.

sql/cell_value.cc
nicolaskruchten: MLDBFB-479 created for this

guyd: I had already created MLDB-1558 for that.

PR #298, opened 2016-04-06
Check for 'modified' and 'new file' in submodules

jraby:
The original code would break in mysterious ways when the only change to
a submodule was new files and when using DOCKER_ALLOW_DIRTY=1.
Terminal would stay red and build would continue even if revision could
not be obtained.


jraby:
When a submodule only has 'new file'  changes,
`cat $SMFILESUSED | xargs git status -uno | grep 'modified:' > /dev/stderr `  would fail because 'modified'  was not found in the output.
The get_git_revision.sh  script would abort without setting the git revision and the build would continue because `DOCKER_ALLOW_DIRTY` is set.

The build would fail much later when trying to tag the docker image.

jeremybarnes:
+1

PR #299, opened 2016-04-07
Fixed empty literal parsing; improved error message for table expressions (MLDB-1567)

jeremybarnes:
This also improves the error message for when table expressions are mis-parsed to be more helpful to the user.

### DIFF
--- a/sql/sql_expression.cc
+++ b/sql/sql_expression.cc
@@ -960,17 +960,6 @@ bool matchConstant(ML::Parse_Context & context, ExpressionValue & result,
     }
 
     else return false;
-
-#if 0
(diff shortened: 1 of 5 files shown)


sql/sql_expression.cc
```diff
@@ -3111,6 +3103,18 @@ parse(ML::Parse_Context & context, int currentPrecedence, bool allowUtf8)
 
     std::shared_ptr<TableExpression> result;
 
+    auto expectCloseParanthesis = [&] ()
```
steadtler: parenthesis?

steadtler:
ok, just minor spelling +1

PR #301, opened 2016-04-08
Fixed use of context in optimization of IN (KEYS OF ...) (MLDB-1563)

jeremybarnes:
This fixes the issue whereby we couldn't find functions inside the rowName() IN (KEYS OF ...) expression when it hit the optimized path for rowName().


guyd:
+1

PR #304, opened 2016-04-08
added const to char *

mikeburkat:


jeremybarnes:
Our first manifestation of constexpr is not const++

jeremybarnes:
(+1)

PR #306, opened 2016-04-08
MLDB-1538 threadpool

jeremybarnes:
Two improvements to the thread pool:

1.  We are more aggressive at backing off busy stealing of work when there is no more work to do, to reduce idle CPU usage (plus test): MLDB-1538
2.  We document, and enforce, a no job exception policy for jobs submitted to the thread pool: MLDB-1532.  Note that the abstractions built on the thread pool, like parallelMap, still return exceptions; it's just that the jobs they launch catch and track them internally.

### DIFF
--- a/base/testing/thread_pool_test.cc
+++ b/base/testing/thread_pool_test.cc
@@ -25,6 +25,21 @@
 using namespace std;
 using namespace Datacratic;
 
+BOOST_AUTO_TEST_CASE (thread_pool_idle_cpu_usage)
+{
+    ThreadPool threadPool(32);
+    // let it start up and settle down
+    std::this_thread::sleep_for(std::chrono::seconds(1));
+    ML::Timer timer;
+    std::this_thread::sleep_for(std::chrono::seconds(1));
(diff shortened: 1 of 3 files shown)


steadtler:
Why the no exception policy?

jeremybarnes:
Because there is no way to reasonably handle an exception from a job launched asynchronously, and return it to the caller, so the caller needs to implement the exception policy that makes sense in their callback.  It also ensures that the same thing happens whether a job is executed synchronously or asynchronously.

In our experience with RTBkit, the "exceptions were silently eaten by the thread pool" behaviour made it almost impossible to debug issues.

steadtler:
Ok +1. Whats the preferred pattern if we encounter an error in a job?

jeremybarnes:
The pattern is to catch the exception within the job, and propagate it however makes sense for the application context.  If it's a synchronous job, then it should probably be rethrown by the calling process.  If it's asynchronous, it should use whatever failure/error reporting procedure is normally used.

PR #307, opened 2016-04-11
MLFBFB-470 py unjoined threads crash test case

FinchPowers:


testing/MLFBFB-470_python_unjoined_threads_crash.py
```diff
@@ -0,0 +1,33 @@
...
+
+    def test_join(self):
+        t = threading.Thread(target=thread_fct)
+        t.start()
+        t.join()
+
+    @unittest.skip("MLDBFB-470")
+    def test_no_join(self):
+        t = threading.Thread(target=thread_fct)
+        t.start()
```
nicolaskruchten: @jeremybarnes is this in the class of things that Python's model makes very difficult to solve? or can we envision a solution here?

jeremybarnes:
Test passes since Python refactor.

nicolaskruchten:
!!! :)

PR #309, opened 2016-04-12
Mldb 1245 query order

steadtler:
Performance improvements.
Dont sort output by rowHash unless required by the user.
Dont specify if you allow BoundQuery to use threads as the ThreadPool manages it better now.
Allow caller to specify is he wants its output in parallel (and any order) or not (and in a stable order).


server/bound_queries.cc
```diff
@@ -163,22 +165,40 @@ struct UnorderedExecutor: public BoundSelectQuery::Executor {
...
+            if (aggregateInParallel) {
+
+                parallelMap(offset, upper, doRow);
             }
+            else
+            {
+                //Todo: to reduce memory usage, we should fill blocks of output on worker threads
+                // in order as much as possible
+                // while calling the aggregator on the caller thread.
+                std::vector<std::tuple<NamedRowValue, std::vector<ExpressionValue> > > output(upper-offset);
```
guyd: The validation that offset >= 0 and offset <= upper is probably done elsewhere but if it is not, I suggest we add an assert here before the allocation of upper - offset.

guyd:
I very much like the fact that now it is crystal clear if the callback function of the dataset iterator is run in parallel or not.  
Things that I am less comfortable with:
- I was hoping that we would not need to do a rowHash ordering at any place.  My expectation was either we would output out of order (-> faster) or in the natural order of the dataset (whatever it is).   Do we need to still sort because otherwise the speed of some operations (e.g. join) would be unacceptable?
- The naming of the callback function of the dataset iterator, `aggregator` is somewhat misleading because this is often a function that do not aggregate anything (e.g export.csv).
- I could not pin point if the memory usage when processing large datasets was fixed.

steadtler:
@guyd 

- I removed the sorts that were there to enforce the _ordering_, but I left the sorts that are there to enforce determinism (like in the rowgenerator, in some cases) and sorts that are necessary algorithmically (like in the join, in the dataset.a = dataset.b case).

- Suggestion?

- No, its is not.

guyd:
Regarding the naming of the callback function, I would suggest  `RowCallback`.

guyd:
+1

PR #311, opened 2016-04-13
MLFBFB-474 test showing a dataset query too long to run

FinchPowers:
I didn't have the courage to run it through. It was still runing after 5 minutes.


FinchPowers:
This is using a beh file and should be moved to PRO.

PR #315, opened 2016-04-15
MLDB-1574 check alias in TableLexicalScope::doGetAllColumns

steadtler:
This is to unblock Alexis while waiting for the larger refactoring.


jeremybarnes:
+1

PR #316, opened 2016-04-15
Mldb 1552 where and rowname optim

steadtler:


core/dataset.cc
```diff
@@ -944,38 +945,159 @@ generateRowsWhere(const SqlBindingScope & scope,
...
+        if (comparison) {
+
+            auto clhs = getConstant(*comparison->lhs);
+            auto crhs = getConstant(*comparison->rhs);
+            auto flhs = getFunction(*comparison->lhs);
+            auto frhs = getFunction(*comparison->rhs);
+            auto blhs = getBoundParameter(*comparison->lhs);
+            auto brhs = getBoundParameter(*comparison->rhs);
+
+            if (frhs && (clhs || blhs) && comparison->op == "!=") {
```
guyd: Is there already an optimization for '=='?

guyd:
+1

core/dataset.cc
steadtler: There is for the generic case. in the case of "blah AND rowName() == constant", you would probably want to do something different though.

PR #317, opened 2016-04-16
Test and fix for thread jobs disappearing with thread exit (MLDB-1579)

jeremybarnes:
If we created a thread, submitted jobs, and then the thread exited before we had a chance to run the jobs, those jobs would be lost.  Instead, when a thread exits we wait for its queue to empty before doing so.

This fixes the issue with some tests in pro hanging on exit.

The test failed before the fix was made.


guyd:
+1

PR #318, opened 2016-04-20
[MLDB-1468] implemented the lookup of credential using the most speci…

guyd:
…fic associated resource


soa/credentials/credentials.h
```diff
@@ -16,41 +16,35 @@ namespace Datacratic {
 
 struct Credential {
     std::string provider; ///< Path through which credential was obtained
-    std::string protocol; ///< Protocol to use to get to service
-    std::string location; ///< URI to call to get resource
-    std::string id;       ///< User ID
+    std::string protocol; ///< Protocol to use to get to service (e.g. http, https)
```
jeremybarnes: Consider Utf8 and Url types where appropriate

guyd: I added this to MLDBFB-492.

jeremybarnes:
I am concerned about:
- No high-level documentation of what the credential system design is (I know there already wasn't any, but here we're changed some things and we should document what the design requirements are)
- Layering violations.  The credentials system is tied into the vfs layer, but the code has been implemented in the MLDB layer.  I would push for the credentials code and base interfaces to be pushed down to the vfs layer, with only the application-layer functionality exposed in the MLDB layer.  That would also help with unit testability.  Credentials don't need to know anything about MLDB and so shouldn't mention it in their code.
- Lacking a list of use-cases / requirements and tests that would expose them.  It's fine that these start off simple, but they may include things like knowing where an invalid/failed credential came from, etc.

None of these are necessarily blockers, but I think that at minimum we should address the layering violations by pushing code down.

guyd:
@jeremybarnes 

- I see two audiences for the documentation I am going to write: the typical end-user (e.g. the one using the hub or the provided OVA/AMI/docker image) and the expert user (this team and anyone with enough expertise to build and modify MLDB).  For the first one, I will augment the documentation in Url.md and give a specific example.  For the second one, I will augment the section about credentials in https://github.com/mldbai/mldb/blob/master/Building.md to discuss the different ways credentials can be provided and what the effect of each.  Would this cover the concern?
- the code in MLDB layer (under server folder)  is the credential collection and an associated credential provider fetching from the collection.  I believe this code belong there along with the other collections.  Note that this code does not understand anything about credentials other than the fields that need to be provided to store them.  I also moved `StoredCredentials` from the server folder to soa.  All the rest of the code already belong to files is soa.  Unless I am missing something, I think this is already the desired split.
- I agree that at least for debugging we should be able to know where the credentials are coming from.  I added two messages to cerr but this is clearly not the end of it.  I will add a JIRA for this specifically.

nicolaskruchten:
Is there more blocking discussion to be had here?

server/credential_collection.cc
```diff
@@ -97,43 +67,36 @@ struct CollectionCredentialProvider: public CredentialProvider {
...
+
+        // getCredentialsOfType called after destruction of the collection!
+        ExcAssert(sharedRules);
+
+        vector<StoredCredentials> matchingCreds;
+        if (sharedRules) {
+            auto keys = sharedRules->getKeys();
+            for (const auto & key : keys) {
+                auto cred = sharedRules->tryGetEntry(key);
+                ExcAssert(!cred.second); // still under construction!
```
mailletf: Can a user ever trigger this?

soa/credentials/credentials.cc
```diff
@@ -19,8 +19,8 @@ CredentialDescription::
 CredentialDescription()
 {
     addField("provider", &Credential::provider,
-             "Provider of credentials.  This is normally an address on the internet that "
-             "can be connected to in order to request the provided service.");
+             "Provider of credentials.  This is an identifier that could help "
```
mailletf: La description n'est pas très claire pour moi. C'est simplement le nom du service utiliser pour l'identifier à l'interne et n'est pas utilisé pour se connecter?

soa/credentials/credentials.cc
```diff
@@ -44,11 +44,30 @@ CredentialDescription()
...
 
-DEFINE_STRUCTURE_DESCRIPTION(CredentialContext);
+DEFINE_STRUCTURE_DESCRIPTION(StoredCredentials);
 
-CredentialContextDescription::
-CredentialContextDescription()
+StoredCredentialsDescription::
+StoredCredentialsDescription()
 {
+    addField("resourceType", &StoredCredentials::resourceType,
```
mailletf: Devrait être un enum?

soa/credentials/credentials.cc
```diff
@@ -44,11 +44,30 @@ CredentialDescription()
...
+             "Type of resource that this credential rule applies to.  Currently, "
+             "resources of type aws:s3 are supported when accessing AWS S3.");
+    addField("resource", &StoredCredentials::resource,
+             "Resource that this credential rule applies to.  This is a URI style path. "
+             "If the requested resource match this prefix, the credentials will be used. "
+             "The most specific match is used.  If this is empty, the credentials will "
+             "be used for all resources of type `resourceType`.");
+    addField("expiration", &StoredCredentials::expiration,
+             "Date on which credentials expire.  After this date they will no "
+             "longer match.");
```
mailletf: .. "and will be removed from the system" ? ou "and will be marked as expired"

soa/service/s3.cc
```diff
@@ -2586,36 +2574,28 @@ struct RegisterS3Handler {
...
-        virtual std::vector<Credential>
-        getSync(const std::string & resourceType,
-                const std::string & resource,
-                const CredentialContext & context,
-                Json::Value extraData) const
+        virtual std::vector<StoredCredentials>
+        getCredentialsOfType(const std::string & resourceType) const
         {
-            if (disableCloudCredentials)
+            if (disableCloudCredentials || resourceType != "aws:s3")
```
mailletf: Pas un peu trop hardcodé pour aws? Si on ajoute un autre type et qu'on oubli de modifier ceci, ça va faire un bug je pense?

testing/MLDB-1468-credentials-test.py
```diff
@@ -5,32 +5,118 @@
...
+                "runOnCreation": True
+            }
+        }
+
+        # this is expected to pick the most specific but invalid credentials
+        with self.assertRaises(mldb_wrapper.ResponseException) as re:
+            resp = mldb.put("/v1/procedures/import", csv_conf)
+
+
+    @unittest.skipIf(not isfile(join(expanduser('~'), '.cloud_credentials')),
```
mailletf: Do we have something in place to make sure jenkins will always have one and be alerted if this is not run on jenkins?

mailletf:
minorish comments; +1

server/credential_collection.cc
guyd: It would be very unlikely since adding credentials is nothing like running a procedure.  It is probably why this was an assert instead of an http exception.

soa/credentials/credentials.cc
guyd: Ce n'est pas vraiment utilisé.  J'avais créé MLDBFB-492 pour simplifier la configuration des credentials.

soa/credentials/credentials.cc
guyd: Ça pourrait être un enum mais comme le resourceType est persisté, par exemple dans le fichier .cloud_credentials, il faudrait persister un entier pour l'enum ce qui serait moins facile à comprendre que le string actuel.  Je suggère que l'on revisite la question lorsque l'on aura plus qu'un type de credential supporté.

soa/credentials/credentials.cc
guyd: Je ne suis pas sûr de ce que l'on veut ici.  MLDBFB-492 mentionne que certains champs ne sont pas encore supporté.  C'est le cas de celui-ci.  Je suggère encore que l'on reporte la décision lorsque l'on évaluera MLDBFB-492.

testing/MLDB-1468-credentials-test.py
guyd: Jenkins includes a .cloud_credentials file so all tests that require credentials to run including this one are run on every build.  Not sure how to implement this.  There are some tests that we want to skip (e.g those that are marked as manual because they are awaiting a fix) and some other that should not be skipped.  Any idea how we would do that?

PR #320, opened 2016-04-21
[MLDB-1547] added validation of modelFileUrl when functionName is spe…

guyd:
- added the validation to all procedure that can generate a function pointing to a model.  When the modelFileUrl is invalid or not specified, the creation of the procedure will fail
- simplified the config validation code to use variadic template and remove the use of template where it was providing little value
- moved the name of a procedure (e.g kmeans.train) to its config declaration so that we can reuse it easily


steadtler:
The high number of changes might make the upcoming refactoring difficult to integrate

guyd:
I will wait until the row and column naming changes are in.

nicolaskruchten:
The big changes have hit master, we can finish this one up now :)

steadtler:
+1

PR #321, opened 2016-04-22
PLAT-994: ensure that s3 code does not issue errors on stdout

wsourdeau:


soa/service/xml_helpers.h
```diff
@@ -90,5 +91,15 @@ T extractDef(const tinyxml2::XMLDocument & doc,
     return extractDef<T>(element, path, def);
 }
 
+inline std::string
+xmlDocumentAsString(const tinyxml2::XMLDocument & xmlDocument)
```
guyd: Specifying const here is a bit misleading since you do a const_cast after.

guyd:
+1 with a very minor comment

soa/service/xml_helpers.h
wsourdeau: That's because it is a good practice to constify refs by default and that XMLDocument::Print is not declared as const-safe even though it is.

PR #322, opened 2016-04-22
PLAT-993: detect and handle XML errors occurring without an HTTP code…

wsourdeau:
… 200

### DIFF
--- a/soa/service/s3.cc
+++ b/soa/service/s3.cc
@@ -401,13 +401,20 @@ performSync() const
 
         myRequest.get_info(CURLINFO_RESPONSE_CODE, responseCode);
 
-        /* Detect so-called "REST error", which may or may not have an HTTP
-           status >= 200
+        /* Detect so-called "REST error"
            (http://docs.aws.amazon.com/AmazonS3/latest/API/ErrorResponses.html)
+
+           Some S3 methods may return an XML error AND still have a 200 HTTP
+           status code:
+           http://docs.aws.amazon.com/AmazonS3/latest/API/mpUploadComplete.html
+           Explanation of the why:
+           https://github.com/aws/aws-sdk-go/issues/501.
         */
         pair<string, string> xmlError; /* {code, message} */
-        if (responseHeaders.find("Content-Type: application/xml")
-            != string::npos) {
+        if (!(responseCode == 200
+              && (params.verb == "GET" || params.verb == "HEAD"))
+            && (responseHeaders.find("Content-Type: application/xml")
+                != string::npos)) {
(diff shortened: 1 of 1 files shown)


soa/service/s3.cc
```diff
@@ -401,7 +401,27 @@ performSync() const
 
         myRequest.get_info(CURLINFO_RESPONSE_CODE, responseCode);
 
-        if (responseCode >= 300 && responseCode != 404) {
+        /* Detect so-called "REST error", which may or may not have an HTTP
+           status >= 200
+           (http://docs.aws.amazon.com/AmazonS3/latest/API/ErrorResponses.html)
```
guyd: Could this be a bug on S3 side because no where in the documentation they say that they could send a status code 200 and an error in the body.  You mentioned one case to me this morning.  If that case is documented somewhere, we should point to it.

soa/service/s3.cc
```diff
@@ -412,38 +432,28 @@ performSync() const
...
-                    .ToElement();
-                if (element) {
-                    message += ("S3 REST error: ["
-                                + extract<string>(element, "Code")
-                                + "] message ["
-                                + extract<string>(element, "Message")
-                                +"]\n");
-                }
-            }
+        if (!xmlError.first.empty()
```
guyd: This is unlikely but we would now interpret a valid download of an xml file that contains a field error as an error.

soa/service/s3.cc
```diff
@@ -412,38 +432,28 @@ performSync() const
...
+            || (responseCode >= 300 && responseCode != 404)) {
+            string message(makeErrorMsg());
             ::fprintf(stderr, "%s\n", message.c_str());
 
             /* retry on 50X range errors (recoverable) */
-            if (responseCode >= 500 && responseCode < 505) {
+            if ((responseCode >= 500 && responseCode < 505)
+                || xmlError.first == "InternalError") {
```
guyd: This seems to be another weirdness with the S3 API.  We could get InternalError with a status code != 500.  If we have a link to this bug/feature we should add it here.

guyd:
I am concern that the next time I read the code I will not understand it unless there are specific pointers to why we are doing these somersaults.

soa/service/s3.cc
wsourdeau: There is here: http://docs.aws.amazon.com/AmazonS3/latest/API/mpUploadComplete.html I don't know whether this is specific to CompleteMultipartUpload. I suggest we restrict this handling to non-GET requests.

guyd:
+1

PR #323, opened 2016-04-25
[MLDB-1493] Improved CellValue type casting exceptions

mailletf:


sql/cell_value.cc
```diff
@@ -337,7 +337,8 @@ toString() const
         return string(longString->repr, longString->repr + strLength);
     case ST_UTF8_SHORT_STRING:
     case ST_UTF8_LONG_STRING:
-        throw HttpReturnException(400, "cannot call toString on utf8 string");
+        throw HttpReturnException(400, "cannot call toString on CellValue of type '"+to_string(cellType())+
```
guyd: `toString` is referencing an internal name.  I feel this phrasing is better
```
Can't convert value 'bla' to ASCII string.
```

guyd:
+1

PR #325, opened 2016-04-25
Mldb 1427 sampled dataset default args

mailletf:


guyd:
I was looking at that code last night and I was wondering if we should not change the interface to make it less error prone to users by:
- changing the parameter `rows` in the function `sample` and in the dataset config by `limit` so that this parameter can coexist with `fraction` and so that a limit set higher to the number of rows of the dataset would succeed instead of returning an error
- aligning the default in the `sample` function with the one in the dataset config

builtin/sampled_dataset.cc
```diff
@@ -48,10 +48,10 @@ SampledDatasetConfigDescription()
     addField("rows", &SampledDatasetConfig::rows,
             "Number of rows to sample. This option cannot be set when fraction is "
             "set. This option cannot be set higher than the number of rows of "
-            "input unless withReplacement is set to true.");
+            "input unless withReplacement is set to true.", unsigned(0));
```
guyd: I suggest we align these defaults with the one in the `sample` function.

builtin/sampled_dataset.cc
```diff
@@ -299,29 +299,34 @@ SampledDataset(MldbServer * owner,
...
+            }
+            else if (columnName == ColumnName("withReplacement")) {
+                config.withReplacement = std::get<1>(elem).asBool();
+            }
+            else if (columnName == ColumnName("seed")) {
+                config.seed = std::get<1>(elem).toInt();
+            }
+            else {
+                auto expVal2 = std::get<1>(elem);
+                throw ML::Exception("unknown option: '"+columnName.toString()+"'");
```
guyd: I know this is not new code but I really feel we should not reference an internal name, here `option`, when returning an error.  `params` is probably more appropriate since it is mentioned in the `sampled` dataset documentation.  Also, we should mention that this is returned by the `sample` function.

sql/builtin_dataset_functions.cc
```diff
@@ -138,15 +138,17 @@ BoundTableExpression sample(const SqlBindingScope & context,
                                   "options", options,
                                   "alias", alias);
 
-    if(!options.isRow())
-        throw HttpReturnException(400, "options should be a row; we got "
-                                  + jsonEncodeStr(options),
-                                  "options", options,
-                                  "alias", alias);
+    if(!options.empty() && !options.isRow()) {
+        throw HttpReturnException(400, "options should be a row containing "
```
guyd: Here also, we mention an internal name `options`.

mailletf:
@guyd: To rephrase your suggestion, `rows` -> `limit`, and then we return the min of `limit` and `fraction` ?

guyd:
Yes

mailletf:
Talked about it with some internal users and your suggestion does allow things that the current way doesn't, while taking nothing away. I'll make the change.

builtin/sampled_dataset.cc
```diff
@@ -46,12 +62,12 @@ SampledDatasetConfigDescription()
...
-            "set. This option cannot be set higher than the number of rows of "
-            "input unless withReplacement is set to true.");
+            "Number of rows to sample from `dataset`. Cannot be used with "
+            "`fraction`. Cannot be higher than the number of rows in `dataset` "
+            "unless `withReplacement` = 1. Default = 1 if `fraction` is 0.");
     addField("fraction", &SampledDatasetConfig::fraction,
-            "Fraction of rows in dataset to sample. It cannot be set when "
-            "rows is set.");
+            "Fraction of rows to sample from `dataset`. Cannot be used when "
+            "`rows` != 0. Value should be between 0 and 1.");
```
nicolaskruchten: is the default value for `fraction` = 0 ?

nicolaskruchten:
+1

PR #327, opened 2016-04-26
Mldb 1525 rowname generator explain

steadtler:
Add a new field to GenerateRowsWhereFunction to track its complexity so that we can reason on it.


sql/sql_expression.h
```diff
@@ -1301,14 +1301,24 @@ PREDECLARE_VALUE_DESCRIPTION(TupleExpression);
 */
 struct GenerateRowsWhereFunction {
 
+    enum Complexity {        
+       
+        CONSTANT = 0,
+        BETTER_THAN_TABLESCAN,        
+        UNFILTERED_TABLESCAN,
```
guyd: What would be an example of `UNFILTERED_TABLESCAN`? I could not find any example in the test case.

guyd:
+1

sql/sql_expression.h
steadtler: When you provide no WHERE clause, or some WHERE clause that evaluate to always true. Basically anytime we want all rows.

guyd: So how is it different than TABLESCAN then?

PR #328, opened 2016-04-26
MLDBFB-474 add an option to the transform procedure to specify determi…

steadtler:
…nistic ordering or not. Deterministic is slower.

### DIFF
--- a/plugins/sql_functions.cc
+++ b/plugins/sql_functions.cc
@@ -503,7 +503,7 @@ regSqlExpressionFunction(builtinPackage(),
 
 TransformDatasetConfig::
 TransformDatasetConfig()
-    : skipEmptyRows(false)
+    : skipEmptyRows(false), deterministicOrdering(false)
 {
     outputDataset.withType("sparse.mutable");
 }
@@ -524,6 +524,9 @@ TransformDatasetConfigDescription()
              "which will be created by the procedure.", PolyConfigT<Dataset>().withType("sparse.mutable"));
     addField("skipEmptyRows", &TransformDatasetConfig::skipEmptyRows,
              "Skip rows from the input dataset where no values are selected",
+             false);
+    addField("deterministicOrdering", &TransformDatasetConfig::deterministicOrdering,
+             "Enforce a deterministic ordering of the rows in the transformed dataset",
              false);
     addParent<ProcedureConfig>();
 }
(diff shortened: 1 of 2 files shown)


FinchPowers:
+1, mais il faudrait que @nicolaskruchten donne son accord à ce paramètre.

nicolaskruchten:
Hmmm I'm not sure I understand the need for such a parameter vs just using `ORDER BY` in the query... what's the downside of just hardcoding this to false all the time? and letting the user use order-by if they care?

plugins/sql_functions.cc
```diff
@@ -611,7 +614,7 @@ run(const ProcedureRunConfig & run,
                          *runProcConf.inputData.stm->where,
                          runProcConf.inputData.stm->orderBy,
                          { runProcConf.inputData.stm->rowName })
-            .execute({recordRowInOutputDataset, true /*processInParallel*/},
+            .execute({recordRowInOutputDataset, !runProcConf.deterministicOrdering /*processInParallel*/},
```
guyd: It was already executed in parallel.  The flag will only allow it to go slower.  That does not seem to fit the JIRA description!

steadtler:
Ok! Lets explain this:

-Allowing output in parallel is the fastest, but the rows in the output dataset will be in a *different* order everytime you run it, since it depends on threading.
-Disallowing output in parallel will give the output in an *arbitrary* order, but it will always be the same order if you run it several times -> deterministic order.
-Having the user put an order by will give the result in a *specific* order, and it is the slowest.

1st option takes 20s, second option 1m15s (because some cores are starved), last option over an hour (according to Mich, I didnt test it) on Mich's dataset.

nicolaskruchten:
Given that the target here is a dataset and not a query result, I'm ok with
the output of a transform to be nondeterministic always if it's fastest,
instead of making the user choose.

On Wednesday, 27 April 2016, Mathieu Marquis Bolduc <
notifications@github.com> wrote:

> Ok! Lets explain this:
>
> -Allowing output in parallel is the fastest, but the rows in the output
> dataset will be in a *different* order everytime you run it, since it
> depends on threading.
> -Disallowing output in parallel will give the output in an *arbitrary*
> order, but it will always be the same order if you run it several times ->
> deterministic order.
> -Having the user put an order by will give the result in a *specific*
> order, and it is the slowest.
>
> 1st option takes 20s, second option 1m15s (because some cores are
> starved), last option over an hour (according to Mich, I didnt test it) on
> Mich's dataset.
>
> —
> You are receiving this because you were mentioned.
> <https://github.com/mldbai/mldb/pull/328#issuecomment-215066811>
>


-- 
http://datacratic.com/

PR #329, opened 2016-04-26
MLDB-1478 add error message when trying to read a column in a no-data…

steadtler:
…set context


server/dataset_context.cc
```diff
@@ -83,7 +83,13 @@ getMldbServer() const
     return mldb;
 }
 
-
+VariableGetter
+SqlExpressionMldbContext::
+doGetVariable(const Utf8String & tableName,
+              const Utf8String & variableName)
+{
+    throw HttpReturnException(400, "Cannot read column \"" + variableName + "\" with no dataset.");
```
guyd: We have use single quotation mark elsewhere to highlight the name.

guyd:
+1

PR #330, opened 2016-04-27
MLDBFB-475 add a method to skip assert checking for null chars in Coo…

steadtler:
…rd string

Needed by a change in pro version.


guyd:
Not sure I understand where we are going with this change.  I think Coord should either support strings with embedded null characters in such case the assert should go and its interface should make it clear or it does not support embedded nulls.

steadtler:
Its a temporary measure to unblock Mich.

guyd:
+1

PR #331, opened 2016-04-28
Mldb 1486 embedding rownames rebased

jeremybarnes:
This modifies the handling of rownames so that they are structured.  It includes a huge set of bubbled-down fixes to underlying functionality.


nicolaskruchten:
Nice job on pushing this through Jeremy! I'll do a mock release on this branch to see what breaks in the demos etc, and I'll do a test by test review to understand the modified semantics of the system. I'll then add some documentation commits to this PR which should be reviewed by @jeremybarnes to make sure I caught the nuances.

arch/simd_vector.cc
```diff
@@ -541,6 +541,44 @@ double vec_dotprod_sse2(const double * x, const double * y, size_t n)
...
+        }
+
+        for (; i + 4 <= n;  i += 4) {
+            v4sf yyyy0 = _mm_loadu_ps(y + i + 0);
+            v4sf xxxx0 = _mm_loadu_ps(x + i + 0);
+            xxxx0 -= yyyy0;
+            __builtin_ia32_storeups(r + i + 0, xxxx0);
+        }
+        
+        for (; i < n;  ++i) r[i] = x[i] - y[i];
```
guyd: Shouldn't it be one or the other but not both for loops?

arch/simd_vector_avx.cc
```diff
@@ -112,6 +133,133 @@ float vec_dotprod(const float * x, const float * y, size_t n)
...
+        }
+
+        for (; i + 8 <= n;  i += 8) {
+            v8sf yy0 = _mm256_loadu_ps(y + i);
+            v8sf xx0 = _mm256_loadu_ps(x + i);
+            _mm256_storeu_ps(r + i, xx0 - yy0);
+        }
+    }
+
+    for (; i < n;  ++i) r[i] = x[i] - y[i];
```
guyd: Here also.

arch/simd_vector_avx.cc
```diff
@@ -112,6 +133,133 @@ float vec_dotprod(const float * x, const float * y, size_t n)
...
+        }
+
+        // This performs exactly the same operation as as the sse2 version,
+        // and so gives exactly the same result (bit for bit).  That's
+        // the reason for the convoluted code here.
+        result += horiz_sum_8_avx(rr0, rr1);
+    }
+#endif
+
+    for (; i < n;  ++i) result += (x[i] - y[i]) * (x[i] - y[i]);
```
guyd: and here

container_files/public_html/doc/builtin/datasets/EmbeddingDataset.md
```diff
@@ -55,8 +44,9 @@ giving the highest distance that a point can be in order to be included.
 ## See Also
 
 * [Vantage Point Tree] is the data structure used to allow quick lookups
+* the ![](%%doclink nearest.neighbors function) is used to find nearest neighbours in an embedding dataset.
```
guyd: We use neighbour or neighbor but not both.

core/dataset.cc
```diff
@@ -1146,7 +1100,8 @@ generateRowsWhere(const SqlBindingScope & scope,
     if (inExpression) 
     {
         auto fexpr = getFunction(*(inExpression->expr));
-        if (fexpr && removeTableName(alias, fexpr->functionName) == "rowName" ) {
+        // TODO BEFORE MERGING: functionName should be a compound
```
guyd: Was this completed?

arch/simd_vector.cc
jeremybarnes: No... this is a typical geometric unrolling strategy.  The first loop is unrolled to 16 values at once, but must have at least 16 to run.  Rather than do the remainder one by one, we then do it 4 by 4, and the remaining 0 to 3 individually.

arch/simd_vector_avx.cc
jeremybarnes: Same

arch/simd_vector_avx.cc
jeremybarnes: Same

plugins/embedding.cc
```diff
@@ -1211,13 +1189,166 @@ getKnownColumnInfos(const std::vector<ColumnName> & columnNames) const
...
+/*****************************************************************************/
+
+DEFINE_STRUCTURE_DESCRIPTION(NearestNeighborsFunctionConfig);
+
+NearestNeighborsFunctionConfigDescription::
+NearestNeighborsFunctionConfigDescription()
+{
+    addField("defaultNumNeighbors",
+             &NearestNeighborsFunctionConfig::defaultNumNeighbors,
+             "Default number of neighbors to return. This can be overritten "
```
guyd: overritten -> overwritten

plugins/embedding.cc
```diff
@@ -1211,13 +1189,166 @@ getKnownColumnInfos(const std::vector<ColumnName> & columnNames) const
...
+             "when calling the function.", unsigned(10));
+    addField("defaultMaxDistance",
+             &NearestNeighborsFunctionConfig::defaultMaxDistance,
+             "Default maximum distance from the original row returned "
```
guyd: It took me three time to understand the sentence.  This feels better to me:

> Set the maximum distance neighbors can be from the original row returned.

core/dataset.cc
jeremybarnes: Fixed...

container_files/public_html/doc/builtin/datasets/EmbeddingDataset.md
jeremybarnes: Fixed... neighbor FTW!

plugins/embedding.cc
jeremybarnes: Fixed

plugins/kmeans.cc
```diff
@@ -197,7 +198,15 @@ run(const ProcedureRunConfig & run,
     if (!runProcConf.modelFileUrl.empty()) {
         try {
             Datacratic::makeUriDirectory(runProcConf.modelFileUrl.toString());
-            kmeans.save(runProcConf.modelFileUrl.toString());
+            Json::Value md;
```
guyd: Why is this inline here? We also lost backward compatibility.  I am not sure if any use kmeans model in production but they should be made aware of this.

plugins/embedding.cc
jeremybarnes: Fixed

plugins/kmeans.cc
```diff
@@ -294,72 +303,79 @@ KmeansFunctionConfigDescription()
...
+    {
+        filter_istream stream(modelFileUrl.toString());
+        std::string firstLine;
+        std::getline(stream, firstLine);
+        Json::Value md = Json::parse(firstLine);
+        if (md["algorithm"] != "MLDB k-Means model") {
+            throw HttpReturnException(400, "Model file is not a k-means model");
+        }
+        if (md["version"].asInt() != 1) {
+            throw HttpReturnException(400, "k-Means model version is wrong");
```
guyd: I feel this should be a 500.

plugins/lang/python/python_loader.cc
```diff
@@ -750,14 +750,14 @@ class MldbUnitTest(unittest.TestCase):
 
 namespace {
 
-// The set function in FunctionOutput and FunctionContext are both overloaded due to
+// The set function in ExpressionValue and ExpressionValue are both overloaded due to
```
guyd: :-)

plugins/nlp.h
```diff
@@ -30,24 +29,31 @@ struct ApplyStopWordsFunctionConfig {
 
 DECLARE_STRUCTURE_DESCRIPTION(ApplyStopWordsFunctionConfig);
 
+struct Words {
```
guyd: This feels a bit strange.  Could a typedef for Words and Documenr be used instead?  If not, may be we can renames
```
struct Words {
   ExpressionValue counts;
}
```
and
```
struct Document {
    ExpressionValue text;
}
```

plugins/pooling_function.h
```diff
@@ -38,28 +38,36 @@ DECLARE_STRUCTURE_DESCRIPTION(PoolingFunctionConfig);
 /* POOLING FUNCTION                                                          */
 /*****************************************************************************/
 
-struct PoolingFunction: public Function {
+struct PoolingInput {
```
guyd: Or may be the renaming mentioned above could be like this one.

plugins/sqlite_dataset.cc
```diff
@@ -541,7 +533,10 @@ struct SqliteSparseDataset::Itl
 
         sqlite3pp::command command(*db, "INSERT OR IGNORE INTO vals VALUES (?, ?, ?, ?)");
 
+        std::vector<ColumnName> tmp;
```
guyd: Not use.

plugins/stats_table_procedure.cc
```diff
@@ -110,18 +122,18 @@ save(const std::string & filename) const
 void StatsTable::
 serialize(ML::DB::Store_Writer & store) const
 {
-    int version = 1;
+    int version = 2;
```
guyd: Do we need to bump the version because colName has changed?

plugins/stats_table_procedure.cc
```diff
@@ -110,18 +122,18 @@ save(const std::string & filename) const
...
 void StatsTable::
 reconstitute(ML::DB::Store_Reader & store)
 {
     int version;
-    int REQUIRED_V = 1;
+    int REQUIRED_V = 2;
     store >> version;
     if(version!=REQUIRED_V) {
-        throw ML::Exception(ML::format(
+        throw HttpReturnException(400, ML::format(
```
guyd: We don't care about backward compatibility here also?

plugins/stats_table_procedure.cc
guyd: This format is different than other persisted model.  I suggest we add the name of the algorithm that created the file as well.

plugins/svd.cc
```diff
@@ -1005,37 +1003,62 @@ SvdEmbedConfigDescription()
...
+             "the SVD of each of the keys of the row.");
+}
+
+DEFINE_STRUCTURE_DESCRIPTION(SvdOutput);
+
+SvdOutputDescription::
+SvdOutputDescription()
+{
+    addField("embedding", &SvdOutput::embedding,
+             "Embedding of the row into the vector spaced defined by the "
```
guyd: spaced -> space

plugins/svm.cc
```diff
@@ -376,6 +402,16 @@ SVMFunction(MldbServer * owner,
...
     try {
         filter_istream in(functionConfig.modelFileUrl.toString());
+        std::string firstLine;
+        std::getline(in, firstLine);
+        Json::Value md = Json::parse(firstLine);
+        if (md["algorithm"] != "MLDB SVM model") {
+            throw HttpReturnException(400, "Model file is not an SVM model");
+        }
+        if (md["version"].asInt() != 1) {
+            throw HttpReturnException(400, "SVM model version is wrong");
```
guyd: I would use 500 here also.

plugins/tsne.cc
```diff
@@ -379,34 +379,45 @@ TsneEmbedConfigDescription()
 /* TSNE EMBED ROW                                                            */
 /*****************************************************************************/
 
+DEFINE_STRUCTURE_DESCRIPTION(TsneInput);
+
+TsneInputDescription::TsneInputDescription()
+{
+    addField("embedding", &TsneInput::embedding,
+             "Undocumented.");
```
guyd: Should this be completed?

server/analytics.cc
```diff
@@ -436,20 +417,25 @@ queryFromStatement(SelectStatement & stm,
...
     if (!rowNameEV.isAtom())
         throw HttpReturnException(400, "NAMED expression must evaluate to a single value");
 
-    Utf8String rowName = rowNameEV.toUtf8String();
+    Utf8String rowNameStr = rowNameEV.toUtf8String();
+
+    if (rowNameStr.empty())
+        throw HttpReturnException(400, "Can't create a row with an empty name");
+
+    auto rowName = RowName::parse(rowNameEV.toUtf8String());
```
guyd: You can parse rowNameStr instead.

guyd:
Reviewed with mostly minor comments.  The function implementation and the manipulation of names feel more natural now.  +1

At one point, we should agree on a line length and formatting to make reviewing easier.  It was hard to pin-point the important changes among the renaming, the reformatting and the moving.

plugins/kmeans.cc
jeremybarnes: The problem is that people using saved k-means models were probably getting the wrong results from them, silently, because no record was made of which columns were provided.  Agreed that people who have saved models need to be made aware of it, but I don't think there are any in the wild ( @nicolaskruchten please verify).

plugins/kmeans.cc
jeremybarnes: It's not an internal server error; it's an error with a resource that the user supplied.  IMO this makes it a 4xx.

plugins/nlp.h
jeremybarnes: They are like that so that the names match up with what they were previously called.  In other words, this is a refactoring of the same explicit structure that was present in the previous version, just put all in one place.   I agree we can improve things a lot, but for the moment we want to keep compatibility with previous usage of the functions and so we should keep them the same as what they were.  At least this way we have a place to document the parameters.

plugins/stats_table_procedure.cc
jeremybarnes: @nicolaskruchten can you weigh in on this please?  If we do care about backwards compatibility, do you have a stats table binary file generated with a previous version that we can use in a unit test?

plugins/kmeans.cc
guyd: I see.  I suggest we delete the old serialization code then to avoid further problem.

plugins/tsne.cc
jeremybarnes: That is a separate DOD; especially since the function is disabled there was no point in writing the doc.

plugins/kmeans.cc
guyd: This is very minor but I will at a minimum clarify why I think this is a 500.
My interpretation is that the first error is a 400 (a user provided the wrong model) but the second one is a 500 since, as a user, my expectation is that the latest version of a software should support all previous versions of its artefacts.  Here a user provided the right model but MLDB does not know how to read it.

plugins/stats_table_procedure.cc
guyd: Don't we use stat tables extensively with CommunitySift?

plugins/kmeans.cc
jeremybarnes: I don't follow... we should delete which part of the code?

plugins/kmeans.cc
jeremybarnes: Or, the user provided a path to something that's not a k-means model...

plugins/kmeans.cc
guyd: https://github.com/mldbai/mldb/blob/master/ml/kmeans.cc#L236

nicolaskruchten: we can sacrifice back-compatibility of model files for k-means today.

plugins/stats_table_procedure.cc
nicolaskruchten: we can sacrifice back-compatibility of model files for stats tables today. Anyone using them will need to migrate a bunch of code anyway to track this change, and we will help.

nicolaskruchten: but thanks for raising the flag :)

jeremybarnes:
(Note: test failure is spurious)

nicolaskruchten:
+1 to merge!

PR #332, opened 2016-04-28
[MLDB-1596] Added output file writability check *before* training proc…

mailletf:
…edures are run. Automatically created functions now override previously existing functions.

### DIFF
--- a/vfs/fs_utils.cc
+++ b/vfs/fs_utils.cc
@@ -560,6 +560,9 @@ checkWritability(const std::string & url, const std::string & parameterName)
                 "in parameter '%s'. Value: '%s'. Exception: %s",
                 parameterName, url, ex.what()));
     }
+
+    // remove empty file
+    tryEraseUriObject(url);
 }
 
 
--- a/vfs/fs_utils.cc
+++ b/vfs/fs_utils.cc
@@ -537,6 +537,33 @@ dirName(const std::string & filename)
 }
 
 /****************************************************************************/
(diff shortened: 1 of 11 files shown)


vfs/fs_utils.cc
```diff
@@ -537,6 +537,33 @@ dirName(const std::string & filename)
...
+/* UX FUNCTIONS                                                             */
+/****************************************************************************/
+
+void
+checkWritability(const std::string & url, const std::string & parameterName)
+{
+    // try to create output folder and write open a writer to make sure 
+    // we have permissions before
+    try {
+        Datacratic::makeUriDirectory(url);
```
guyd: This will create the directory and the file.  I believe we don't want that.  We can use stat() instead (e.g. via boost::file_system) to determine if the folder has the proper access.  Also, what would be the effect of that call if the path point to S3?

mailletf: The directory and the file will be created anyways when the procedure runs. Doing it with the filter output streams makes it work for s3 de manière transparente, which is why I did it this way.

guyd: > the file will be created anyways when the procedure runs
Not if it fails before.   Do we have support to create a tmp file (that will automatically delete itself when the program exit)?  We probably don't have that for S3.  :-(

mailletf: Check latest commit. If nothing throws (ie empty file was created), we remove the empty file before returning.

guyd:
+1

PR #333, opened 2016-04-28
MLDB-1073 fix bug in generic GetColumnBuckets

steadtler:


core/dataset.cc
```diff
@@ -385,7 +385,7 @@ getColumnBuckets(const ColumnName & column,
 
     for (auto& v : vals) {
         if (values.insert({v,0}).second)
-            valueList.push_back(std::move(v));
+            valueList.push_back(v);
```
steadtler: v is used below so we can't move it.

guyd:
+1

PR #334, opened 2016-04-28
Explain std::bad_alloc message (MLDB-1355)

jeremybarnes:
More friendly error messages when out of memory:

```javascript
--------------------------[Exception thrown]---------------------------
time:   2016-04-28T21:11:04Z
type:   std::bad_alloc
pid:    66164; tid: 66165
what:   std::bad_alloc

2016-04-28 17:11:04.096 javascript {
   "details" : {
      "context" : {
         "error" : "Out of memory.  A memory allocation failed when performing the operation.  Consider retrying with a smaller amount of data or running on a machine with more memory.  (std::bad_alloc)"
      },
      "functionArgs" : [],
      "functionName" : "_fail_memory_allocation"
   },
   "error" : "Executing builtin function _fail_memory_allocation: Out of memory (std::bad_alloc)",
   "httpCode" : 400
}
```

### DIFF
--- a/http/http_exception.cc
+++ b/http/http_exception.cc
@@ -24,6 +24,14 @@ void rethrowHttpException(int httpCode, const Utf8String & message, Any details)
         details2["context"]["details"] = jsonEncode(http.details);
         details2["context"]["error"] = http.message;
         throw HttpReturnException(httpCode == -1 ? http.httpCode : httpCode, message, details2);
+    } catch (const std::bad_alloc & exc) {
+        Json::Value details2 = jsonEncode(details);
+        details2["context"]["error"]
+            = "Out of memory.  A memory allocation failed when performing "
+            "the operation.  Consider retrying with a smaller amount of data "
+            "or running on a machine with more memory.  "
+            "(std::bad_alloc)";
+        throw HttpReturnException(httpCode == -1 ? 400 : httpCode, message, details2);
     } catch (const std::exception & exc) {
         Json::Value details2 = jsonEncode(details);
         details2["context"]["error"] = exc.what();
--- a/rest/rest_request_router.cc
+++ b/rest/rest_request_router.cc
@@ -1007,6 +1007,8 @@ Json::Value extractException(const std::exception & exc, int defaultCode)
 {
     const HttpReturnException * http
         = dynamic_cast<const HttpReturnException *>(&exc);
+    const std::bad_alloc * balloc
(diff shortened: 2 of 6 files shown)


http/http_exception.cc
```diff
@@ -24,6 +24,14 @@ void rethrowHttpException(int httpCode, const Utf8String & message, Any details)
         details2["context"]["details"] = jsonEncode(http.details);
         details2["context"]["error"] = http.message;
         throw HttpReturnException(httpCode == -1 ? http.httpCode : httpCode, message, details2);
+    } catch (const std::bad_alloc & exc) {
```
wsourdeau: I seem to recall a discussion about avoiding allocations when handing std::bad_alloc. I doubt such a small amount of memory would be a problem but I think it's better to think twice about this.

jeremybarnes: Here we're rethrowing an exception once all unwinding has been done, in order to determine the type of the current exception.  It would be a very large amount of work to ensure that we could do good error handling and reporting without allocating any memory.  This solution is a 99% solution and the other 1% will be so much work I don't think it's worth doing.  (And this patch doesn't make things any worse as compared to without it).

http/http_exception.cc
```diff
@@ -24,6 +24,14 @@ void rethrowHttpException(int httpCode, const Utf8String & message, Any details)
...
+        Json::Value details2 = jsonEncode(details);
+        details2["context"]["error"]
+            = "Out of memory.  A memory allocation failed when performing "
+            "the operation.  Consider retrying with a smaller amount of data "
+            "or running on a machine with more memory.  "
```
guyd: double space

rest/rest_request_router.cc
```diff
@@ -1015,7 +1017,13 @@ Json::Value extractException(const std::exception & exc, int defaultCode)
         val["httpCode"] = http->httpCode;
         if (!http->details.empty())
             val["details"] = jsonEncode(http->details);
-    } else {
+    } else if (balloc) {
+        val["error"] = "Out of memory.  A memory allocation failed when performing "
+            "the operation.  Consider retrying with a smaller amount of data "
+            "or running on a machine with more memory.  "
```
guyd: double space

guyd:
+1 (you can of course ignore the double spaces in the message :-)

http/http_exception.cc
jeremybarnes: I prefer double spaces.  I don't understand people who remove them :)

PR #336, opened 2016-04-29
Mldb 1570 proc progress

guyd:
This is the _minimal_ implementation for the procedure progress.  This is so you can provide early feedback before diving too deep into the code.  What it includes:
- tracking of the state of the run in the background task
- simple encapsulation of progress in Progress and Step objects
- progress reporting in the bucketize procedure
- minimal progress reporting in the dataset iteration

What is not included:
- change of the onProgress callback signature to avoid serializing and deserializing to json
- progress in other procedures
- modification of the collection entries to avoid duplication of the progress data in the background task (while the procedure run) and the procedure itself (once it is completed).


server/bound_queries.cc
```diff
@@ -162,6 +162,12 @@ struct UnorderedExecutor: public BoundSelectQuery::Executor {
                         if (!doRow(it))
                             return false;
                     }
+
+                    if (onProgress) {
```
steadtler: There is no guarantee that the buckets will be done in order...

guyd: Very good point.  That would make a funny progress!  :-)

plugins/bucketize_procedure.cc
```diff
@@ -30,6 +30,80 @@ using namespace std;
...
+            nextStep->started = Date::now();
+        return nextStep;
+    }
+
+    std::string name;
+    Date started;
+    Date ended;
+    float percent;
+    std::weak_ptr<Step> _nextStep;
+};
```
FinchPowers: Déplacer dans un autre fichier.

nicolaskruchten:
please don't merge this until after this week's release :)

FinchPowers:
@nicolaskruchten Est-ce que j'ai +1 maintenant que le release est fait?

mailletf:
@FinchPowers  +1

PR #337, opened 2016-04-29
MLDB-1599 add clamp function

steadtler:


sql/builtin_functions.cc
```diff
@@ -2432,6 +2433,50 @@ BoundFunction flatten(const std::vector<BoundSqlExpression> & args)
...
+    return {[=] (const std::vector<ExpressionValue> & args,
+                 const SqlRowScope & scope) -> ExpressionValue
+            {
+                double lower = args[1].toDouble();
+                double upper = args[2].toDouble();
+
+                Date limitsTs = std::max(args[1].getEffectiveTimestamp(), args[2].getEffectiveTimestamp());
+
+                if (args[0].isAtom()) {
+                    double clamped = boost::algorithm::clamp(args[0].toDouble(), lower, upper);
```
jeremybarnes: There is a more pervasive problem... in that 64 bit integer values will end up being molested by being converted to doubles.  Instead, I would define clamp using operator < on the ExpressionValue objects and ignore doubles altogether.

sql/builtin_functions.cc
```diff
@@ -2432,6 +2433,50 @@ BoundFunction flatten(const std::vector<BoundSqlExpression> & args)
 
 static RegisterBuiltin registerFlatten(flatten, "flatten");
 
+BoundFunction clamp(const std::vector<BoundSqlExpression> & args)
+{
+    // Return the result indexed on a single dimension
```
jeremybarnes: Comment doesn't make sense

sql/builtin_functions.cc
jeremybarnes: (or operator < on CellValues)

jeremybarnes:
Missing documentation and the test case seems to fail.  Once issues are addressed I'm +1.

steadtler:
DoD got expanded.

PR #338, opened 2016-04-29
Mldb 1599 clamp

steadtler:
I will hold it to refactor it slightly after the row/column names refactoring is merged.


nicolaskruchten:
+1

container_files/public_html/doc/builtin/sql/ValueExpression.md
```diff
@@ -387,6 +387,7 @@ expression|result
 - `replace_inf(x, y)`: replace all `Inf`s and `-Inf`s in `x` by `y`.
 - `binomial_lb_80(trials, successes)` returns the 80% lower bound using the Wilson score.
 - `binomial_ub_80(trials, successes)` returns the 80% upper bound using the Wilson score.
+- `clamp(x,lower,upper)` will clamp the value 'x' between the lower and upper bounds.
```
guyd: We can be more explicit about what the function does.  I think you gave examples in the previous version of that PR.

testing/MLDB-162-min-max-function.py
```diff
@@ -118,4 +118,111 @@
...
+
+result = mldb.query('SELECT clamp(2, null, null) as v')
+expected = [["_rowName","v"],["result",2]]
+assert result == expected
+
+result = mldb.query('SELECT clamp(null, null, null) as v')
+expected = [["_rowName","v"],["result",None]]
+assert result == expected
+
+result = mldb.query('SELECT clamp(nan, null, null) as v')
```
guyd: I would test valid cases where lower and higher are strings.  I would test failure cases where lower and higher are rows.

guyd:
I feel we should add more tests before merging.

testing/MLDB-162-min-max-function.py
nicolaskruchten: `clamp` isn't meant to accept strings...

nicolaskruchten: but yeah, would be good to have a couple of tests with input of invalid types

guyd:
If string values are not allowed for lower and upper then we should have validation.  I believe the current implementation will happily accept any atom and do a comparison of string against number.

nicolaskruchten:
That's kind of fun actually... `select clamp('a','b','c')` -> `'b' `

PR #340, opened 2016-04-29
Allow space between COLUMN EXPR and ( to be optional (MLDB-1590)

jeremybarnes:
Fixes issues with COLUMN EXPR(...) that users have encountered.


steadtler:
+1

PR #343, opened 2016-04-29
Improved error message for training regression on inf/NaN (MLDB-1605)

jeremybarnes:
Also includes some more functions we can use in the WHERE to fix it.

### DIFF
--- a/container_files/public_html/doc/builtin/sql/ValueExpression.md
+++ b/container_files/public_html/doc/builtin/sql/ValueExpression.md
@@ -365,7 +365,7 @@ With `{arrays: 'encode'}` the output will be:
 - `mod(x, y)`: returns x modulo y.  The value of x and y must be an integer.
 - `abs(x)`: returns the absolute value of x.
 - `sqrt(x)`: returns the square root of x.  The value of x must be greater or equal to 0.
-- `isnan(x)`: return true if x is not-a-number in the floating point representation.
+- `isnan(x)`: return true if x is 'NaN' in the floating point representation.
 - `isinf(x)`: return true if x is infinity in the floating point representation.
 - `isfinite(x)`: return true if x is neither infinite nor not-a-number.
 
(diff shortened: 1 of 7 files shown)


container_files/public_html/doc/builtin/sql/ValueExpression.md
```diff
@@ -365,6 +365,10 @@ With `{arrays: 'encode'}` the output will be:
 - `mod(x, y)`: returns x modulo y.  The value of x and y must be an integer.
 - `abs(x)`: returns the absolute value of x.
 - `sqrt(x)`: returns the square root of x.  The value of x must be greater or equal to 0.
+- `isnan(x)`: return true if x is not-a-number in the floating point representation.
```
nicolaskruchten: would be good to have underscores to match replace_*

jeremybarnes:
I added replace_null to the PR also.

nicolaskruchten:
Do we also need is_nad() for detecting not-a-date? Or can I pass a
timestamp into is_nan() and it will coerce and return the right thing?

On Saturday, 30 April 2016, Jeremy Barnes <notifications@github.com> wrote:

> I added replace_null to the PR also.
>
> —
> <https://github.com/mldbai/mldb/pull/343#issuecomment-215960052>
>


-- 
http://datacratic.com/

jeremybarnes:
It doesn't handle timestamps.  We could add an equivalent function, but maybe that should be another DOD.

nicolaskruchten:
+1

plugins/classifier.cc
```diff
@@ -528,6 +528,15 @@ run(const ProcedureRunConfig & run,
         if (!isfinite(weight))
             throw HttpReturnException(400, "classifier example weights must be finite");
 
+        if (runProcConf.mode == CM_REGRESSION
+            && !isfinite(label)) {
+            throw HttpReturnException
+                (400,
+                 "Regression labels must not be infinite or NaN.  Should you "
+                 "add a condition like `WHERE isfinite(label)` to your data, "
+                 "or preprocess your labels with `replace_if_not_finite(label, 0)`?");
```
guyd: The name of the newly added function is `replace_not_finite`.

nicolaskruchten: i'll hot-patch the error message.

nicolaskruchten: done. nice catch @guyd

PR #344, opened 2016-05-02
MLDB-1592 make subdataset getcolumnvalues handle absence of filter

steadtler:


steadtler:
Issue is actually MLDB-1591

nicolaskruchten:
+1

PR #345, opened 2016-05-02
Coord -> PathElement, Coords -> Path renaming (MLDB-1610)

jeremybarnes:
Pure renaming: Coord -> PathElement, Coords -> Path


guyd:
+1

PR #349, opened 2016-05-03
[MLDBFB-506] add a test case

simlmx:
This was suppose to outline a bug but merging with master just fixed it. @nicolaskruchten you can decide if it's worth merge it anyway.

### DIFF
--- a/testing/MLDBFB-506-stats-tbl-sql-expr.py
+++ b/testing/MLDBFB-506-stats-tbl-sql-expr.py
@@ -1,0 +1,104 @@
+mldb = mldb_wrapper.wrap(mldb) # noqa
+
+
+class BowSqlExprTest(MldbUnitTest):
+
+    def check_true(self, q, truth):
+        data = mldb.get('/v1/query', q=q, format='aos', rowNames=0).json()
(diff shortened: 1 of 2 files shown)


testing/MLDBFB-506-stats-tbl-sql-expr.py
```diff
@@ -0,0 +1,104 @@
+mldb = mldb_wrapper.wrap(mldb) # noqa
+
+
+class BowSqlExprTest(MldbUnitTest):
+
+    def check_true(self, q, truth):
```
nicolaskruchten: we have a dedicated assert equal query result function... check https://github.com/mldbai/mldb/blob/master/testing/MLDBFB-336-sample_test.py

nicolaskruchten:
+1 when Jenkins passes

nicolaskruchten:
thanks for the test :)

testing/MLDBFB-506-stats-tbl-sql-expr.py
simlmx: Cool was wondering about that.

PR #355, opened 2016-05-07
MLDB-1631 join transpose where rowname

jeremybarnes:
Fixes two things:
- Segfault caused by binary operations not filling in the expression that they came from
- Refactor of regex builtins to allow a non-constant regex string

### DIFF
--- a/sql/sql_expression_operations.cc
+++ b/sql/sql_expression_operations.cc
@@ -814,6 +814,7 @@ struct BinaryOpHelper {
                                 std::placeholders::_1,
                                 std::placeholders::_2,
                                 GET_LATEST);
+        result.expr = expr->shared_from_this();
 
         return result;
     }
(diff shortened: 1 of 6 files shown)


sql/sql_expression_operations.cc
```diff
@@ -814,6 +814,7 @@ struct BinaryOpHelper {
                                 std::placeholders::_1,
                                 std::placeholders::_2,
                                 GET_LATEST);
+        result.expr = expr->shared_from_this();
```
steadtler: I really wish that we would phase those out.

steadtler:
+1 with one comment.

sql/sql_expression_operations.cc
jeremybarnes: I believe that the pattern is OK for expression trees; otherwise any tree manipulation will need to pass std::shared_ptr<SqlExpression> thisIsASharedPointerToMe (which is what it was early on, and believe me the current way is a lot better).  What we really need to do is to make it impossible to allocate an SqlExpression apart from via std::make_shared<>.

PR #356, opened 2016-05-07
MLDB-1629 fix nesting in column names within join

jeremybarnes:
Fixes problems with the wildcard matching and prefix replacement from MLDB-1486.

### DIFF
--- a/testing/MLDB-1235-temporal-aggregators.py
+++ b/testing/MLDB-1235-temporal-aggregators.py
@@ -439,11 +439,10 @@ class TemporalTest(MldbUnitTest):
                 ]
             )
 
-    @unittest.expectedFailure
     def test_mldbfb_516_aggregator_incorrect_with_join(self):
         ds = mldb.create_dataset({
-            'id' : 'ds_join',
-            'type' : 'beh.binary.mutable'
+            'id' : 'ds516',
+            'type' : 'sparse.mutable'
         })
         ds.record_row('user3', [['behA', 1, 11], ['conv', 1, 70],
                                 ['behB', 1, 14], ['behA', 1, 14]])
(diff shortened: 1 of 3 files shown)


testing/MLDB-180-basic-join.js
```diff
@@ -258,7 +258,7 @@ for (var i in funcs) {
 // MLDB-1088
 
 expected = [
-   [ "_rowName", "test1.x", "test1.y", "test1.z" ],
+   [ "_rowName", "x", "y", "z" ],
```
nicolaskruchten: is this a generalized change? i.e. when I join `a` and `b` the column names are no longer prefixed by `a` and `b` ?

jeremybarnes: No, it's not generalized (we should work on doing that, as it's currently all over the place).  The change I made caused an existing codepath to be used for the two cases, and so it had to change either one way or the other.  Changing it this way led to the smallest number of tests being changed.

nicolaskruchten: ok, so in some cases a join will lead to the table name in the column name and in some cases it won't?

... I really need to spec out the behaviour we want here :(

nicolaskruchten: OK, final answer (until the next change) ... we should keep the prefixes for all joins

jeremybarnes:
Should be good to go now.  There was also an issue in the testcase: temporal_xxx only iterates keys one level down.  So select temporal_min({ds516.*}) will only have one result: ds516, since the output of {ds516.*} is an object call ds516.  To get the temporal aggregation of the next level down, it needs to be temporal_min({ds516.* as *}) to bring the contents of the ds516 object back to the top level.

nicolaskruchten:
why is this PR titled "join rownames"? is there some impact on row names?

jeremybarnes:
It probably has something to do with the original ticket.

testing/MLDB-1235-temporal-aggregators.py
```diff
@@ -439,36 +439,37 @@ def test_mldbfb_344_temporal_segfault(self):
...
+        mldb.log(res)
         self.assertTableResultEquals(res, [
             ['_rowName', 'behA', 'behB', 'conv'],
             ['user3', 2, 1, 1]
         ])
 
         res = mldb.query("""
-            SELECT temporal_count({ds_join.*}) AS *
-            FROM ds_join INNER JOIN conv ON ds_join.rowName() = conv.rowName()
+            SELECT temporal_count({ds516.* as *}) AS *
```
nicolaskruchten: why is the `as *` required here but not above?

nicolaskruchten:
1629 is about inputs to aggregators, and 1637 is about rownames. I think this PR is indeed about 1629 but isn't about rownames :)

testing/MLDB-1235-temporal-aggregators.py
jeremybarnes: That's a long-standing and deeper (but separate) thing.  In the context of a single dataset `x` with columns `a=1` and `b=2`, `select * from x` returns `{ a: 1, b: 2}`, not `{ x: { a: 1, b: 2} }`.   When the support for select `x.* from x` was added, it was done in such a way that the superfluous prefix was detected and simply removed; so that it gives the same result.  In the context of a join, however, the table name is always added.

I would say that choosing and implementing a consistent solution (that's consistent with the standard, or at least with other SQL implementations) is a separate issue.

jeremybarnes:
So what will it take to get a +1 on this pull request?  I don't think renaming the branch should be necessary :)

nicolaskruchten:
@steadtler can you review this PR please?

steadtler:
+1

PR #357, opened 2016-05-07
Made SVD training return row (MLDB-1434)

jeremybarnes:
Makes the SVD output be a real embedding in a column.  Also changes the format of the row names in the column dataset (they now look like "column1.stringEquals.cat" or "column1.stringEmpty" or "column1.isNull" or "column1.numberEquals.5" instead of "column1|" or "column1|cat" or "column1|5".

This will have consequences for client code; this PR is to enable exploration of what those are.

A small improvement to SQL parsing is required to make the mldb_reddit_test (currently failing) pass.


nicolaskruchten:
NB: this refers to MLDB-1634 not MLDB-1434

A few questions:

1. are the changes to the row names of the output datasets of `svd.train` required in order to get the output of `svd.embedRow` to be compatible with the input of `embedding.neighbors` ?
2. I assume that if a column `c`'s unique values are (null, 5) then the rows of the column output will be `c.numberEquals.5` and `c.isNull` right? But if the unique values of `c` are (1,2,3,4,5) then the rows of the column output will just be called `c` because they are not one-hotted?

jeremybarnes:
1.  No, it is not required.  It could be split into two parts if necessary.
2.  Exactly.  I was considering using c.numberValue in order to make this explicit.

nicolaskruchten:
Ok. I like the direction of the column renaming, it uses the new structured
system to nearly solve an annoying problem. In order to really exploit it
though we will need the row path element accessor functions (to be able to
eg join on 'c' only etc)


On Sunday, 8 May 2016, Jeremy Barnes <notifications@github.com> wrote:

>
>    1. No, it is not required. It could be split into two parts if
>    necessary.
>    2. Exactly. I was considering using c.numberValue in order to make
>    this explicit.
>
> —
> <https://github.com/mldbai/mldb/pull/357#issuecomment-217743042>
>


-- 
http://datacratic.com/

nicolaskruchten:
+1 on the requirements, including the `c.numberValue` to make things explicit. Let's get this merged concurrently with MLDB-1636 (you could even just pile those commits on top of this branch)

nicolaskruchten:
@atremblay  reports that when we merge this branch into master, the following functions no longer return anything:

```python
mldb.put("/v1/functions/x", {
        "type":"sql.query",
        "params": {
            "query": "select * from dataset",
        }
    })
```

jeremybarnes:
That doesn't seem to be possible to me; none of the code that has changed is called there.

jeremybarnes:
Code has been added to implement rowPath() and rowPathElement() functions (and their column equivalents).  I don't think that @atremblay 's problem has anything to do with this patch; is it possible that the dataset is empty?

This seems to me to implement the DOD of MLDB-1634, MLDB-1636 and MLDB-1648; please give a +1 if we're good to go or let me know what is missing.

nicolaskruchten:
Unfortunately I still can't use the output of `svd.embedRow` as the input to `embedding.neighbors` ... I'll add to MLDB-1634

jeremybarnes:
Should be fixed now.  Previously the SVD put its output in the "svd" column, but the neighbours had an "embedding" column.  I changed the default SVD column to "embedding".

nicolaskruchten:
OK for the defaults, but this will occur again if a user uses non-default values for the svd output right? we should either make that not configurable in training (my preference) or make it the function pick up the name automatically (harder to document, so not my preference)

nicolaskruchten:
and: w00t, it works! :)

sql/execution_pipeline_impl.cc
```diff
@@ -256,7 +270,7 @@ take()
     //cerr << "got row " << current[currentDone].rowName << " "
     //     << jsonEncodeStr(current[currentDone].columns) << endl;
 
-    result->values.emplace_back(current[currentDone].rowName.toSimpleName(),
+    result->values.emplace_back(current[currentDone].rowName,
```
guyd: Thanks.  I wasn't sure how to store a path in an ExpressionValue.  Having a dedicated cell value type is better.

guyd:
Reviewed briefly.  +1

nicolaskruchten:
Actually there's something fishy here... In order for `embedding.neighbor` to accept the output of `svd.embedRow` I had to *not* extract the embedding, suggesting that the input to `embedding.neighbour` can't be an embedding but rather a row containing an embedding named whatever the embedding is named in the dataset it refers to ... is that right? is that good?

nicolaskruchten:
Finally, upon further thought, I think the rules for svd column names should be as follows:

1. if the columns have multiple numerical values OR a single value+null (string or not) then the rows of the column output should be named `c`.
2. if and only if the columns have more than a single value *of which one is a string*, should the rows be called `c.isNull` vs `c.numberEquals.5` vs `c.stringEquals.thing`.

that will "just work" for most inputs IMO.

sql/execution_pipeline_impl.cc
jeremybarnes: Indeed, we were always going to be fighting uphill to not have one.

nicolaskruchten:
OK, so the interface works and the tests pass etc but I'm seeing very strange results in movie recommendations. I believe I've traced the problem to the `svd.embedRow` function... If I take a vector representing only "toy story has a rating of 5" then the embedded coordinates are nowhere near the coordinates of the "toy story" point. The nearest neighbour is "Top Gun".

jeremybarnes:
So is there something we can do to unblock this issue?  The failing test is clearly invalid.

nicolaskruchten:
I need to validate whether there is a problem in the SVD results. The
output of the movie recommendation demo is radically worse now than before.

On Tuesday, 17 May 2016, Jeremy Barnes <notifications@github.com> wrote:

> So is there something we can do to unblock this issue? The failing test is
> clearly invalid.
>
> —
> <https://github.com/mldbai/mldb/pull/357#issuecomment-219886017>
>


-- 
http://datacratic.com/

nicolaskruchten:
TL;DR I strongly suspect the bug/regression/difference is in the `embedding.neighbors` function...

I trained an SVD and ran the resulting `svd.embedRow` function on one row of input for the tip of master, the tip of this branch, and the latest release. This branch and master had the same `md5sum` output for the `.svd` file, but it differed from the one generated by the latest release. However, the output of `svd.embedRow` is identical for all three cases. I also spot-checked a single row of the column output dataset and the embedding values are identical. I then made sure that my new version of the Movie Reco demo doesn't feed data into the `svd.embedRow` function differently by checking the output of that function on a manually-created row (i.e. one not in the original input) and the output is identical. So there is no change in the SVD part of the world.

That said, the `embedding.neighbors` function gives totally different output on this branch as compared to the previous release's `embedding_dataset/routes/neighbours`. I tend to think that the previous release was giving "more correct" output based on the nature of the recommendations it was returning.

nicolaskruchten:
(I also got rid of the new test-case, as per our conversation)

jeremybarnes:
OK.  So let's hardcode a test case that requires that the output of embedding.neighbors be identical to the last release.

nicolaskruchten:
Update: on this branch, the distance returned along with the nearest neighbour doesn't match the distance I compute by hand outside of MLDB, so that is where the issue lies. I will add a test case to the Reddit test suite for this.

jeremybarnes:
(FWIW I don't think it's this branch per se, but the MLDB-1486 branch that has the problem).

nicolaskruchten:
We can merge now and fix on master

On Thursday, 19 May 2016, Jeremy Barnes <notifications@github.com> wrote:

> (FWIW I don't think it's this branch per se, but the MLDB-1486 branch that
> has the problem).
>
> —
> <https://github.com/mldbai/mldb/pull/357#issuecomment-220320099>
>


-- 
http://datacratic.com/

PR #358, opened 2016-05-07
Allow empty path elements (MLDB-1635)

jeremybarnes:
Self explanatory... this allows for paths to contain empty elements, just like JSON (which allows empty keys).

### DIFF
--- a/sql/testing/path_test.cc
+++ b/sql/testing/path_test.cc
@@ -40,13 +40,6 @@ BOOST_AUTO_TEST_CASE(test_coord_constructor)
 
     vector<PathElement> coords2;
     coords2.push_back(coord1);
-
-    // Make sure we can't create a Path with an empty element
-    {
-        JML_TRACE_EXCEPTIONS(false);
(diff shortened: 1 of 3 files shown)


sql/testing/path_test.cc
```diff
@@ -138,13 +130,23 @@ BOOST_AUTO_TEST_CASE(test_coords_parsing)
     }
 
     {
+        Path coords1 = Path::parse("..");
+        BOOST_CHECK_EQUAL(coords1.size(), 3);
+        BOOST_CHECK_EQUAL(coords1.toUtf8String(), "..");
```
nicolaskruchten: should this not match to `""."".""` ?

steadtler:
+1

sql/testing/path_test.cc
jeremybarnes: The "" aren't actually necessary as it's not ambiguous.  Do you think we should put them in anyway?

nicolaskruchten: I think that it would be preferable to allowing `..` and `x.` or `.x` etc yes

PR #359, opened 2016-05-08
Allow joining on table expressions (eg, row_table) (MLDB-1616)

jeremybarnes:
This allows for joins on things that aren't backed by a dataset, for example row_table(), also fixing a segfault when it was attempted.

The implementation will detect when there is no materialized dataset behind one or both sides of a join, and fall back to the new executor instead.

It also lays the groundwork for correlated subselects.

### DIFF
--- a/builtin/joined_dataset.cc
+++ b/builtin/joined_dataset.cc
@@ -119,12 +119,11 @@ struct JoinedDataset::Itl
     /// Mapping from the table column hash to output column name
     std::unordered_map<ColumnHash, ColumnName> leftColumns, rightColumns;
 
-    /// Datasets that were actually joined.  There will be a maximum of 31
-    /// of them, as any more will be sub-joined
-    std::vector<std::shared_ptr<Dataset> > datasets;
+    /// Structure used to implement operations from each table
+    TableOperations leftOps, rightOps;
 
-    /// Matrix view.  Length is the same as that of datasets.
-    std::vector<std::shared_ptr<MatrixView> > matrices;
+    /// Datasets on the left and right side
+    std::shared_ptr<Dataset> leftDataset, rightDataset;
 
     // is the dataset on the left a join too?
     bool isChainedJoin;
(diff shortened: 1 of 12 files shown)


sql/execution_pipeline.cc
```diff
@@ -251,6 +251,29 @@ doResolveTableName(const ColumnName & fullVariableName,
     return fullVariableName;
 }
 
+std::vector<Utf8String>
+PipelineExpressionScope::
+getTableNames() const
+{
+    std::vector<Utf8String> result;
```
steadtler: Reserve?

steadtler:
+1 with one comment but maybe @guyd would like to take a look as he was working in there last week.

sql/execution_pipeline.cc
jeremybarnes: It's not known ahead of time, and it's only used in binding.  I don't think it will make enough of difference to be worth the complexity of calculating the final size.

jeremybarnes:
OK.  Holding off on the merge until @guyd has had a chance to have a look.

builtin/joined_dataset.cc
```diff
@@ -768,7 +804,8 @@ struct JoinedDataset::Itl
 
         RowName subRowName = JOIN_SIDE_LEFT == side ? entry.leftName : entry.rightName;
 
-        return datasets[side]->getOriginalRowName(tableName, subRowName);
+        return (side ? *rightDataset : *leftDataset)
```
guyd: I would not rely on the enum value and test against `JOIN_SIDE_RIGHT`

sql/execution_pipeline.h
```diff
@@ -331,6 +346,20 @@ struct PipelineElement: public std::enable_shared_from_this<PipelineElement> {
          std::shared_ptr<SqlExpression> where = SqlExpression::TRUE,
          OrderByExpression orderBy = OrderByExpression());
     
+    /** Add a join to the pipeline that is pre-bound into the immediate outer
+        scope.
+    */
+    std::shared_ptr<PipelineElement>
+    join(std::shared_ptr<TableExpression> left,
```
guyd: It seems that we use either the table expression or the provided "pre-bound" expression.  Why don't we have two overloaded functions (here and above) instead?

sql/execution_pipeline_impl.h
```diff
@@ -351,15 +356,23 @@ struct RootElement: public PipelineElement {
 /** Element that generates rows according to the FROM clause. */
 
 struct FromElement: public PipelineElement {
+
+    /** Create the from clause, which will forward a query to the executor
+        of the given table.  The table may be pre-bound (if boundFrom is
+        filled in), or otherwise will be bound by this element (if a default
```
guyd: last parenthesis is probably 

> if no default constructed ...

guyd:
Minor comments.  +1

sql/execution_pipeline.h
jeremybarnes: The second one is an optimization for when the expression has already been bound previously; the first one stayed for compatibility with existing code.  Note that the first forwards to the second in the implementation.

PR #361, opened 2016-05-09
MLDBFB-520 temporal_earliest incorrect result demo

FinchPowers:


guyd:
This is a mix of fixes and clean-up while working on @FinchPowers issue.  Note that the fix to the original issue is still not complete but I am splitting the PR to make it easier to review.  This PR contains:
- a fix to MLDB-1586 which was related (possibly a pre-requisite) to this issue.  `ExpressionValue::searchRow` is now supporting superposition correctly.
- a fixed Doxygen configuration to match the new directory structure (that was broken for a while but since I don't re-generate the doc often, it went unnoticed)
- some clean-up including renaming of ComputedColumn to NamedColumnExpression and using the existing typedef to make `ExpressionValue` functions easier to read

jeremybarnes:
+1.  Are there any non-trivial changes?

guyd:
The bug fix is [here](https://github.com/mldbai/mldb/pull/361/files#diff-093cb1898c54fffbdceac9d90f913359L4586).  Basically, we no longer create the `ExpressionValue` with a `RowValue` which leads to an object but use a `StructValue` which leads to the proper superposition.

PR #362, opened 2016-05-09
Allow structured input column names (MLDB-1638)

jeremybarnes:
Implements the following DOD from the ticket:

- new option on import.text called structuredNames (boolean, defaults to false)
- if true, then row/col names are parsed into paths such that a.b -> [a,b]
- if false, then row/col names are taken as literal strings such that a.b -> ["a.b"] (i.e. no change from current behaviour)

NOTE that the default was true, not false, and so this is a change in behaviour.

@nicolaskruchten you may want to review the documentation and/or add a test case.


plugins/importtext_procedure.cc
```diff
@@ -56,6 +56,12 @@ ImportTextConfigDescription::ImportTextConfigDescription()
     addField("ignoreBadLines", &ImportTextConfig::ignoreBadLines,
              "If true, any line causing a parsing error will be skipped. "
              "Empty lines are considered bad lines.", false);
+    addField("structuredNames", &ImportTextConfig::structuredNames,
```
nicolaskruchten: I thought this would apply to row names as well but clearly it's just column names, so can we actually name this option `structuredColumnNames` please? (sorry)

guyd:
Can we also add one test?

jeremybarnes:
Updated ticket and added test-case as requested.

nicolaskruchten:
+1

PR #363, opened 2016-05-09
Mldb 1241 v1

simlmx:
First version of the Enron Spam Filtering demo. Only a simple bag of word which leads to awesome AUC. We then say "yes but" and propose to use a generalization of the F-score as a more interesting score for our spam filter. The results then don't appear so good but... to be continued!

Replaces the last pull request with essentially the same code. It has been updated with the latest mldb master branch.


nicolaskruchten:
Can you please inline the script into the top of the notebook? I want the demos to be as self-contained as possible :)

nicolaskruchten:
let's skip the big block of data-processing at the top: if you can export.csv the dataset to somewhere on our server, I can save it to http://public.mldb.ai so the beginning of the demo is just one import.text call.

also, please fix the title :)

PR #365, opened 2016-05-11
Fixed trailing comma on header row leading to CSV import failure (MLDB-1649)

jeremybarnes:
Makes import of the following CSV file work:

```
a,b,c,
1,2,3,
```


mailletf:
+1

PR #366, opened 2016-05-11
Mldb 1581 bound parameters inception

steadtler:
New pipeline element to execute sub queries inside the pipeline executor instead of branching to the batch executor.


plugins/sql_functions.cc
```diff
@@ -125,45 +125,7 @@ struct SqlQueryFunctionApplier: public FunctionApplier {
...
-                = getMldbRoot(function->server)
-                ->params(getParamInfo)
-                ->from(config.query.stm->from, config.query.stm->when,
-                       SelectExpression::STAR, config.query.stm->where)
-                ->where(config.query.stm->where)
-                ->select(config.query.stm->orderBy)
-                ->sort(config.query.stm->orderBy)
-                ->select(config.query.stm->select);
-        }
+        pipeline = getMldbRoot(function->server)->statement(*config.query.stm, getParamInfo);
```
jeremybarnes: This is just a no-op refactoring, right, to avoid repeating code?

server/analytics.cc
```diff
@@ -341,42 +341,8 @@ queryFromStatement(SelectStatement & stm,
         if (!params)
             params = [] (const Utf8String & param) -> ExpressionValue { throw HttpReturnException(500, "No query parameter " + param); };
 
+        std::shared_ptr<PipelineElement> pipeline = PipelineElement::root(scope)->statement(stm, getParamInfo);
```
jeremybarnes: And here?

sql/execution_pipeline.cc
```diff
@@ -456,7 +468,14 @@ std::shared_ptr<PipelineElement>
 PipelineElement::
 select(SelectExpression select)
 {
-    return std::make_shared<SelectElement>(shared_from_this(), select);
+    return std::make_shared<SelectElement>(shared_from_this(), select, false /*unique*/);
+}
+
+std::shared_ptr<PipelineElement>
+PipelineElement::
+selectunique(SelectExpression select)
```
jeremybarnes: Use selectUnique. We don't use nocapsorunderscores style anywhere.

sql/execution_pipeline.cc
```diff
@@ -378,12 +389,13 @@ from(std::shared_ptr<TableExpression> from,
...
-     OrderByExpression orderBy)
+     OrderByExpression orderBy,
+     GetParamInfo getParamInfo)
 {
-    return std::make_shared<FromElement>(shared_from_this(), from,
+    return std::make_shared<FromElement>(shared_from_this(), from, 
                                          BoundTableExpression(),
                                          when,
-                                         select, where, orderBy);
+                                         select, where, orderBy, getParamInfo);
```
jeremybarnes: Would be worth a comment to describe why we need params in the table expression.  It's not obvious.

sql/execution_pipeline.h
```diff
@@ -382,8 +383,13 @@ struct PipelineElement: public std::enable_shared_from_this<PipelineElement> {
     select(SelectExpression select);
 
     std::shared_ptr<PipelineElement>
+    selectunique(SelectExpression select);
```
jeremybarnes: camelCase and please add a comment.

sql/execution_pipeline.h
```diff
@@ -382,8 +383,13 @@ struct PipelineElement: public std::enable_shared_from_this<PipelineElement> {
...
+
+    std::shared_ptr<PipelineElement>
     select(std::shared_ptr<SqlExpression> select);
 
+    std::shared_ptr<PipelineElement>
+    statement(SelectStatement& statement, GetParamInfo getParamInfo);
```
jeremybarnes: Comment please.

sql/execution_pipeline_impl.cc
```diff
@@ -22,18 +22,15 @@ using namespace std;
...
 namespace MLDB {
 
-
 /*****************************************************************************/
 /* TABLE LEXICAL SCOPE                                                       */
 /*****************************************************************************/
 
 TableLexicalScope::
-TableLexicalScope(TableOperations table_,
+TableLexicalScope(std::shared_ptr<RowValueInfo> rowInfo,
```
jeremybarnes: I like this!

sql/execution_pipeline_impl.cc
```diff
@@ -953,10 +1128,10 @@ createOutputScope()
          << " total scope is " << tableScope->numOutputFields()
          << endl;
 
-    cerr << "known tables: " << endl;
```
jeremybarnes: Spurious change... not needed.

sql/execution_pipeline_impl.cc
```diff
@@ -1332,6 +1531,8 @@ start(const BoundParameters & getParam) const
     auto result = std::make_shared<Executor>();
     result->parent = this;
     result->source = source_->start(getParam);
+    result->unique = unique;
```
jeremybarnes: I don't understand what the unique is for.

sql/execution_pipeline_impl.h
```diff
@@ -125,6 +125,103 @@ struct GenerateRowsElement: public PipelineElement {
...
 
+/*****************************************************************************/
+/* SUB SELECT LEXICAL SCOPE                                                  */
+/*****************************************************************************/
+
+/** Lexical scope for a sub select.  It allows for the output of the SELECT to be
+    used in wildcards (SELECT * from (SELECT 1 AS X))
+*/
+
+struct SubSelectLexicalScope: public TableLexicalScope {
```
jeremybarnes: Should add to the comment what its output looks like, ie what elements it will add when it runs.

sql/execution_pipeline_impl.h
```diff
@@ -368,9 +465,9 @@ struct FromElement: public PipelineElement {
                 BoundTableExpression boundFrom_,
                 WhenExpression when_,
                 SelectExpression select_ = SelectExpression::parse("*"),
-                std::shared_ptr<SqlExpression> where_
-                    = SqlExpression::parse("true"),
-                OrderByExpression orderBy_ = OrderByExpression());
+                std::shared_ptr<SqlExpression> where_ = SqlExpression::parse("true"),
```
jeremybarnes: Should be SqlExpression::TRUE

jeremybarnes:
The taken and unique thing bothers me.  If we need to take twice, we should be calling restart().  I don't see how we should need it.

This code is complex.  It's worth adding quite a lot more comments and explanation to what you've added.

plugins/sql_functions.cc
steadtler: Yes, I regrouped the code.

steadtler:
Thanks for the review, I'll document better. The 'unique' is for subqueries without dataset. Because there is no FromElement to limit the number of 'take', we need something to tell the pipeline to only take a single result. It seemed more efficient to tell the select to only take 1 result than to create a dummy FromElement (complete with Bound and Executor) down the pipeline which only purpose is to limit to 1 take.

jeremybarnes:
I would create a one-row dataset and use that for the "from" for a from-less query.  To me it's much clearer as to what the semantics of a no-table query are.

steadtler:
@jeremybarnes ok, done!

PR #367, opened 2016-05-11
[MLDB-1652] Added flag to enable single threaded quoted multi-line st…

mailletf:
…ring support in import.text

### DIFF
--- a/plugins/importtext_procedure.cc
+++ b/plugins/importtext_procedure.cc
@@ -1,7 +1,5 @@
 /** importext_procedure.cc
     Mathieu Marquis Bolduc, February 12, 2016
-    Copyright (c) 2016 Datacratic Inc.  All rights reserved.
-
     This file is part of MLDB. Copyright 2016 Datacratic. All rights reserved.
 
     Procedure that reads text files into an indexed dataset.
@@ -81,6 +79,11 @@ ImportTextConfigDescription::ImportTextConfigDescription()
     addField("timestamp", &ImportTextConfig::timestamp,
              "Expression for row timestamp.",
              SqlExpression::parse("fileTimestamp()"));
+    addField("multiLineSupport", &ImportTextConfig::multiLineSupport,
+             "Activates support for columns with multi-line quoted strings. "
+             "This option disables many optimisations and makes the procedure "
+             "run much slower. Only use if necessary. The `offset` parameter "
+             "will not be reliable when this is activated.", false);
 
     addParent<ProcedureConfig>();
     onUnknownField = [] (ImportTextConfig * config,
@@ -310,6 +313,9 @@ Encoding parseEncoding(const std::string & encodingStr)
 
 } // file scope
 
+const string unclosedQuoteError = "Unclosed quoted CSV value";
+const string notEnoughColsError = "not enough columns in row";
+
 /** Parse a single row of CSV into an array of CellValues.
 
     Carefully designed to not perform any memory allocations in the
@@ -481,7 +487,7 @@ parseFixedWidthCsvRow(const char * & line,
             }
 
             if (!ok)
-                errorMsg = "Unclosed quoted CSV value";
+                errorMsg = unclosedQuoteError.c_str();
 
             if (errorMsg)
                 break;
@@ -563,7 +569,7 @@ parseFixedWidthCsvRow(const char * & line,
     }
 
     if (colNum != numColumns)
-        return "not enough columns in row";
+        return notEnoughColsError.c_str();
 
     return errorMsg;
 }
@@ -876,7 +882,7 @@ struct ImportTextProcedureWorkInstance
                            int chunkNum,
                            int64_t lineNum)
 	    {
-	        int64_t actualLineNum = lineNum + lineOffset;
+            int64_t actualLineNum = lineNum + lineOffset;
 #if 0
 	        uint64_t linesDone = totalLinesProcessed.fetch_add(1);
 
@@ -888,7 +894,7 @@ struct ImportTextProcedureWorkInstance
 	        }
 #endif
 
-                // MLDB-1111 empty lines are treated as error
+            // MLDB-1111 empty lines are treated as error
 	        if (length == 0)
 	            return handleError("empty line", actualLineNum, 0, "");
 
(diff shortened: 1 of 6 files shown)


testing/dataset/MLDB-1652_multiline.csv
```diff
@@ -0,0 +1,5 @@
+colA,colB,colC
+a,b,c
+a,"bouya ,
```
nicolaskruchten: I assume that we only wrap around when there's an unclosed quote?

mailletf: yeah. without, it treats it as missing columns and puts them as null, which is the current behaviour

nicolaskruchten: super.

nicolaskruchten:
+1 from an external interface pov

plugins/importtext_procedure.cc
```diff
@@ -81,6 +79,11 @@ ImportTextConfigDescription::ImportTextConfigDescription()
     addField("timestamp", &ImportTextConfig::timestamp,
              "Expression for row timestamp.",
              SqlExpression::parse("fileTimestamp()"));
+    addField("multiLineSupport", &ImportTextConfig::multiLineSupport,
```
guyd: I would personally name it `allowMultiLines` match existing `ignoreBadLines`.

plugins/importtext_procedure.cc
```diff
@@ -81,6 +79,11 @@ ImportTextConfigDescription::ImportTextConfigDescription()
...
+             "Activates support for columns with multi-line quoted strings. "
+             "This option disables many optimisations and makes the procedure "
```
guyd: optimisations -> optimizations

plugins/importtext_procedure.cc
```diff
@@ -81,6 +79,11 @@ ImportTextConfigDescription::ImportTextConfigDescription()
...
+             "run much slower. Only use if necessary. The `offset` parameter "
+             "will not be reliable when this is activated.", false);
```
guyd: I think we should fix the offset problem may be in a separate JIRA.

plugins/importtext_procedure.cc
```diff
@@ -310,6 +313,9 @@ Encoding parseEncoding(const std::string & encodingStr)
 
 } // file scope
 
+const string unclosedQuoteError = "Unclosed quoted CSV value";
```
guyd: I would put it inside the anonymous namespace just in case.

plugins/importtext_procedure.cc
```diff
@@ -876,7 +882,7 @@ struct ImportTextProcedureWorkInstance
                            int chunkNum,
                            int64_t lineNum)
 	    {
-	        int64_t actualLineNum = lineNum + lineOffset;
+            int64_t actualLineNum = lineNum + lineOffset;
```
guyd: Indentation problem.

plugins/importtext_procedure.cc
```diff
@@ -888,7 +894,7 @@ struct ImportTextProcedureWorkInstance
 	        }
 #endif
 
-                // MLDB-1111 empty lines are treated as error
+            // MLDB-1111 empty lines are treated as error
```
guyd: Here too.

plugins/importtext_procedure.cc
```diff
@@ -901,7 +907,7 @@ struct ImportTextProcedureWorkInstance
 
 	        const char * lineStart = line;
 
-                const size_t numInputColumn = inputColumnNames.size();
+            const size_t numInputColumn = inputColumnNames.size();
```
guyd: And here.

plugins/importtext_procedure.cc
```diff
@@ -967,15 +983,52 @@ struct ImportTextProcedureWorkInstance
...
-                         32 /* parallelism */,
-                         startChunk, doneChunk);
+
+        if(!config.multiLineSupport) {
+            forEachLineBlock(stream, onLine, config.limit,
+                             32 /* parallelism */,
+                             startChunk, doneChunk);
+        }
+        else {
+            // very simplistic and not efficient way of doing multi-line. we send
```
guyd: What will happen if the newline is in the header?

| column | long\nexcerpt |

testing/dataset/MLDB-1652_multiline.csv
```diff
@@ -0,0 +1,5 @@
+colA,colB,colC
```
guyd: Will this work if we change the header to

colA, "colB\nmultiline", colC

plugins/importtext_procedure.cc
mailletf: done MLDBFB-524

plugins/importtext_procedure.cc
mailletf: I actually fixed the problem for my lines. There are a bunch of tabs in the file. I'll fix it post merge and have a commit that just replaces all the tabs so the whole file doesn't show up as modified in this pr

plugins/importtext_procedure.cc
mailletf: fixed

plugins/importtext_procedure.cc
```diff
@@ -689,14 +691,36 @@ struct ImportTextProcedureWorkInstance
...
-                vector<string> fields
-                    = ML::expect_csv_row(pcontext, -1, separator);
+                    try {
+                        ML::Parse_Context pcontext(filename,
+                                               header.c_str(), header.length(), 1, 0);
+                        fields = ML::expect_csv_row(pcontext, -1, separator);
+                        break;
+                    }
+                    catch (std::exception & exp) {
+                        if(config.allowMultiLines &&
```
guyd: This is risky.  Someone changing the error message will never think that this could be breaking some functionality.  Can we use a specific Exception type or even better can we change expect_csv_row to return a "line continuation" type of code?

testing/MLDB-1212_csv_import_long_quoted_lines.py
```diff
@@ -63,7 +63,7 @@ def test_multiline(self):
         self.assertTableResultEquals(
             mldb.query("select * from multiline"),
             [
-                ["_rowName", "colA", "colB", "colC"],
+                ["_rowName", "colA", "colB BB", "colC"],
```
guyd: Should this be containing a new line?  The diff does not show it.

guyd:
+1 once the logic using the error message from an exception thrown by a different function is fixed.

plugins/importtext_procedure.cc
mailletf: Agreed on risky. I was leaning on the test breaking to catch such a problem.

PR #368, opened 2016-05-12
MLDB-1654: disable auth requests for /error/

jraby:


nicolaskruchten:
this seems fine, but does it solve the problem in MLDB-1654 ? I'm not seeing the relationship

nicolaskruchten:
(ok, it does mostly resolve the issue)

PR #370, opened 2016-05-13
MLDB-1615 fix bug in row name generator optim

steadtler:


core/dataset.cc
```diff
@@ -1138,7 +1138,7 @@ generateRowsWhere(const SqlBindingScope & scope,
     // Optimize for rowName() IN (constant, constant, constant)
     // Optimize for rowName() IN ROWS / IN KEYS (...)
     auto inExpression = dynamic_cast<const InExpression *>(&where);
-    if (inExpression) 
+    if (inExpression && !inExpression->isnegative)
```
guyd: This disables the optimization for `NOT IN`, isn't?  Shouldn't we handle it inside this optimized path instead?

guyd:
+1

PR #375, opened 2016-05-16
Adds mldbGet/Delete/Post/Put calls to MldbServer

FinchPowers:
@jeremybarnes J'ai pas mal copié le code de cette section. https://github.com/mldbai/mldb/blob/master/plugins/lang/js/mldb_js.cc#L707

Veux-tu que je modifie quoi que ce soit ou est-ce bon comme ça?

DOD: Avoir une interface minimaliste comparable à celle des scripts python et javascript pour effectuer des appels REST avec MLDB en C++.


guyd:
Je pense que ce code devrait etre a l'exterieur de mldb_server.

guyd:
I would implement that as a header-only helper file with inline functions under the ./mldb/http directory.

guyd:
Also, the pull request should convert existing use of `InProcessRestConnection` to these helpers wherever it makes sense.

server/mldb_server.h
```diff
@@ -142,6 +143,28 @@ struct MldbServer: public ServicePeer, public EventRecorder {
     Utf8String prefixUrl(Utf8String url) const;
     std::string prefixUrl(std::string url) const;
     std::string prefixUrl(const char* url) const;
+    
+    InProcessRestConnection mldbPerform(
```
jeremybarnes: Why do we use the mldb prefix here?  It doesn't make sense: we know it's MLDB because this is an MldbServer.

FinchPowers: I agree that "mldbGet" is "ugly". With what would you go? I don't like "get" either because since this class is a server it's unclear what it would be. A couple of alternatives:
* getFromRoute, postToRoute, ...
* getFromUri, postToUri, ...
* getResource, postResource, ...

We can involve the nameing dept. if needed. ;)

jeremybarnes: restGet, restPut, restPost, restDelete

jeremybarnes: Also, I don't think it's ambiguous: it's a REST server, so get is pretty clear.  But if we have a prefix, it should be rest.

FinchPowers:
Done. +1?

jeremybarnes:
+1 alors :)

PR #376, opened 2016-05-17
Mldb 1639 join where

steadtler:
- same row name convention as regular executor
- properly consider outer joins in cross executor
- Allow any kind of 'where' even when there is no 'on' expression


sql/execution_pipeline_impl.cc
```diff
@@ -670,11 +650,21 @@ doGetFunction(const Utf8String & functionName,
...
             -> ExpressionValue
             {
-                return ExpressionValue
-                (leftRowName(args, context).toUtf8String()
-                 + "-"
-                 + rightRowName(args, context).toUtf8String(),
-                 Date::notADate());
+                Utf8String rowName;
+                ExpressionValue left = leftRowName(args, context);
+                if (left.empty())
```
guyd: The ternary operator would make this more readable.

sql/execution_pipeline_impl.cc
```diff
@@ -670,11 +650,21 @@ doGetFunction(const Utf8String & functionName,
...
+                    rowName = "[]-";
+                else
+                    rowName = "[" + left.toUtf8String() + "]-";
+
+
+                ExpressionValue right = rightRowName(args, context);
+                if (right.empty())
```
guyd: Here too

sql/join_utils.cc
```diff
@@ -302,47 +303,62 @@ AnnotatedJoinCondition(std::shared_ptr<TableExpression> leftTable,
...
+
     // No join clause = cross join
-    if (on)
+    if (on) {
         analyze(on);
-    if (where)
+        checkAndClauses(andClauses.begin());
+    }
+
+    size_t numOnCrossConditions = crossConditions.size();
```
guyd: numOnCrossConditions -> numOfCrossConditions?

guyd:
+1

PR #378, opened 2016-05-18
Mldb 1664 join with non-equijoin condition (e.g. inequality)

guyd:
This is fixing one case where a join with cross conditions other than equality was not accepted.

### DIFF
--- a/testing/MLDB-1624-more-join-test.py
+++ b/testing/MLDB-1624-more-join-test.py
@@ -1471,7 +1471,7 @@ class JoinTest(MldbUnitTest):
         )
 
 
-    @unittest.expectedFailure
+    @unittest.expectedFailure # MLDB-1672
     def test_left_outer_join(self):
         """
         row 11 on the left table should be outputted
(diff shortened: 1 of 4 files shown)


testing/MLDB-1624-more-join-test.py
```diff
@@ -1251,34 +1251,27 @@ def test_equi_join(self):
             ]
         )     
 
-    @unittest.expectedFailure
+    # MLDB-1664
```
steadtler: Maybe try with other cross conditions than inegalities, like function calls.

steadtler: @nicolaskruchten This will not be as efficient as could be. Maybe we could create a story to have an efficient codepath for join on inequalities?

guyd: @steadtler Good point for function calls.  I am also fixing cases where ON conditions are more complex.

steadtler:
+1

PR #380, opened 2016-05-18
Mldb 1500 order by aggregators

steadtler:
Allow aggregator functions in order by expressions

### DIFF
--- a/sql/execution_pipeline_impl.cc
+++ b/sql/execution_pipeline_impl.cc
@@ -1870,7 +1870,6 @@ std::shared_ptr<BoundPipelineElement>
 PartitionElement::
 bind() const
 {
-    cerr << numValues << endl;
     return std::make_shared<Bound>(source->bind(), numValues);
 }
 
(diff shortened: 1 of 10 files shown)


sql/execution_pipeline_impl.cc
```diff
@@ -1870,6 +1870,7 @@ std::shared_ptr<BoundPipelineElement>
 PartitionElement::
 bind() const
 {
+    cerr << numValues << endl;
```
guyd: uncommented debug statement

sql/sql_expression.cc
```diff
@@ -1877,12 +1848,13 @@ findAggregators(bool withGroupBy) const
        - "SELECT sum({*}), temporal_earliest({*})" is not valid
        Other invalid SELECT expressions in presence of GROUP BY like referring to a variable
        not in the GROUP BY expression are detected later when binding these expressions.
+       We use the same logic for group by expressions.
```
guyd: You probably mean here 

> for order by expressions.

guyd:
Very good cleanup in sql_expression.  I feel however that the test coverage is a little thin.  Can you add tests that demonstrate that the ordering is as expected and possible tests that demonstrate that the validation of the order by clause in presence of aggregators is correct.

+1 once that's done

PR #386, opened 2016-05-20
Fixed problems with rowName() optimizations & added rowPath() optimizations (MLDB-1678)

jeremybarnes:
1.  Makes sure that the results of optimized queries on rowName() are equivalent to non-optimized versions.  This means parsing the rowNames to look up the parsed version.
2.  Adds rowPath() optimized versions of the same functions.


guyd:
+1

PR #388, opened 2016-05-21
Fixed DT regression, classifier file saving (MLDB-1660)

jeremybarnes:
Fixes two things:
1.  Classifier training silently ignored (with just a log message) when it couldn't save the file; this now will lead to failure
2.  Decision tree would not properly detect when there was no weight in a leaf, and continuously re-split.  This would lead to us trying to serialize a MISSING feature when serializing.


nicolaskruchten:
+1

nicolaskruchten:
thanks for fixing point 1 above, that's the one that had me most worried. is it worth checking all the procedures to make sure they handle this correctly? I don't want procedures to silently fail to write their file, and then reload an old version of the file when they create their functions...

PR #391, opened 2016-05-26
PLAT-1004: ensure the field names are stored using constant pointers …

wsourdeau:
This PR fixes the problem where the pointers returned by the std::string::c_str() are made invalid as field names are added to "fieldNames" on Ubuntu 16.04. This does not affect 14.04, even with g++ 5.3, which makes this PR nice but less urgent.

### DIFF
--- a/types/structure_description.h
+++ b/types/structure_description.h
@@ -67,7 +67,16 @@ struct StructureDescriptionBase {
     typedef std::map<const char *, FieldDescription, StrCompare> Fields;
     Fields fields;
 
-    std::vector<std::string> fieldNames;
+    /* A deleter that works with buffers allocated with malloc */
+    struct FreeDeleter {
+        void operator () (void * p)
+            const
+        {
+            ::free(p);
+        }
+    };
+
+    std::vector<std::unique_ptr<char, FreeDeleter> > fieldNames;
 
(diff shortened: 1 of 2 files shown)


types/structure_description.h
```diff
@@ -171,8 +180,8 @@ struct StructureDescription
         if (fields.count(name.c_str()))
             throw ML::Exception("field '" + name + "' added twice");
 
-        fieldNames.push_back(name);
-        const char * fieldName = fieldNames.back().c_str();
+        fieldNames.emplace_back(::strdup(name.c_str()));
```
jeremybarnes: should use std::strdup not ::

jeremybarnes:
Is there a test case we could write that would show up the problem in 16.04 to avoid regressions?

Minor comments, +1 once addressed.   Thanks for porting this to MLDB.

wsourdeau:
@jeremybarnes the problem was actually revealed while executing value_description_test (case test_structure_description). So this proves our tests are already sufficient.

types/structure_description.h
wsourdeau: Actually strdup is not part of std, due to it being from POSIX rather than the ISO C std. It is not even forwarded by the gnu c++ lib. Therefore using "std::" would cause compilation failures (I checked).

PR #394, opened 2016-05-27
MLDBFB-545 missing filter in merge dataset getColumnValues

steadtler:
This will solve the issue of the WHERE not getting applied on the merged dataset in some cases.


FinchPowers:
+1 pending test

PR #396, opened 2016-05-30
MLDB-1692 comment out part of thread-pool-test depending on timing

steadtler:


jeremybarnes:
Would prefer that do something like if num_cpus() < 4 { return; } at the start than comment out.

PR #397, opened 2016-05-30
MLDB-1617 fix typos in sql parsing

steadtler:


jeremybarnes:
+1

We should consider removing the allowUtf8 flag.  It is no longer very useful.

PR #398, opened 2016-05-30
Mldb 1692 thread pool test

steadtler:
We put an environnement variable on jenkins to disable those tests.


jeremybarnes:
+1

@nicolaskruchten this needs an environment variable to be set in Jenkins

steadtler:
@nicolaskruchten Its already setup.

PR #403, opened 2016-05-31
MLDB-1632 path optimizations

jeremybarnes:
This patch moves from a vector representation to an indexed string representation for paths.  It results in roughly a 2x speed improvement for query performance, due to less pointer chasing, and mostly closes the performance gap with the pre-singularity version of MLDB.

There are other improvements possible; this is the low-hanging fruit.

### DIFF
--- a/mldb_macros.mk
+++ b/mldb_macros.mk
@@ -4,6 +4,7 @@
 # $(3) options for the test
 #      - manual: run the test manually
 #      - virtualenv: set up the Python virtualenv even for a non-python test
+#      - valgrind: run the test within valgrind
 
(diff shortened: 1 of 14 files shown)


steadtler:
+1

nicolaskruchten:
@jeremybarnes the path_test fails in this branch...?

jeremybarnes:
@nicolaskruchten it's valgrind being hyper-sensitive about some static initialization in jml.  I've taken it back out of that test for now.

nicolaskruchten:
Wow, now there are 19 failures! :)

PR #404, opened 2016-06-01
MLDB-1694 flatten embedding

jeremybarnes:
Improves the overall robustness and predictability of MLDB by making sure that the columns with numeric elements sort in natural numerical order.  This happens to fix MLDB-1694, although there is a deeper issue that the embeddings are being turned into non-embeddings as they pass through the select expressions.

It also fixes a problem with null path elements being created by complex joins, which also affected sorting.

Most of the changes are in test-cases.

There appears to be a serious problem identified by the test-case in MLDB-961-glz-categorical.js which @mailletf or @steadtler should have a look at.  That's the only test case that fails.


jeremybarnes:
Also fixed the deeper problem.

steadtler:
+1

nicolaskruchten:
@steadtler can you please look at the failing MLDB-961-glz-categorical.js test case in this PR? Merging this PR is the last known blocker before the release and this test appears hairy according to Jeremy's note above... :)

steadtler:
@nicolaskruchten Im looking and it is hairy!

PR #411, opened 2016-06-06
Path sorting optimizations (MLDB-1699)

jeremybarnes:
Gains back the speed lost from the natural sorting of paths.

### DIFF
--- a/sql/path.cc
+++ b/sql/path.cc
@@ -24,6 +24,114 @@ namespace MLDB {
 
 
 /*****************************************************************************/
+/* COMPARISON FUNCTIONS                                                      */
+/*****************************************************************************/
+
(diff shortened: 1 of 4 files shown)


sql/path.cc
```diff
@@ -261,6 +387,19 @@ bool
 PathElement::
 operator <  (const PathElement & other) const
 {
+    //ExcAssertEqual(digits_, calcDigits(data(), dataLength()));
+    //ExcAssertEqual(other.digits_, calcDigits(other.data(), other.dataLength()));
+
+#if 1
```
steadtler: Do we still need the #if ?

steadtler:
+1, minor comment

PR #414, opened 2016-06-06
MLDBFB-558 need to cast ts issue demo

FinchPowers:


FinchPowers:
Désolé pour le diff, mon vim semble avoir fait le ménage.

nicolaskruchten:
Oh I had already created MLDB-1706 ticket with test cases in the issue... Also, I don't think I agree that `horizontal_avg` should work, or that strings should auto-cast to timestamps :)

FinchPowers:
Ok misunderstanding. I'll rewrite the test.

FinchPowers:
Updated.

nicolaskruchten:
can you add the TIMESTAMP keyword before the timestamps plz? Otherwise the tests are the same: string comparisons. We don't auto-cast ISO strings as timestamps

testing/MLDB-461_horizontal_ops_test.py
```diff
@@ -208,5 +207,29 @@ def test_horizontal_latest(self):
...
+        self.assertEqual(res[1][1], 'b')
 
+        res = mldb.query("""
+            SELECT horizontal_min({'2015-01-01T00:00:00Z',
+                                   '2016-01-01T00:00:00Z'})
+        """)
+        self.assertEqual(res[1][1], '2015-01-01T00:00:00Z')
+
+        res = mldb.query("""
+            SELECT horizontal_max({CAST ('2015-01-01T00:00:00Z' AS TIMESTAMP),
```
nicolaskruchten: can we just use `TIMESTAMP '2015-01-01T00:00:00Z'` which is the preferred way of creating timestamps?

testing/MLDB-461_horizontal_ops_test.py
```diff
@@ -208,5 +207,29 @@ def test_horizontal_latest(self):
...
+        """)
+        self.assertEqual(res[1][1], 'a')
+
+        res = mldb.query("""
+            SELECT horizontal_max({'a', 'b'})
+        """)
+        self.assertEqual(res[1][1], 'b')
 
+        res = mldb.query("""
+            SELECT horizontal_min({'2015-01-01T00:00:00Z',
```
nicolaskruchten: these should be timestamps

FinchPowers:
Comme `SELECT TIMESTAMP '2015-01-01T00:00:00Z'` ?

mailletf:
@FinchPowers oui ajouter TIMESTAMP avant chaque string ayant un timestamp, de manière à le caster, et tester le horizontal_min sur un timestamp et non une string. C'était ça ta question?

PR #415, opened 2016-06-06
MLDB-1705 function application path

jeremybarnes:
Adds a test case for being able to apply functions with structured fields coming in.


mailletf:
+1

PR #416, opened 2016-06-06
Escape \ and $ in command_key. Fixes cache miss.

jraby:
When not escaped, \ and $ would cause a mismatch between the key stored
in the .make_hash_cache file and the key returned to make, thereby
causing costly execution of `bash -c 'echo ... | md5sum'` and adding a
duplicate target to .make_hash_cache

This fixes the cache miss cases and speeds up the build startup time.

### DIFF
--- a/jml-build/functions.mk
+++ b/jml-build/functions.mk
@@ -46,12 +46,13 @@ ifneq ($(__BASH_MAKE_COMPLETION__),1)
 NOTHING :=
 SPACE := $(NOTHING) $(NOTHING)
 TAB := $(NOTHING)	$(NOTHING)
+DOLLAR := $$
 
 hash_command2 = $(wordlist 1,1,$(shell echo $(strip $(1)) | md5sum))
 
 hash_command1 = $(eval HASH:=$(call hash_command2,$(1)))$(shell echo $(1)_hash:=$(HASH) >> .make_hash_cache)$(eval $(1)_hash:=$(HASH))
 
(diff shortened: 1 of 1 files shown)


jeremybarnes:
Well, the test failure has nothing to do with the patch... but it also shows a potentially serious bug in MLDB.  @steadtler it would be worth having a look.

jeremybarnes:
+1 for the code

nicolaskruchten:
What's the speedup like for running a single test? We used to have a 4-8s
hit just on rechecking all the targets...

On Monday, 6 June 2016, Jeremy Barnes <notifications@github.com> wrote:

> +1 for the code
>
> —
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly, view it on GitHub
> <https://github.com/mldbai/mldb/pull/416#issuecomment-224132998>, or mute
> the thread
> <https://github.com/notifications/unsubscribe/AAMbA70LsTF7uaNjMESwEN0aYjX-Stm7ks5qJMI4gaJpZM4IvTzZ>
> .
>


-- 
Nicolas Kruchten, P.Eng.  |  Staff Software Engineer at Datacratic  |
http://datacratic.com/

nicolaskruchten:
And the answer, correctly measured using /usr/bin/time instead of the time shell command, is that we get a consistent 0.4s speedup, on a base of 2.4s, so one-sixth faster. Nice job! :)

/cc @guyd

PR #417, opened 2016-06-06
[MLDB-1701] add elections 2015 demo based on francois' blog post

simlmx:
Add election 2015 demo.


nicolaskruchten:
+1

note: when showing stuff like "top keywords" or "count of PRs by party" if you just put something like `.plot(kind='barh')` at the end of the call, you get a nice readable chart instead of a boring table :)

simlmx:
Do you want me to change those or just for next time?

mailletf:
Je dirais que ça vaut la peine de passer 30 minutes à l'arranger. Sinon on le fera probablement jamais :)

PR #423, opened 2016-06-08
MLDB-1702 changing comparison operators (>, <, etc) to work like arit…

steadtler:
…hmetic operators in regard to timestamp filtering.


FinchPowers:
+1 (Attendre la fin du code freeze avant de merger)

PR #430, opened 2016-06-14
Mldb 1718 invalid utf 8 string long

FinchPowers:


sql/path.h
```diff
@@ -599,11 +599,7 @@ struct Path {
         }
     }
 
-    Path(Path && other) noexcept
```
jeremybarnes: Don't do this.  The function is inlined on purpose so that the construct/move/destroy cycle can be inlined by the compiler.

FinchPowers: :( It's sooooo long to compile whenever we change it though.

PR #432, opened 2016-06-15
Mldb 1710 join test spurious fail

steadtler:
There were two problems, both had to do with row ordering.

The first issue was that in the pipeline executor, the SubSelectElement was ignoring the orderby requested by the join, in cases of equi joins. However when I fixed that, it screwed up the left/right row name extraction when the JoinDataset was using the pipeline executors (because there were more values in the pipeline than the logic expected). That is why I had to implement the "leftRowName" and "rightRowName" functions. It would have been necessary for more complex tests cases anyway.

The second issue was that when some rows had the exact same sorting values in the OrderedExecutor, those rows would not always come in the same order. I fixed that by making sure that in the OrderedExecutor, we always sort by rowHash if all the previous sorting values are equal.

I also found a bug where we accessed the input parameter after its been std::move'd to a member variable. Be careful when moving input values.


builtin/joined_dataset.cc
```diff
@@ -282,6 +263,8 @@ struct JoinedDataset::Itl
 
             PipelineElement::root(scope)
                 ->join(leftExpr, left, rightExpr, right, on, qualification)
+                ->select(SqlExpression::parse("leftRowName()"))
+                ->select(SqlExpression::parse("rightRowName()"))
```
FinchPowers: Comme `leftRowName` et `rightRowName` sont référés dans plusieurs fichiers, je recommande des les extraire dans des variables.

server/bound_queries.cc
```diff
@@ -349,6 +349,8 @@ struct OrderedExecutor: public BoundSelectQuery::Executor {
         ssize_t limit,
         std::function<bool (const Json::Value &)> onProgress)
     {
+        //STACK_PROFILE(OrderedExecutor);
```
FinchPowers: Debug line à supprimer.

steadtler: J'aimerais mieux la garder, j'en ai besoin a tout bout de champ et on les a pour les autres executors

testing/MLDB-1710-left-right-rowname.py
```diff
@@ -0,0 +1,85 @@
...
+        self.assertEqual(res1, expected);
+
+        res1 = mldb.query("select rightRowName() from ds1 left join ds2 on ds1.a = ds2.a + 1")
+        mldb.log(res1)
+
+        expected = [["_rowName", "rightRowName()"],["[x]-[]",""]]
+
+        self.assertEqual(res1, expected);
+
+mldb.run_tests()
```
FinchPowers: Beau travail sur le test python! 👍

FinchPowers:
+1

PR #433, opened 2016-06-15
MLDB-1724 naive bayes

jeremybarnes:
Fixes the Naive Bayes trainer by updating a superseded function signature.


steadtler:
That was simple enough! +1

PR #444, opened 2016-06-20
Mldb 1732 column count in where

FinchPowers:


sql/sql_expression.cc
```diff
@@ -610,6 +610,10 @@ hasUnboundVariables() const
             return true;
     }
 
+    if (funcs.find("columnCount") != funcs.end()) {
```
steadtler: I understand how it fix your case, but thats not really the purpose of this method. If hasUnboundVariables is true, you would expect to have something in the variable or wildcards part of getUnbound.

FinchPowers:
Ok je change ça. Et je viens de réaliser que le premier test échoue aussi. C'est un double bug fix.

core/dataset.cc
```diff
@@ -1617,7 +1617,9 @@ generateRowsWhere(const SqlBindingScope & scope,
     UnboundEntities unbound = where.getUnbound();
 
     // Look for a free variable
-    bool needsColumns = unbound.hasUnboundVariables();
+    bool needsColumns = unbound.hasUnboundVariables() ||
+        unbound.funcs.find("columnCount") != unbound.funcs.end();
```
FinchPowers: Corrige le problème du WHERE.

sql/sql_expression.cc
```diff
@@ -3832,7 +3832,16 @@ SelectStatement::parse(ML::Parse_Context& context, bool acceptUtf8)
     skip_whitespace(context);
 
     //cerr << jsonEncode(statement) << endl;
-    
+
+    if (statement.from->getType() == "null") {
+        for (const auto & c: statement.select.clauses) {
+            if (typeid(*c.get()) == typeid(WildcardExpression)) {
+                throw HttpReturnException(
+                    400, "Wildcard usage requires a FROM statement");
```
FinchPowers: Corrige le problème de "SELECT *" sans FROM.

FinchPowers:
On a un «problème» plus large où les fonctions ne sont pas reconnues s'il n'y a pas de FROM.
Exemple: `SELECT count(*)` retourne `Unable to find function 'count' binding function call 'count(*)'.  The function is not a built-in function, and either it's not a registered user function, or user functions are not available in the scope of the expression.`

sql/sql_expression.cc
```diff
@@ -3832,7 +3832,16 @@ SelectStatement::parse(ML::Parse_Context& context, bool acceptUtf8)
     skip_whitespace(context);
 
     //cerr << jsonEncode(statement) << endl;
-    
+
+    if (statement.from->getType() == "null") {
+        for (const auto & c: statement.select.clauses) {
+            if (typeid(*c.get()) == typeid(WildcardExpression)) {
+                throw ML::Exception(
+                    "Wildcard usage requires a FROM statement");
```
FinchPowers: @jeremybarnes Ça ajoute des validations au parser. J'aimerais que tu donnes ton +1/-1 pour confirmer que ça va dans le sens que tu veux.

FinchPowers: @jeremybarnes ping

jeremybarnes: It doesn't go in the parser.  It should go in the query executor.  The parser deals with syntax, not semantics.

core/dataset.cc
jeremybarnes: This fix feels like a band-aid to me.  We should have an unbound.needsRow() function that looks for hasUnboundVariables() || hasRowFunction(), and implement those in some kind of a generic manner.

sql/sql_expression.cc
```diff
@@ -620,6 +620,13 @@ hasUnboundVariables() const
     return false;
 }
 
+bool
+UnboundEntities::
+hasRowFunctions() const
+{
+    return funcs.find("columnCount") != funcs.end();
```
jeremybarnes: Add a comment here saying that we should make this more generic (nobody will remember to add other functions to this list).  Note that rowName(), rowHash(), rowPath(), leftRowHash(), rightRowHash(), etc are also functions that require the row.

jeremybarnes:
OK to merge, but please add another issue to improve the implementation of the isRowFunction() function.  These should be marked by adding a field in BoundFunction describing if the function requires a row, and then collecting this information from the ExpressionValue info.

PR #446, opened 2016-06-21
MLDB-1724 naive bayes (again)

jeremybarnes:
This time, with an assert that the algorithm actually does something sensible.

It fixed a bug in serialization and reconstitution of the classifier.

Also had to fix up clang compilation as doing it on a plane (first airborne pull request FTW)!


simlmx:
+1
FYI I fixed a few details in the test.

PR #448, opened 2016-06-21
Clang compilation fixes (TRIVIAL)

jeremybarnes:
Another round of fixes to make it all compile with clang again.


FinchPowers:
+1

PR #449, opened 2016-06-21
Compress integer valued columns (MLDB-1742)

jeremybarnes:
Improves the ability to encode integer valued columns effectively and reduce the memory required.

Includes a test-case for fidelity in the reconstruction of the columns including corner cases.


steadtler:
I get the memory improvement, but how does the performance compares? +1 regardless

jeremybarnes:
Performance should be improved, since there is less encoding/decoding in the CellValue format, and there will be less memory and cache pressure.

jeremybarnes:
Test failure is spurious.

PR #455, opened 2016-06-27
Disable access logging for uwsgi_validator_runner

jraby:
Option name is misleading, it disables access logging, not all logging.
Exceptions and operational messages such as child proc restarting are
still logged.


jraby:
Lazy consensus.

PR #464, opened 2016-06-28
Json where filter

FinchPowers:


plugins/json_importer.cc
```diff
@@ -211,6 +250,14 @@ struct JSONImporter: public Procedure {
                 return handleError(exc.what(), actualLineNum, string(line, lineLength));
             }
 
+            if (useWhere) {
+                JsonScope jsonScope(server, expr);
+                auto whereBound = config.where->bind(jsonScope);
```
jeremybarnes: Binding should happen outside of the onLine handler.  Otherwise it will be horrendously slow.

plugins/json_importer.cc
```diff
@@ -68,14 +71,46 @@ JSONImporterConfigDescription()
...
              "with an invalid JSON object will cause an error.", false);
+    addField("where", &JSONImporterConfig::where,
+             "Which lines to use to create rows.",
+             SqlExpression::TRUE);
 
     addParent<ProcedureConfig>();
 }
 
+struct JsonScope : SqlExpressionMldbScope {
+    vector<tuple<PathElement, ExpressionValue>> row;
```
jeremybarnes: You should keep a const reference to the ExpressionValue, and use getNestedColumn(), rather than destructure it (getStructured() is not guaranteed to work and you're making a copy of each row which is highly inefficient).

plugins/json_importer.cc
```diff
@@ -68,14 +71,46 @@ JSONImporterConfigDescription()
     addField("limit", &JSONImporterConfig::limit,
              "Maximum number of lines to process");
     addField("offset", &JSONImporterConfig::offset,
-            "Skip the first n lines.", int64_t(0));
+             "Skip the first n lines.", int64_t(0));
     addField("ignoreBadLines", &JSONImporterConfig::ignoreBadLines,
              "If true, any line causing an error will be skipped. Any line "
              "with an invalid JSON object will cause an error.", false);
+    addField("where", &JSONImporterConfig::where,
```
jeremybarnes: You should mention the default is `where true` here.

jeremybarnes:
A few comments.

plugins/json_importer.cc
FinchPowers: Isn't that covered by the "default" value? (4th parameter)

plugins/json_importer.cc
FinchPowers: How do I do that considering that expr is defined in the onLine handler?

plugins/json_importer.cc
FinchPowers: Nvm. I changed it to push the data via the rowScope instead, which is likely the way it was intended tp be.

plugins/json_importer.cc
```diff
@@ -250,12 +245,9 @@ struct JSONImporter: public Procedure {
                 return handleError(exc.what(), actualLineNum, string(line, lineLength));
             }
 
-            if (useWhere) {
-                JsonScope jsonScope(server, expr);
-                auto whereBound = config.where->bind(jsonScope);
-                if (!whereBound(row, storage, GET_ALL).isTrue()) {
-                    return true;
-                }
+            JsonRowScope row(expr);
```
jeremybarnes: Put this inside an if (useWhere) { ... } so that we don't execute if there is no where clause.

jeremybarnes:
Minor follow-up comments; +1 when addressed.

PR #466, opened 2016-06-28
json import select filter

FinchPowers:
2 nouveautés
* select
* named

### DIFF
--- a/plugins/json_importer.cc
+++ b/plugins/json_importer.cc
@@ -42,6 +42,7 @@ struct JSONImporterConfig : ProcedureConfig {
           limit(-1),
           offset(0),
           ignoreBadLines(false),
+          select(SelectExpression::STAR),
           where(SqlExpression::TRUE)
     {
         outputDataset.withType("tabular");
@@ -53,6 +54,7 @@ struct JSONImporterConfig : ProcedureConfig {
     int64_t limit;
     int64_t offset;
     bool ignoreBadLines;
+    SelectExpression select;
     std::shared_ptr<SqlExpression> where;
 };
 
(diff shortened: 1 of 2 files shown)


plugins/json_importer.cc
```diff
@@ -110,6 +123,57 @@ struct JsonScope : SqlExpressionMldbScope {
...
 
+    GetAllColumnsOutput
+    doGetAllColumns(const Utf8String & tableName,
+                    std::function<ColumnName (const ColumnName &)> keep) override
+    {
+        std::vector<KnownColumn> columnsWithInfo;
+
+        auto exec = [=] (const SqlRowScope & scope, const VariableFilter & filter)
+        {
+            StructValue result;
```
FinchPowers: result.reserve(col size)

plugins/json_importer.cc
```diff
@@ -227,7 +298,7 @@ struct JSONImporter: public Procedure {
         {
             auto & threadAccum = accum.get();
 
-            int64_t actualLineNum = lineNumber + lineOffset;
```
jeremybarnes: int64_t is better here.  It is possible that a 32 bit machine may have more than 4 billion lines.

plugins/json_importer.cc
```diff
@@ -251,22 +322,40 @@ struct JSONImporter: public Procedure {
...
-                    return true;
-                }
-            }
-
             skipJsonWhitespace(*parser.context);
             if (!parser.context->eof()) {
                 return handleError("extra characters at end of line", actualLineNum, "");
             }
 
+            unique_ptr<RowName> rowName;
```
jeremybarnes: Don't use a unique_ptr here.  It adds a useless alloc.  Use a rowName and assign it directly.

jeremybarnes:
More comments, +1 once addressed.

PR #468, opened 2016-06-29
Removed spdlog from public API (MLDB-1757)

jeremybarnes:
Avoids including spdlog as part of the plugin API headers and thus exposing our users to it.

### DIFF
--- a/core/procedure.h
+++ b/core/procedure.h
@@ -12,7 +12,6 @@
 #include "mldb/rest/rest_entity.h"
 #include "mldb/sql/sql_expression.h"
 #include "mldb/sql/sql_expression_operations.h"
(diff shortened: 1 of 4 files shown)


core/procedure.h
```diff
@@ -12,7 +12,7 @@
 #include "mldb/rest/rest_entity.h"
 #include "mldb/sql/sql_expression.h"
 #include "mldb/sql/sql_expression_operations.h"
-#include "mldb/utils/log.h"
+//#include "mldb/utils/log_fwd.h"
```
mailletf: remove the line alltogether?

mailletf:
+1 once tests pass

PR #469, opened 2016-06-29
Updated name for row stream (MLDB-1754)

jeremybarnes:
Renamed a hard to understand field.


server/bound_queries.cc
```diff
@@ -238,7 +238,7 @@ struct UnorderedExecutor: public BoundSelectQuery::Executor {
         // Do we select *?  In that case we can avoid a lot of copying
         bool selectStar = boundSelect.expr->isIdentitySelect(context);
 
-        int numRows = whereGenerator.upperBound;
+        int64_t numRows = whereGenerator.rowStreamTotalRows;
```
FinchPowers: Devrait pas être unsigned?

FinchPowers:
Commentaire mineur.
+1 conditionnel au succès des tests.

server/bound_queries.cc
jeremybarnes: No.  We use -1 as a guard value, which we don't want to become MAXINT if we don't check the value properly.

jeremybarnes:
And another spurious test failure.

PR #472, opened 2016-06-30
Sharded row index for Tabular dataset (MLDB-1762)

jeremybarnes:
Also fixes issues with row name generation when offset and limit aren't at their default values.

This reduces by about 80% the time elapsed creating the row index of a large tabular dataset post-import.

### DIFF
--- a/plugins/tabular_dataset.cc
+++ b/plugins/tabular_dataset.cc
@@ -199,8 +199,15 @@ struct TabularDataset::TabularDataStore: public ColumnIndex, public MatrixView {
     // Everything below here is protected by the dataset lock
     std::vector<TabularDatasetChunk> frozenChunks;
 
+    static int getRowShard(RowHash rowHash)
+    {
+        return (rowHash.hash() >> 23) % ROW_INDEX_SHARDS;
+    }
+
     /// Index from rowHash to (chunk, indexInChunk) when line number not used for rowName
-    ML::Lightweight_Hash<RowHash, std::pair<int, int> > rowIndex;
+    static constexpr size_t ROW_INDEX_SHARDS=32;
+    ML::Lightweight_Hash<RowHash, std::pair<int, int> > rowIndex
+        [ROW_INDEX_SHARDS];
     std::string filename;
     Date earliestTs, latestTs;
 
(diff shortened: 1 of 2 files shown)


plugins/tabular_dataset.cc
```diff
@@ -577,17 +597,65 @@ struct TabularDataset::TabularDataStore: public ColumnIndex, public MatrixView {
...
+                    int shard = getRowShard(rowHash);
+                    toInsert[shard].emplace_back(rowHash, j);
+                }
+
+                // Secondly, add them to the row index
+                for (size_t i = 0;  i < ROW_INDEX_SHARDS;  ++i) {
+                    size_t shard = (i + chunkNum) % ROW_INDEX_SHARDS;
+                    std::unique_lock<std::mutex> guard(rowIndexLock[shard]);
+                    
+                    for (auto & e: toInsert[shard]) {
```
FinchPowers: const

plugins/tabular_dataset_chunk.h
```diff
@@ -102,6 +102,16 @@ struct TabularDatasetChunk {
         }
     }
 
+    RowName getRowName(size_t index) const
+    {
+        return rowNames.at(index);
+    }
+
+    const RowName & getRowName(size_t index, RowName & storage) const
```
FinchPowers: À quoi sert cette signature si storage n'est pas lu? C'est pcq ça override?

plugins/tabular_dataset.cc
jeremybarnes: Why?  It adds noise, and we don't care whether or not they are modified

plugins/tabular_dataset_chunk.h
jeremybarnes: It's for compatibility with the next PR, which has that interface (it's a way to return something that can be a reference if it already exists, or will outlive the function if not, without doing an allocation).

plugins/tabular_dataset.cc
FinchPowers: Ok. Je pensais qu'on était religieux à propos de ça. ;)

FinchPowers:
+1

PR #475, opened 2016-06-30
Reduced storage requirements where row names are integers (MLDB-1763)

jeremybarnes:
Goes from roughly 80 byte overhead per row to a 16 byte overhead when row names are just integers.

### DIFF
--- a/sql/path.cc
+++ b/sql/path.cc
@@ -538,7 +538,9 @@ toIndex() const
     const char * p = data();
     const char * e = p + dataLength();
     if (e == p)
-        return false;
+        return -1;
+    if (*p == '0' && dataLength() != 1)
+        return -1;
     for (; p != e;  ++p) {
         if (!isdigit(*p))
             return -1;
@@ -982,6 +984,25 @@ toUtf8String() const
         result += at(i).toEscapedUtf8String(); 
         first = false;
     }
+    return result;
+}
+
+ssize_t
+Path::
+toIndex() const
+{
+    if (length_ != 1 || digits(0) != PathElement::DIGITS_ONLY)
+        return -1;
+    return at(0).toIndex();
+}
+
+size_t
+Path::
+requireIndex() const
+{
+    ssize_t result = toIndex();
+    if (result == -1)
+        throw HttpReturnException(400, "Path was not an index");
     return result;
 }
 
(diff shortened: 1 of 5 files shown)


sql/path.cc
```diff
@@ -985,6 +987,25 @@ toUtf8String() const
     return result;
 }
 
+ssize_t
+Path::
+toIndex() const
+{
+    if (length_ != 1 || digits(0) != PathElement::DIGITS_ONLY)
+        return -1;
+    return at(0).toIndex();
```
FinchPowers: `return isIndex() ? at(0).toIndex() : -1;`

sql/testing/path_test.cc
```diff
@@ -231,10 +231,10 @@ BOOST_AUTO_TEST_CASE(test_indexes)
...
     BOOST_CHECK_EQUAL(PathElement(0).toIndex(), 0);
     BOOST_CHECK_EQUAL(PathElement("0").toIndex(), 0);
-    BOOST_CHECK_EQUAL(PathElement("00").toIndex(), 0);
+    BOOST_CHECK_EQUAL(PathElement("00").toIndex(), -1);
     BOOST_CHECK_EQUAL(PathElement(123456789).toIndex(), 123456789);
     BOOST_CHECK_EQUAL(PathElement("123456789").toIndex(), 123456789);
-    BOOST_CHECK_EQUAL(PathElement("0123456789").toIndex(), 123456789);
+    BOOST_CHECK_EQUAL(PathElement("0123456789").toIndex(), -1);
     BOOST_CHECK_EQUAL(PathElement(-1).toIndex(), -1);
     BOOST_CHECK_EQUAL(PathElement(-1000).toIndex(), -1);
```
FinchPowers: Je vois des exemples de tests négatifs (-1) mais exemple où «ça fonctionne» (>= 0).

FinchPowers:
Vois ce que tu fais avec les commentaires.
+1

sql/testing/path_test.cc
jeremybarnes: BOOST_CHECK_EQUAL(PathElement("123456789").toIndex(), 123456789);

That tests the positive side.

sql/path.cc
jeremybarnes: In this case, we know that we have length 1.  So toIndex() (which returns -1 if not) is the same as at(0).toIndex().

PR #477, opened 2016-06-30
MLDB-1759 fix err 500 when using * with GROUP BY

FinchPowers:


server/bound_queries.cc
```diff
@@ -1449,6 +1449,13 @@ execute(RowProcessor processor,
     typedef std::map<RowKey, GroupMapValue> GroupByMapType;
     std::vector<GroupByMapType> accum(numBuckets);
 
+    for (const auto & c: select.clauses) {
+        if (typeid(*c.get()) == typeid(WildcardExpression)) {
```
jeremybarnes: I would implement an isWildcardExpression() function in SqlExpression, with a default implementation that returns false, and override this for WildcardExpression.  It reduces coupling and will allow (for example) a derivative of WildcardExpression to work properly.  Direct typeid checks like that should usually be replaced with a virtual method.

jeremybarnes:
Right direction, but will be better with a virtual method.

server/bound_queries.cc
jeremybarnes: Also, a test based upon dynamic_cast is also more reliable as it will work for derived classes.

FinchPowers: Bon ben fascinant. Ladite fonction existait déjà. :p

server/bound_queries.cc
```diff
@@ -1449,6 +1449,13 @@ execute(RowProcessor processor,
     typedef std::map<RowKey, GroupMapValue> GroupByMapType;
     std::vector<GroupByMapType> accum(numBuckets);
 
+    for (const auto & c: select.clauses) {
+        if (c->isWildcard()) {
```
jeremybarnes: You shouldn't use a function like isWildcard().  It makes the API too heavy.  You should compare the typeid() instead.

FinchPowers: Are you trolling me?

jeremybarnes: Yes, I was trolling you.  +1 to merge.

FinchPowers: Haha. Je ne suis pas assez habitué à ça ;)

PR #480, opened 2016-07-01
MLDB-1768 naive bayes categorical

jeremybarnes:
Fixes classifier.apply assuming that the optimize() function of its classifiers would always work, and then the test-case.

This should make naive bayes officially unborken again.


mailletf:
+1 on the code directly modified. I didn't dig to see the implication of the optInfo not being defined

simlmx:
+1 same...

jeremybarnes:
The implication was that the classifier would fail, as we were calling a method that asserted on the optinfo not being false.

PR #481, opened 2016-07-01
MLDB-1766 dt categorical iris

jeremybarnes:
Fixes a refactoring bug that always caused the confusion matrix to be diagonal.


mailletf:
+1

PR #482, opened 2016-07-01
Copying optimizations for query execution (MLDB-1770)

jeremybarnes:
Optimization to reduce copying that becomes a bottleneck when creating small embeddings like 3-space coordinates.


FinchPowers:
+1

PR #483, opened 2016-07-01
Mldb 1771 cls categorial weighting

mailletf:


jeremybarnes:
Why did the mldb-961 test results change?

mailletf:
What I'm thinking is that it's because the weighting of the classes has changed, leading to a different weighting for each example, leading to a slightly different model.

jeremybarnes:
Why did the weighting of the classes change?  If it was due to a bug fix, good.  I can't see if this was an expected change or not.

mailletf:
That's the PR. Before it was only weighting labels True/False. It was hardcoded for boolean. Now, if it's categorical, it does more than 2. Since the glz test was on iris and there are 3 classes, the 3rd one used to not get weighted and now it does.

jeremybarnes:
+1

PR #484, opened 2016-07-04
Mldbfb 601 http client integration

wsourdeau:
This PR provides an implementation of HTTPClient from upstream, as a milestone towards MLDBFB-239. The curl-based implementation was chosen as its performance was sufficiently good for S3 and as it is considered to be the most stable one.
Additionally:
* some related classes have been fixed after bugs were revealed
* libc_value_description has been added

### DIFF
--- a/http/http_client.h
+++ b/http/http_client.h
@@ -1,0 +1,175 @@
+/** http_client.h                                                   -*- C++ -*-
+    Wolfgang Sourdeau, January 2014
+    This file is part of MLDB. Copyright 2014 Datacratic. All rights reserved.
+
+    An asynchronous HTTP client.
+
+    HttpClient is meant to provide a featureful, generic and asynchronous HTTP
+    client class. It supports strictly asynchronous (non-blocking) operations,
+    HTTP pipelining and concurrent requests while enabling streaming responses
+    via a callback mechanism. It is meant to be subclassed whenever a
+    synchronous interface or a one-shot response mechanism is required. In
+    general, the code should be complete enough that existing and similar
+    classes could be subclassed gradually (HttpRestProxy, s3 internals). As a
+    generic class, it does not make assumptions on the transferred contents.
+    Finally, it is based on the interface of HttpRestProxy.
+
+    Caveat:
+    - since those require header interpretation, there is not support for
+    cookies per se
+*/
+
+#pragma once
+
+#include <memory>
+#include <string>
+
+#include "mldb/http/http_client_callbacks.h"
+#include "mldb/http/http_header.h"
+#include "mldb/http/http_request.h"
+
+
+namespace Datacratic {
+
+/* Forward declarations */
+
+struct HttpClientImpl;
+struct LegacyEventLoop;
+
+
+/****************************************************************************/
+/* HTTP CLIENT                                                              */
+/****************************************************************************/
+
+struct HttpClient
+{
+    /** This sets the requested version of the underlying HttpClientImpl. By
+     * default, this value is deduced from the "HTTP_CLIENT_VERSION"
+     * environment variable. It not set, this falls back to 1. */
+    static void setHttpClientImplVersion(int version);
+
+    /** HttpClient constructor.
+       "eventLoop": LegacyEventLoop to which the instance will be attached,
+       the loop *must* be started when the destructor is called.
+       "baseUrl": scheme, hostname and port (scheme://hostname[:port]) that
+       will be used as base for all requests
+       "numParallels": number of requests that can be handled simultaneously
+       "queueSize": size of the backlog of pending requests, after which
+       operations will be refused (0 = infinite)
+       "implVersion": use version X of the HttpClientImpl, fallback
+       to HTTP_CLIENT_IMPL
+    */
+    HttpClient(LegacyEventLoop & eventLoop,
+               const std::string & baseUrl,
+               int numParallel = 1024, int queueSize = 0,
+               int implVersion = 0);
+    HttpClient(HttpClient && other) noexcept;
+    HttpClient(const HttpClient & other) = delete;
+
+    ~HttpClient();
+
+    /* AsyncEventSource interface */
+    virtual int selectFd() const;
+    virtual bool processOne();
+
+    /** Enable debugging */
+    void enableDebug(bool value);
+
+    /** SSL checks */
+    void enableSSLChecks(bool value);
+
+    /** Enable the TCP_NODELAY option, also known as the Nagle's algorithm */
+    void enableTcpNoDelay(bool value);
+
+    /** Enable the requesting of "100 Continue" responses in preparation of
+     * a PUT request */
+    void sendExpect100Continue(bool value);
+
+    /** Use with servers that support HTTP pipelining */
+    void enablePipelining(bool value);
+
+    /** Performs a GET request, with "resource" as the location of the
+     *  resource on the server indicated in "baseUrl". Query parameters
+     *  should preferably be passed via "queryParams".
+     *
+     *  Returns "true" when the request could successfully be enqueued.
+     */
+    bool get(const std::string & resource,
+             const std::shared_ptr<HttpClientCallbacks> & callbacks,
+             const RestParams & queryParams = RestParams(),
+             const RestParams & headers = RestParams(),
+             int timeout = -1)
+    {
+        return enqueueRequest("GET", resource, callbacks,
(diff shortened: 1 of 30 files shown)


http/http_client.h
```diff
@@ -0,0 +1,432 @@
...
+   - since those require header interpretation, there is not support for
+     cookies per se
+*/
+
+#pragma once
+
+#include <memory>
+#include <string>
+
+#include "mldb/ext/jsoncpp/value.h"
```
jeremybarnes: Would be good to not include the async_event_source header here, and to put more of the methods in the .cc file.

http/http_client.h
```diff
@@ -0,0 +1,432 @@
...
+    RestParams headers_;
+    int timeout_;
+};
+
+
+/****************************************************************************/
+/* HTTP CLIENT IMPL                                                         */
+/****************************************************************************/
+
+struct HttpClientImpl : public AsyncEventSource {
```
jeremybarnes: Could we use composition rather than inheritance here?

http/http_client.h
```diff
@@ -0,0 +1,432 @@
...
+};
+
+std::ostream & operator << (std::ostream & stream, HttpClientError error);
+
+
+/****************************************************************************/
+/* HTTP CLIENT                                                              */
+/****************************************************************************/
+
+struct HttpClient : public AsyncEventSource {
```
jeremybarnes: Same here.

http/http_client.h
```diff
@@ -0,0 +1,432 @@
...
+
+    /* "baseUrl": scheme, hostname and port (scheme://hostname[:port]) that
+       will be used as base for all requests
+       "numParallels": number of requests that can be handled simultaneously
+       "queueSize": size of the backlog of pending requests, after which
+       operations will be refused (0 = infinite)
+       "implVersion": use version X of the HttpClientImpl, fallback
+       to HTTP_CLIENT_IMPL
+    */
+    HttpClient(const std::string & baseUrl,
```
jeremybarnes: Should all be in the .cc file.

http/http_client.h
```diff
@@ -0,0 +1,432 @@
...
+
+
+/****************************************************************************/
+/* HTTP CLIENT SIMPLE CALLBACKS                                             */
+/****************************************************************************/
+
+/* This class is a child of HttpClientCallbacks and offers a simplified
+ * interface when support for progressive responses is not necessary. */
+
+struct HttpClientSimpleCallbacks : public HttpClientCallbacks
```
jeremybarnes: Why do we need a second class here?

http/tcp_socket_handler_impl.h
```diff
@@ -64,6 +64,7 @@ struct TcpSocketHandlerImpl {
     typedef std::function<void(const boost::system::error_code & ec,
                                size_t bufferSize)> OnReadSome;
     OnReadSome onReadSome_;
+    bool closed_;
```
jeremybarnes: atomic?

jeremybarnes:
General comments
- Attempt to have as few includes and inlined methods as possible.  MLDB's compilation is slowed down by them and they tend to inroduce ABI couplings
- Please use Doxygen style /** comments for the method documentation, as Doxygen is used by the team to help generate code
- Please add a PR description (copied from the ticket is fine) as well as an indication as to what this will improve, and what the path forward is for further changesets that will come on top of this one

There were a few other minor comments in the code, but generally it seems to be in decent shape.

http/http_client.h
wsourdeau: HttpClient is meant to be added to a MessageLoop (pending the asio version). In this case, it's required by the interface of MessageLoop::addSource.

http/http_client.h
wsourdeau: HttpClientCallbacks is the full version that enables on-the-fly handling of requests and can therefore provide a version that is more memory efficient. The HttpClientSimpleCallbacks version enables easy handling of the requests through the need of a single callback with, in counterpart, the requirement to keep the full response body in memory.

http/http_client.h
jeremybarnes: You could still decouple them, by having a getSource() method that returned a reference to an internal class that implemented MessageLoop::addSource.  The users of HttpClient shouldn't need to drag in files based upon the implementation.

http/http_client.h
jeremybarnes: Why not make the HtpClientCallbacks implement the simple version by default?  Then we only have one class, and can override the other.

http/http_client.h
wsourdeau: Good idea. Then I could require the constructor to take a MessageLoop as parameter, similarly to how the EventLoop class is used. Would take make sense?

http/http_client.h
wsourdeau: I could be it does mix both point of views in the same class, is that desirable?

http/tcp_socket_handler_impl.h
wsourdeau: Indeed!

http/http_client.h
wsourdeau: Done. I have segregated most classes in their own files and put all the code in the .cc. Should make things much clearer and reduce the number of required includes.

http/http_client.h
jeremybarnes: Exactly.  And that parameter can be an opaque type.

wsourdeau:
@jeremybarnes I have fixed the documentation for HttpClient, HttpClientImpl and LegacyEventLoop. Also, I have added a new module named "libc_value_description" with descriptions for struct provided by the libc (rusage, timeval).

If that looks fine to you, can I have your +1 ?

http/http_client_impl_v1.cc
```diff
@@ -0,0 +1,669 @@
...
+    string headerLine(data, size);
+    if (headerLine.find("HTTP/1.1 100") == 0) {
+        afterContinue_ = true;
+    }
+    else if (afterContinue_) {
+        if (headerLine == "\r\n")
+            afterContinue_ = false;
+    }
+    else {
+        const auto & cbs = request_->callbacks();
```
FinchPowers: Je suppose que cette fonction aussi est noexcept?

http/http_client_impl_v1.cc
```diff
@@ -0,0 +1,669 @@
...
+        if (headerLine.find("HTTP/") == 0) {
+            size_t lineSize = headerLine.size();
+            size_t oldTokenIdx(0);
+            size_t tokenIdx = headerLine.find(" ");
+            if (tokenIdx == string::npos || tokenIdx >= lineSize) {
+                throw ML::Exception("malformed header");
```
FinchPowers: LOL. noexcept mais ça throw?!?

http/http_client_impl_v1.cc
```diff
@@ -0,0 +1,669 @@
...
+    return size;
+}
+
+size_t
+HttpClientImplV1::
+HttpConnection::
+onCurlWrite(const char * data, size_t size)
+    noexcept
+{
+    request_->callbacks()->onData(*request_, data, size);
```
FinchPowers: noexcept aussi?

http/http_client_impl_v1.h
```diff
@@ -0,0 +1,146 @@
...
+/****************************************************************************/
+
+struct HttpClientImplV1 : public HttpClientImpl {
+    HttpClientImplV1(const std::string & baseUrl,
+                     int numParallel, int queueSize);
+    ~HttpClientImplV1();
+
+    /* AsyncEventSource */
+    virtual int selectFd() const;
+    virtual bool processOne();
```
FinchPowers: override specifier si je ne me trompe pas?
http://en.cppreference.com/w/cpp/language/override

FinchPowers:
L'entête des fichiers devrait avoir le format suivant.
```
<filename>
<author>, <date>
This file is part of MLDB. Copyright <year> Datacratic. All rights reserved.
```

FinchPowers:
Donc, 
* quelques noexcept à vérifier
* des "override" à ajouter là où applicable
* les entêtes à uniformiser

Beau travail sur les tests. Je suppose que la couverture est bonne?

wsourdeau:
@FinchPowers pour les tests oui, la couverture est pratiquement complète. Chaque bug découvert a apporté son test depuis la version initiale, donc c'est relativement solide.

FinchPowers:
+1

PR #487, opened 2016-07-05
Added testcase to verify fix (MLDB-1502)

jeremybarnes:
Verifies that MLDB-1502 is fixed and stays fixed.


FinchPowers:
+1 quand les tests passeront

PR #488, opened 2016-07-05
Added testcase to verify fix (MLDB-1503)

jeremybarnes:
Simple testcase to verify fix.

### DIFF
--- a/testing/MLDB-1503-bad-rowname-error-message.js
+++ b/testing/MLDB-1503-bad-rowname-error-message.js
@@ -1,0 +1,51 @@
+// This file is part of MLDB. Copyright 2016 Datacratic. All rights reserved.
+
+function assertEqual(expr, val, msg)
+{
+    if (expr == val)
+        return;
+    if (JSON.stringify(expr) == JSON.stringify(val))
+        return;
+
+    throw "Assertion failure: " + msg + ": " + JSON.stringify(expr)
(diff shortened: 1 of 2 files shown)


testing/MLDB-1503-bad-rowname-error-message.js
```diff
@@ -0,0 +1,51 @@
...
+function assertEqual(expr, val, msg)
+{
+    if (expr == val)
+        return;
+    if (JSON.stringify(expr) == JSON.stringify(val))
+        return;
+
+    throw "Assertion failure: " + msg + ": " + JSON.stringify(expr)
+        + " not equal to " + JSON.stringify(val);
+}
```
FinchPowers: Unrelated: Any way we can extract that? We could consider putting it into mldb js handle. Copy/pasting it everywhere clutters every test IMO.

FinchPowers:
I assume that before the fix that test was failing? +1.

testing/MLDB-1503-bad-rowname-error-message.js
jeremybarnes: Yes, I would like to add the ability to require a js file with this kind of thing in.  We should add it one day soon...

jeremybarnes:
No, it wasn't failing before.  At some point of time in the past it was but was fixed in the meantime; this takes the reproducing case from the ticket and makes sure it stays fixed.

PR #489, opened 2016-07-05
Made test assert on something & actually fixed (MLDB-1502)

jeremybarnes:
The previous PR didn't add the runOnCreation flag, and didn't query the result, and so was a spurious pass.  This fixes the testcase and the underlying issue.  For real this time.


FinchPowers:
+1

PR #490, opened 2016-07-05
Don't perform extra SqlRowScope type checking (MLDB-1421)

jeremybarnes:
Saves some slightly more expensive checks when converting row scopes.


FinchPowers:
+1

PR #491, opened 2016-07-05
MLDB-1755 column execution memory use

jeremybarnes:
These changes:

1.  Set up a mechanism by which optimized paths that are supposedly identical to non-optimized paths can be instrumented, so that unit tests may be run against both paths and the results compared
2.  Improves the memory and runtime efficiency of executing column expressions, by doing them chunk by chunk and short-circulting direct column extractions.

They allow for a simple 6 variable select from a tabular dataset to run about 10 times faster.


base/optimized_path.cc
```diff
@@ -0,0 +1,188 @@
...
+{
+    auto level = itl->level;
+    if (level == DEFAULT)
+        level = defaultLevel;
+
+    bool result;
+    if (level == NEVER) {
+        result = false;
+    }
+    if (level == ALWAYS) {
```
FinchPowers: else if

base/parallel.cc
```diff
@@ -65,6 +66,65 @@ void parallelMap(size_t first, size_t last,
...
 }
 
+bool parallelMapHaltable(size_t first, size_t last,
+                         const std::function<bool (size_t)> & doWork,
+                         int occupancyLimit)
+{
+    ExcAssertGreaterEqual(last, first);
+    ExcAssertLess((last - first), 1ULL << 31);
+
+    std::atomic<int> has_exc(0);
```
FinchPowers: Why snake case? Does it have special meaning?

FinchPowers:
* Tests failing
* Some code looks untested.

base/parallel.cc
jeremybarnes: no, no special meaning.

base/optimized_path.cc
```diff
@@ -0,0 +1,188 @@
...
+
+    bool result;
+    if (level == NEVER) {
+        result = false;
+    }
+    else if (level == ALWAYS) {
+        result = true;
+    }
+    else {
+        result = rand_r(&itl->randSeed) < level;
```
FinchPowers: If I've followed properly, this will always yield true. randSeed = 1, the only uncovered level is SOMETIMES which is max int / 2. We end up with
rand_r(1) < max int / 2

base/optimized_path.cc
```diff
@@ -0,0 +1,188 @@
...
+    return result;
+}
+
+bool
+OptimizedPath::
+maybeTake(bool canTake) const
+{
+    if (!canTake) {
+        itl->ineligible.fetch_add(1, std::memory_order_relaxed);
+        return false;  // TODO: record statistics
```
FinchPowers: Ticket?

FinchPowers:
`setPathOptimizationLevel` should also be implemented in the python bindings.

base/optimized_path.cc
jeremybarnes: No.  rand_r takes a seed, which allows it to be re-entrant.  The value is always between 0 and RAND_MAX.

base/optimized_path.cc
jeremybarnes: It was done.  I just removed the comment.

base/optimized_path.cc
FinchPowers: :P I only read the first line of the man description. My bad.

FinchPowers:
+1

jeremybarnes:
I added the Python binding.

PR #493, opened 2016-07-05
MLDB-1776 on progress fix

FinchPowers:


FinchPowers:
Verbal +1 from @mailletf

PR #494, opened 2016-07-05
Mldb 1773 json parse fixes

mailletf:


types/json_printing.h
```diff
@@ -215,6 +215,25 @@ struct StringJsonPrintingContext
...
+
+    TODO: we should change this interface (and that of the previous class)
+    to have a str() method rather than pass in its result by reference.
+*/
+
+struct Utf8StringJsonPrintingContext
+    : public StringJsonPrintingContext {
+
+    Utf8StringJsonPrintingContext(Utf8String & str);
+    Utf8String & str;
```
mailletf: why not have this const & and the constructor taking a const & ? instead of the cast_const?

mailletf:
+1

PR #496, opened 2016-07-06
Mldb 1778 json parse

mailletf:


jeremybarnes:
Should update the doc then.  According to the JSON spec, valid JSON is objects or arrays, not atoms.  We should identify in the doc that we support more than that.

mailletf:
Ok will do. jq parsed it ok so I assumed it was fine.

PR #497, opened 2016-07-07
Fixed pathnames for COLUMN EXPR (MLDB-1779)

jeremybarnes:
It was assuming strings and using columnName() not columnPath() as the default AS


FinchPowers:
+1

PR #499, opened 2016-07-07
Mldbfb 615 libservices cleanup

wsourdeau:
This PR cleans up soa/service, in order to reduce the risk of dependency issues in future enhancements.

* a libio_base lib is created and includes EventLoop, MessageLoop and associated classes (including the base tcp handler classes for now)
* liblogging is created and includes the Logging class
* librunner is created and includes the Runner class
* libhttp is slimmed down to only include only the classes that have to do with http
* libaws is created and includes only the classes that deal with the aws protocols (sns, sqs and s3)
* libvfs_handlers is created is created and includes the fs and uri handlers previously found in libcloud, including the s3 ones


jeremybarnes:
+1 once tests pass or they are verified as being spurious fails

I would eventually like to move vfs_handlers under the vfs subdirectory (this is possible now by forward referencing libraries) but it's not worth blocking things on this branch for that.

jeremybarnes:
I would probably squash merge this one

wsourdeau:
Thanks! Squash merged

FinchPowers:
It seems you squash merged although tests were failing.

FinchPowers:
J'ai reverté le squash commit de master pour que le build + tests passent. @wsourdeau si tu peux réparer cette branche et re merge squash ensuite. :)

wsourdeau:
Fixed bug and reverted the revert.

jeremybarnes:
@wsourdeau the +1 was contingent on the tests passing (or being identified as suprious fails).  I would have expected to see some kind of acknowledgement as to how they were spurious before merging; note that the Jenkins environment should have all tests passing even if your local environment doesn't.

In the future, we should make sure that tests really are passing before we merge.

PR #500, opened 2016-07-07
Fixed problem with path functions not keeping prefix (MLDB-1781)

jeremybarnes:


jeremybarnes:
Test failures are unrelated to this change.

sql/expression_value.cc
```diff
@@ -1780,13 +1780,17 @@ ExpressionValue(RowValue row) noexcept
                        RowValue::iterator last,
                        size_t level)
         {
+            cerr << "doing level " << level << " from "
```
mailletf: comment

jeremybarnes: Right... removed.

mailletf:
+1 based on the test

PR #501, opened 2016-07-07
Made fixed tests expected pass (MLDB-1630)

jeremybarnes:
These tests no longer need to be expected failures, as they now pass.


jeremybarnes:
Test failures are spurious

mailletf:
+1

PR #502, opened 2016-07-07
Improved error messages for unclosed strings and identifiers (MLDB-1601)

jeremybarnes:
Instead of saying "Expected UTF-8 character" and pointing to the last character, we will now say that the string was unclosed and point to the first character which is way better for debugging.


jeremybarnes:
Test failures are spurious

steadtler:
+1

PR #503, opened 2016-07-08
MLDB-1603 non printable chars json

jeremybarnes:
Allows all ASCII control characters to be represented in JSON, as per the JSON spec.


FinchPowers:
+1

PR #508, opened 2016-07-10
[MLDBFB-618] Calling {row|column}PathElement() with an index bigger t…

mailletf:
…han the length of the path will simply return NULL instead of throwing


sql/binding_contexts.cc
```diff
@@ -589,15 +589,13 @@ getDatasetDerivedFunction(const Utf8String & tableName,
...
                         firstElement = asPath.size() + firstElement;
-                    if (firstElement < 0 || firstElement >= asPath.size()) {
-                        throw HttpReturnException
-                            (400, "Couldn't extract element '"
-                             + to_string(firstElement) + "' of path '"
-                             + asPath.toUtf8String() + "' with length "
-                             + to_string(asPath.size()));
-                    }
+
+                    if (firstElement < 0 || firstElement >= asPath.size())
```
jeremybarnes: We could consider allowing negative numbers to index from the end as in Python.

jeremybarnes:
It needs some kind of mention in the documentation.

sql/binding_contexts.cc
mailletf: Adding a test for it and it already works

mailletf:
tests qui fail ont pas rapport

FinchPowers:
+1, reroule ton test, Jenkins devrait être réparé.

PR #510, opened 2016-07-11
Mldb 1787 ignore errors js parse

mailletf:
We will investigate replacing this flag with a try() function as part of a different ticket


FinchPowers:
Faut mettre à jour le doc je crois? (+1 pour le code)

PR #511, opened 2016-07-11
MLDB-1793 try builtin fct

FinchPowers:


sql/sql_expression_operations.cc
```diff
@@ -2256,21 +2256,48 @@ bindBuiltinFunction(SqlBindingScope & scope,
...
+                std::vector<ExpressionValue> evaluatedArgs;
+                evaluatedArgs.reserve(boundArgs.size());
+                try {
+                    evaluatedArgs.emplace_back(std::move(
+                        boundArgs[0](row, fn.filter)));
+                }
+                catch (const ML::Exception & exc) {
+                    evaluatedArgs.emplace_back(std::move(
+                        ExpressionValue::exception(exc.what(),
+                                                    Date::now())));
```
FinchPowers: Le message de l'exception est conservé ici, mais ignoré ensuite par le builtin try. On peut le garder si on veut.

sql/sql_expression_operations.cc
```diff
@@ -2256,21 +2256,48 @@ bindBuiltinFunction(SqlBindingScope & scope,
...
-                        evaluatedArgs.emplace_back(std::move(a(row, fn.filter)));
+            {
+                ExcAssert(args.size() == 2);
+                std::vector<ExpressionValue> evaluatedArgs;
+                evaluatedArgs.reserve(boundArgs.size());
+                try {
+                    evaluatedArgs.emplace_back(std::move(
+                        boundArgs[0](row, fn.filter)));
+                }
+                catch (const ML::Exception & exc) {
```
jeremybarnes: The HttpReturnException has a lot more information within it, which we should extract here.  Or at least use ML::getExceptionString().

sql/sql_expression_operations.cc
```diff
@@ -2256,21 +2256,48 @@ bindBuiltinFunction(SqlBindingScope & scope,
                 this,
                 fn.resultInfo};
     }
-    else {
+
+    if (functionName == "try") {
```
jeremybarnes: This is the wrong place to put it.  We shouldn't have logic hardcoded into the generic function wrapper here.

jeremybarnes:
I'm not convinced that a new atom type for exceptions is right.  We will then need to define the semantics of operations on them.  Why not make it much simpler, with the try function as a pure builtin that trys and catches within its lambda?

sql/sql_expression_operations.cc
FinchPowers: Étant donné que la call stack ne passe pas par les fonctions tant que les arguments ne sont pas prêts, je c'était le plus simple à faire. Ça pourrait être l'exception qui confirme la règle...

Autrement, as-tu une approche à me suggérer?

sql/sql_expression_operations.cc
FinchPowers: If the thrown exception was not of type HttpReturnException, will it still be caught?

FinchPowers:
Replaced by https://github.com/mldbai/mldb/pull/525

PR #517, opened 2016-07-12
MLDBFB-239: general preliminary cleanup of S3Api

wsourdeau:
* Removed methods "upload", "uploadRecursive", "download", "downloadToFile"
  as they were not strict s3 primitives and duplicated functionalities
  already found in the s3 vfs handler
* S3Api::Content replaced with HttpRequestContent
* Removed s3_transfer_cmd, which was making use of the above and roughly
  duplicates s3cp


mailletf:
let's wait until after today's release

steadtler:
+1 after today's release.

PR #519, opened 2016-07-12
CUDA build now working (MLDB-1786)

jeremybarnes:
Fixes a few bits and pieces that allow CUDA support to be built for Tensorflow.

It's a no-op in the official build as it requires the WITH_CUDA variable to be set, which isn't by default.

Note that CUDNN isn't currently supported.


FinchPowers:
+1

PR #521, opened 2016-07-12
MLDB-1797 url round tripping + import text filename fix

FinchPowers:


jeremybarnes:
Looks fine, but there are lots of places (everywhere we have a filter_istream or filter_ostream) that should be updated to use the new method.
 If those are going to be in a separate PR, then +1 once tests pass or failures are confirmed spurious.

FinchPowers:
The failing test is the spurious one.

PR #522, opened 2016-07-13
Mldbfb 239 http client s3 handlers

wsourdeau:
This PR modifies the S3 downloader and uploader classes by making use of the asynchronous methods offered by S3Api, rather than using multiple threads along with the blocking methods.
This finalizes MLDBFB-239 and brings the actual benefits expected from it, which is to reduce the number of context switches and thereby enhancing the scalability of the s3 subsystem.

### DIFF
--- a/vfs_handlers/s3_handlers.cc
+++ b/vfs_handlers/s3_handlers.cc
@@ -1,8 +1,7 @@
-// This file is part of MLDB. Copyright 2015 Datacratic. All rights reserved.
-
 /** s3_handlers.cc
     Jeremy Barnes, 3 July 2012
     Wolfgang Sourdeau, 7 July 2012
+    This file is part of MLDB.
     Copyright (c) 2012, 2016 Datacratic.  All rights reserved.
 
     S3 VFS handlers
@@ -20,6 +19,7 @@
 #include "mldb/types/url.h"
 #include "mldb/vfs/filter_streams_registry.h"
 #include "mldb/vfs/fs_utils.h"
+#include "mldb/vfs_handlers/exception_ptr.h"
 #include "mldb/soa/service/s3.h"
 
 using namespace std;
@@ -122,6 +122,10 @@ struct AtInit {
 } atInit;
 
 
+/****************************************************************************/
+/* S3 DOWNLOADER                                                            */
+/****************************************************************************/
+
 size_t getTotalSystemMemory()
 {
     long pages = sysconf(_SC_PHYS_PAGES);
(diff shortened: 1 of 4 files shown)


jml/utils/exception_ptr.cc
```diff
@@ -0,0 +1,51 @@
+/* exception_ptr.cc
```
jeremybarnes: Should not be in jml; we are attempting to get rid of that.

jml/utils/exception_ptr.cc
```diff
@@ -0,0 +1,51 @@
+/** exception_ptr.cc
```
jeremybarnes: Should not be in jml; we are trying to get rid of that.

vfs_handlers/s3_handlers.cc
```diff
@@ -122,292 +122,416 @@ struct AtInit {
...
+            maxRqs = 15;
+        if (fileInfo.size > 256 * 1024 * 1024)
+            maxRqs = 30;
+        chunks.resize(maxRqs);
+
+        /* Hack to ensure that the file's last modified time is earlier than 1
+           seconds before the start time of the download. Useful for the etag
+           check below. Should seldom be executed in practice. */
+        Date now = Date::now();
+        if (fileInfo.lastModified > now.plusSeconds(-1)) {
```
jeremybarnes: I don't like this at all.  Why do we need it here?

vfs_handlers/s3_handlers.cc
```diff
@@ -122,292 +122,416 @@ struct AtInit {
...
 
-        size_t getChunkSize(unsigned int chunkNbr)
-            const
-        {
-            size_t chunkSize = std::min(baseChunkSize * (1 << (chunkNbr / 2)),
-                                        maxChunkSize);
-            return chunkSize;
+    void handleEtagChange()
+    {
+        auto newInfo = api->getObjectInfo(bucket, resource.substr(1));
```
jeremybarnes: Again, I'm not sure what this is trying to do here.  If the etag is different, the file is different irrespective of the last modified time.

vfs_handlers/s3_handlers.cc
```diff
@@ -122,292 +122,416 @@ struct AtInit {
...
+    std::string bucket;
+    std::string resource;
+    S3Api::ObjectInfo fileInfo;
+    uint64_t offset; /* the lower position in the file from which the download
+                      * is started */
+    uint64_t downloadSize; /* total number of bytes to download */
+    size_t baseChunkSize;
+    size_t maxChunkSize;
+
+    bool closed; /* whether close() was invoked */
```
jeremybarnes: should this be atomic?

vfs_handlers/s3_handlers.cc
```diff
@@ -425,306 +549,266 @@ makeStreamingDownload(const std::string & bucket,
...
-        impl->chunkSize = 8 * 1024 * 1024;  // start with 8MB and ramp up
+/****************************************************************************/
+/* S3 UPLOADER                                                              */
+/****************************************************************************/
 
-        impl->start();
-    }
+inline void touchByte(const char * c)
+{
+    __asm__(" # [in]":: [in] "r" (*c):);
```
jeremybarnes: Comment is probably appropriate here.

vfs_handlers/s3_handlers.cc
```diff
@@ -425,306 +549,266 @@ makeStreamingDownload(const std::string & bucket,
...
+          metadata(objectMetadata),
+          onException(excCallback),
+          closed(false),
+          chunkSize(8 * 1024 * 1024), // start with 8MB and ramp up
+          currentRq(0),
+          activeRqs(0)
+    {
+        /* Maximum chunk size is what we can do in 3 seconds, up to 1% of
+           system memory. */
+#if 0
```
jeremybarnes: Why is this commented out?

vfs_handlers/s3_handlers.cc
```diff
@@ -425,306 +549,266 @@ makeStreamingDownload(const std::string & bucket,
...
-        size_t chunkIndex;
-        atomic<bool> shutdown;
-        std::vector<std::thread> tg;
-
-        Date startDate;
-
-        struct Chunk {
-            Chunk() : data(nullptr)
-            {
+        catch (...) {
```
jeremybarnes: JML_CATCH_ALL

jeremybarnes:
Generally good, but lots of little comments.

vfs_handlers/s3_handlers.cc
wsourdeau: Originally, it was a hack designed to ensure that the reason why the etag could differ was the expected reason: that the file had indeed changed during the download. I think this hack can be removed but it does not cause much harm.

vfs_handlers/s3_handlers.cc
wsourdeau: Not necessary because it is only set and read from the client thread.

vfs_handlers/s3_handlers.cc
wsourdeau: I cannot remember, possibly a left over from a test. However, I wonder to what extend this is useful to use larger chunk sizes. Considering the limit of more or less 450mb/s we reach under specific conditions, for a hypothetical max chunk size of 50mbyte, that would represent 9 reqs/second and the overhead in that case is negligible.

vfs_handlers/s3_handlers.cc
wsourdeau: Even if I am rethrowing?

wsourdeau:
@jeremybarnes Thanks! I have addressed most of your comments. I answered or asked questions for the rest of them.

vfs_handlers/s3_handlers.cc
jeremybarnes: We wouldn't want the onException function to run on thread cancellation, so I'd say in this case yes, even if rethrowing.

jeremybarnes:
+1 once comments are addressed

vfs_handlers/s3_handlers.cc
wsourdeau: As long as the writes are executed in a separate threads, the callback hack is needed in order to prevent incomplete files from being "commited" to s3. Thread cancellation does not change that. However, if your concern is that another exception could be thrown from within the callback itself, then I agree and I propose to call it from it's own try/catch block.

wsourdeau: Nvm, I trust there are reasons behind this hack that I am not yet aware of (and would liek to find out). I am going to make that change and investigate later.

PR #523, opened 2016-07-13
Use highwayhash for hash functions (MLDB-1798)

jeremybarnes:
Improves hashing speed and removes a redundant copy of siphash that was in our system.


FinchPowers:
+1

PR #525, opened 2016-07-13
MLDB-1793 try builtin fct alternate impl

FinchPowers:
@jeremybarnes Reimplementation as discussed.
@mailletf We should figure out a way to put the exception message in the catch part. Maybe if in `try(expr1, expr2)` `expr2` is a NULL? But then it would be impossible to have NULL as the error value. We could also have a new construct. A new function that takes no args maybe? Something like `try(expr1, err_msg())` where `err_msg()` is, guess what, the error message.


FinchPowers:
Ok, I have an even simpler idea for the error message part. If you have no second argument, the error message is returned.

sql/sql_expression.h
```diff
@@ -332,17 +336,31 @@ struct BoundFunction {
     {
     }
 
+    // Use this ctor to have a BoundFunction that will override the call to
+    // bindBuiltinFunction. It can hence hence have the control flow when the
+    // expressions within are called.
+    BoundFunction(BindBuiltinFunction bindBuiltinFunction,
```
jeremybarnes: Should be called BindFunction not BindBuiltinFunction

sql/sql_expression_operations.cc
```diff
@@ -2214,7 +2214,7 @@ bind(SqlBindingScope & scope) const
     BoundFunction fn = scope.doGetFunction(tableName, functionName,
                                            boundArgs, scope);
     
-    if (!fn) {
+    if (!fn && !fn.bindBuiltinFunction) {
```
jeremybarnes: If fn is false, we should throw anyway.

sql/builtin_functions.cc
```diff
@@ -3141,6 +3141,35 @@ BoundFunction hash(const std::vector<BoundSqlExpression> & args)
 
 static RegisterBuiltin registerHashFunction(hash, "hash");
 
+BoundFunction tryFct(const std::vector<BoundSqlExpression> & args)
+{
+    checkArgsSize(args.size(), 2);
```
jeremybarnes: You can do a separate implementation for if we have one or two args based upon args.size() here.

sql/builtin_functions.cc
```diff
@@ -3141,6 +3141,35 @@ BoundFunction hash(const std::vector<BoundSqlExpression> & args)
...
+    {
+        return {[=] (const SqlRowScope & row,
+                     ExpressionValue & storage,
+                     const VariableFilter & filter) -> const ExpressionValue &
+        {
+            ExcAssertEqual(boundArgs.size(), 2);
+            try {
+                return storage = std::move(boundArgs[0](row, GET_LATEST));
+            }
+            catch (const HttpReturnException & exc) {
```
jeremybarnes: We should use JML_CATCH_ALL if we don't care about the type of the exception, or at least std::exception.

jeremybarnes:
Much better, a few details to solve

sql/sql_expression_operations.cc
FinchPowers: If I throw if fn is false then it will never reach the bindFunction call a couples of lines lower.

FinchPowers:
Rereview svp.

jeremybarnes:
+1

PR #529, opened 2016-07-15
Added test for already fixed rowPath() with new executor (MLDB-1810)

jeremybarnes:
Adds a testcase that was left behind in previous changes that fixed the underlying problem.


FinchPowers:
+1

PR #531, opened 2016-07-15
Refactored dataset scope (MLDB-1811)

jeremybarnes:
Refactoring towards allowing ExpressionValue to be passed into a Dataset Scope to remove the dependency on MatrixNamedRow et al.

### DIFF
--- a/server/dataset_context.cc
+++ b/server/dataset_context.cc
@@ -133,6 +133,147 @@ doGetColumn(const Utf8String & tableName,
 /* ROW EXPRESSION DATASET CONTEXT                                            */
 /*****************************************************************************/
 
+RowName
+SqlExpressionDatasetScope::RowScope::
+getRowName() const
+{
+    if (rowName) return *rowName;
+    else {
+        ExcAssert(row);
+        return row->rowName;
+    }
+}
+
+const RowName &
+SqlExpressionDatasetScope::RowScope::
+getRowName(RowName & storage) const
+{
+    if (rowName) return *rowName;
+    else return row->rowName;
+}
+
+RowHash
+SqlExpressionDatasetScope::RowScope::
+getRowHash() const
+{
+    if (rowName) return *rowName;
+    else return row->rowHash;
+}
+
+const ExpressionValue &
+SqlExpressionDatasetScope::RowScope::
+getColumn(const ColumnName & columnName,
+          const VariableFilter & filter,
+          ExpressionValue & storage,
+          ssize_t knownOffset) const
+{
+    if (expr) {
+        const ExpressionValue * val
+            = expr->tryGetNestedColumn(columnName, storage, filter);
+        if (!val)
+            return storage = ExpressionValue::null(Date::negativeInfinity());
+        else return *val;
+    }
+    else {
+        const ExpressionValue * fromOutput
+            = searchRow(row->columns, columnName, filter, storage);
+        if (fromOutput)
+            return *fromOutput;
+        
+        return storage = std::move(ExpressionValue::null(Date::negativeInfinity()));
+    }
+}
+
+ExpressionValue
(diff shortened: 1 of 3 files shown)


server/dataset_context.cc
```diff
@@ -133,6 +133,130 @@ doGetColumn(const Utf8String & tableName,
...
+        
+        for (auto & c: row->columns) {
+            columns.insert(std::get<0>(c));
+            ts.setMax(std::get<2>(c));
+        }
+
+        return ExpressionValue(columns.size(), ts);
+    }
+    else {
+        throw HttpReturnException(600, "getColumnCount() from ExpressionValue");
```
FinchPowers: status 600?

server/dataset_context.cc
```diff
@@ -133,6 +133,130 @@ doGetColumn(const Utf8String & tableName,
...
+ExpressionValue
+SqlExpressionDatasetScope::RowScope::
+getFilteredValue(const VariableFilter & filter) const
+{
+    if (row) {
+        // TODO: if one day we are able to prove that this is
+        // the only expression that touches the row, we could
+        // move it into place
+        ExpressionValue val(row->columns);
+        return val.getFilteredDestructive(filter);
```
FinchPowers: If we do and it's not the case, aren't test going to fail?

FinchPowers:
Invalid http status to fix.
+1 pending tests success.

steadtler:
All the extra branches in tight look might affect performance, im curious by how much. @FinchPowers you still keep an eye on the performance dashboard?

FinchPowers:
Yes sir.

server/dataset_context.cc
jeremybarnes: Right you are.  600 is my code for "not finished, don't merge this".  That code path is not used right now, but it should be fixed for sure.

jeremybarnes:
The branch predictor should be able to do a pretty good job there; that's why I did it that way instead of with virtual methods.  If there is a performance impact it's probably due to instruction cache pressure more than anything.

server/dataset_context.cc
```diff
@@ -133,6 +133,130 @@ doGetColumn(const Utf8String & tableName,
...
+            auto it = index.find(std::get<0>(c));
+            if (it == index.end()) {
+                continue;
+            }
+            
+            result.emplace_back(it->second, std::get<1>(c), std::get<2>(c));
+        }
+    }
+    else {
+        auto onAtom = [&] (const Path & columnName,
```
wsourdeau: "onAtom" is unused.

server/dataset_context.h
```diff
@@ -78,12 +79,47 @@ struct SqlExpressionDatasetScope: public SqlExpressionMldbScope {
...
+                 const BoundParameters * params = nullptr)
+            : row(&row), rowName(nullptr), expr(nullptr), params(params)
         {
         }
 
-        const MatrixNamedRow & row;
+        RowScope(const RowName & rowName,
+                 const ExpressionValue & row,
+                 const BoundParameters * params = nullptr)
+            : row(nullptr), rowName(&rowName), expr(&row), params(params)
```
wsourdeau: I think "row", as function parameter, being passed to "expr" is confusing, especially when "row" as member var is set to something else.

server/dataset_context.cc
jeremybarnes: We would need much stronger static analysis in MLDB to prove that something is moveable.  For the moment, it's const and we shouldn't move.

server/dataset_context.cc
jeremybarnes: Fixed it; nice catch.  Thanks.

server/dataset_context.h
jeremybarnes: Updated.

PR #532, opened 2016-07-15
MLDBFB-638 fill orberby and having aggregators for transform dataset

steadtler:


testing/MLDBFB-638-groupby-orderby-transform.py
```diff
@@ -1,5 +1,5 @@
-# MLDB-1794-join-subselect.py
-# Mathieu Marquis Bolduc, 2016-12-07
+# MLDBFB638-groupby-orderby.py
+# Mathieu Marquis Bolduc, 2016-14-07
```
FinchPowers: iso-8601 this is not

FinchPowers: Alternativement tu peux prendre un format plus verbeux
ex.: 14 July 2016

FinchPowers:
+1

PR #533, opened 2016-07-15
MLDBFB-636 fix join dataset rowhash to work like rowName and fix orde…

steadtler:
…r by a rowhash of a sub dataset

### DIFF
--- a/builtin/joined_dataset.cc
+++ b/builtin/joined_dataset.cc
@@ -58,13 +58,7 @@ struct JoinedDataset::Itl
         RowName rowName;   ///< Name of joined row
         RowName leftName, rightName;  ///< Names of joined rows from input datasets
         //compact_vector<RowHash, 2> rowHashes;   ///< Row hash from input datasets
-    };
-
-    enum JoinSide {
-        JOIN_SIDE_LEFT = 0,
-        JOIN_SIDE_RIGHT,
-        JOIN_SIDE_MAX
-    };
+    };   
 
     struct JoinedRowStream : public RowStream {
 
@@ -817,7 +811,14 @@ struct JoinedDataset::Itl
 
         int64_t index = iter->second;
         const RowEntry& entry = rows[index];
+
         return JOIN_SIDE_LEFT == side ? entry.leftName : entry.rightName;
+    };
+
+    RowHash getSubRowHash(const RowName & name, JoinSide side) const
+    {
+        RowName subName = getSubRowName(name, side);
+        return RowHash(subName);      
     };
 
     //Query the original row name down the tree of joined datasets on that side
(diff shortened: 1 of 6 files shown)


builtin/joined_dataset.cc
```diff
@@ -973,53 +985,120 @@ getChildAliases(std::vector<Utf8String> & outAliases) const
...
+                     const SqlRowScope & scope)
+                {
+                    auto & row = scope.as<SqlExpressionDatasetScope::RowScope>();
+                    return ExpressionValue(itl->getSubRowHash(row.row.rowName, tableSide), Date::negativeInfinity());
+                },
+                std::make_shared<Uint64ValueInfo>()
+            };
+    }
+    else {
+        ExcAssert(functionName == "rowPath");
```
jeremybarnes: We shouldn't assert here. Let it fall through, return a null function at the end, and then the generic error handling code further up the call stack will print a nice error message.

builtin/joined_dataset.cc
```diff
@@ -973,53 +985,120 @@ getChildAliases(std::vector<Utf8String> & outAliases) const
...
+                     const SqlRowScope & scope)
+                {
+                    auto & row = scope.as<SqlExpressionDatasetScope::RowScope>();
+                    return ExpressionValue(itl->getSubRowHashFromChildTable(tableName, row.row.rowName, tableSide), Date::negativeInfinity());
+                },
+                std::make_shared<Uint64ValueInfo>()
+            };
+    }
+    else{
+        ExcAssert(functionName == "rowPath");
```
jeremybarnes: Same here

builtin/joined_dataset.cc
```diff
@@ -973,53 +985,120 @@ getChildAliases(std::vector<Utf8String> & outAliases) const
...
+    }
+}
 
-        if (itl->isChildTable(tableName, JoinedDataset::Itl::JOIN_SIDE_LEFT))
-            tableSide = JoinedDataset::Itl::JOIN_SIDE_LEFT;
-        else if (itl->isChildTable(tableName, JoinedDataset::Itl::JOIN_SIDE_RIGHT))
-            tableSide = JoinedDataset::Itl::JOIN_SIDE_RIGHT;
+BoundFunction
+JoinedDataset::
+overrideFunctionFromChild(JoinSide tableSide,
```
jeremybarnes: Isn't this duplicated?

builtin/joined_dataset.h
```diff
@@ -16,6 +16,11 @@
 namespace Datacratic {
 namespace MLDB {
 
+ enum JoinSide {
```
jeremybarnes: Maybe this should go in join_utils.h?

jeremybarnes:
Minor things.  +1 once they are addressed.

builtin/joined_dataset.cc
steadtler: But this is an obvious programmer error... if someone calls this method with a function it doesnt support, its because a programmer explictly made it so. The error should be caught before even the change is merged, and the assert here will make it *much* easier to find the source of the problem?

PR #534, opened 2016-07-15
Store timestamps as numbers to improve compression options in Tabular (MLDB-1815)

jeremybarnes:
Since we have code to compress doubles but not timestamps, this allows us to take advantage of that code to do a better job at storing the timestamp column too.


FinchPowers:
+1

PR #535, opened 2016-07-15
Allow for incremental queries (MLDB-1816)

jeremybarnes:
Provides an interface that can be used to receive results from an analytics query one at a time.  Also fixes constness issues in the interface, and returns the customary bool for a short-circuited exit.

NOTE: the queryStructuredIncremental function is not tested yet.  There is a coming commit that uses it to implement tested code.

### DIFF
--- a/server/analytics.cc
+++ b/server/analytics.cc
@@ -36,7 +36,7 @@ namespace MLDB {
 /** Equivalent to SELECT (select) FROM (dataset) WHEN (when) WHERE (where), and each matching
     row is passed to the aggregator.
 */
-void iterateDataset(const SelectExpression & select,
+bool iterateDataset(const SelectExpression & select,
                     const Dataset & from,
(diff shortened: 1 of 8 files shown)


server/analytics.cc
```diff
@@ -390,6 +391,93 @@ queryFromStatement(SelectStatement & stm,
...
+    else if (table.table.runQuery && stm.from) {
+
+        auto getParamInfo = [&] (const Utf8String & paramName)
+            -> std::shared_ptr<ExpressionValueInfo>
+            {
+                throw HttpReturnException(500, "No query parameter " + paramName);
+            };
+        
+        if (!params)
+            params = [] (const Utf8String & param) -> ExpressionValue { throw HttpReturnException(500, "No query parameter " + param); };
```
FinchPowers: wrap line

FinchPowers:
+1

PR #536, opened 2016-07-17
Refactored uses of ExpressionValue::getStructured() and removed (MLDB-1818)

jeremybarnes:
Removes the getStructured() method from ExpressionValue, which should basically never be used and couples the semantics of the ExpressionValue class to its internal implementation.

The main uses of the method were in parsing options to functions, which were refactored to use the ValueDescription system (which also allows for generated documentation).  As part of those changes, a new `addAuto()` function was added to `StructureValueDescription` to allow default values to be automatically inferred, and some validation errors were fixed in the JSON parsing code.

The optimized path for aggregators on a static set of columns was also fixed to no longer use getStructured; this should avoid several situations where using embeddings with aggregators would have caused issues.

Finally, a couple of tests which were confused between arrays and structures were fixed.


FinchPowers:
From what I understand I already reviewed it in https://github.com/mldbai/mldb/pull/537 .

jeremybarnes:
+1 from #537

PR #537, opened 2016-07-18
MLDB-1813 matrix named row removal

jeremybarnes:
This simplifies and streamlines the query logic by removing the use of MatrixNamedRow (an exploded format) in query processing and sticking to ExpressionValues.

It has an effect on efficiency: in cases where there is SQL processing, it will be faster (as no converting back and forth from ExpressionValues is needed).  However, for a select * it will be somewhat slower.

There is a forthcoming patch to fix that situation by making the ExpressionValue be used through the entire query processing path.

Note that it includes the MLDB-1818 diff, as it's built on top of that code.  Once that branch is merged, this can be rebased and will end up substantially smaller.

### DIFF
--- a/server/bound_queries.cc
+++ b/server/bound_queries.cc
@@ -50,13 +50,48 @@ struct BoundSelectQuery::Executor {
                          bool processInParallel,
                          ssize_t offset,
                          ssize_t limit,
-                         std::function<bool (const Json::Value &)> onProgress) = 0;
+                         std::function<bool (const Json::Value &)> onProgress)
+    {
+        auto newProcessor = [&] (Path & rowName,
+                                 ExpressionValue & val,
+                                 std::vector<ExpressionValue> & calcd,
+                                 int bucketNum)
+            {
+                NamedRowValue output;
+                output.rowName = std::move(rowName);
+                output.rowHash = output.rowName;
+                val.appendToRowDestructive(Path(), output.columns);
+                return processor(output, calcd, bucketNum);
+            };
+
+        return executeExpr(newProcessor, processInParallel, offset, limit, onProgress);
+    }
+
+    virtual bool executeExpr(std::function<bool (Path & rowName,
+                                                 ExpressionValue & val,
+                                                 std::vector<ExpressionValue> & calcd,
+                                                 int bucketNum)> processor,
+                             bool processInParallel,
+                             ssize_t offset,
+                             ssize_t limit,
+                             std::function<bool (const Json::Value &)> onProgress)
+    {
+        auto newProcessor = [&] (NamedRowValue & output,
+                                 std::vector<ExpressionValue> & calcd,
+                                 int bucketNum)
+            {
+                ExpressionValue val(std::move(output.columns));
+                return processor(output.rowName, val, calcd, bucketNum);
+            };
+
+        return execute(newProcessor, processInParallel, offset, limit, onProgress);
+    }
 
     virtual std::shared_ptr<ExpressionValueInfo> getOutputInfo() const = 0;
 };
 
 struct UnorderedExecutor: public BoundSelectQuery::Executor {
-    std::shared_ptr<MatrixView> matrix;
+    const Dataset & dataset;
     GenerateRowsWhereFunction whereGenerator;
     SqlExpressionDatasetScope & context;
     BoundSqlExpression whereBound;
@@ -68,7 +103,7 @@ struct UnorderedExecutor: public BoundSelectQuery::Executor {
                                              std::vector<ExpressionValue> & calcd,
                                              int rowNum)> ExecutorAggregator;
 
-    UnorderedExecutor(std::shared_ptr<MatrixView> matrix,
+    UnorderedExecutor(const Dataset & dataset,
                       GenerateRowsWhereFunction whereGenerator,
                       SqlExpressionDatasetScope & context,
                       BoundWhenExpression whenBound,
@@ -76,7 +111,7 @@ struct UnorderedExecutor: public BoundSelectQuery::Executor {
                       std::vector<BoundSqlExpression> boundCalc,
                       OrderByExpression newOrderBy,
                       int numBuckets)
-        : matrix(std::move(matrix)),
+        : dataset(dataset),
           whereGenerator(std::move(whereGenerator)),
           context(context),
           whenBound(std::move(whenBound)),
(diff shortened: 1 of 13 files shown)


server/bound_queries.cc
```diff
@@ -50,13 +50,48 @@ struct BoundSelectQuery::Executor {
...
+                         std::function<bool (const Json::Value &)> onProgress)
+    {
+        auto newProcessor = [&] (Path & rowName,
+                                 ExpressionValue & val,
+                                 std::vector<ExpressionValue> & calcd,
+                                 int bucketNum)
+            {
+                NamedRowValue output;
+                output.rowName = std::move(rowName);
+                output.rowHash = output.rowName;
```
FinchPowers: double check: hash = name?

server/bound_queries.cc
```diff
@@ -282,43 +324,77 @@ struct UnorderedExecutor: public BoundSelectQuery::Executor {
         return parallelMapHaltable(0, effectiveNumBucket, doBucket);
     }
 
-    std::tuple<NamedRowValue, std::vector<ExpressionValue> >
-    processRow(MatrixNamedRow& row,
+#if 0
```
FinchPowers: Ce code sert à quoi? Débugger?

server/bound_queries.cc
```diff
@@ -408,17 +484,19 @@ struct OrderedExecutor: public BoundSelectQuery::Executor {
...
 
                 //where already checked in whereGenerator
 
                 whenBound.filterInPlace(row, rowContext);
 
                 NamedRowValue outputRow;
-                outputRow.rowName = row.rowName;
-                outputRow.rowHash = row.rowName;
+                outputRow.rowName = rows[rowNum];
+                outputRow.rowHash = rows[rowNum];
```
FinchPowers: Double check: hash = name?

server/bound_queries.cc
```diff
@@ -1101,6 +1178,56 @@ execute(std::function<bool (NamedRowValue & output,
...
+{
+    //STACK_PROFILE(BoundSelectQuery);
+
+    ExcAssert(processor);
+
+    try {
+        return executor->executeExpr(processor, processInParallel,
+                                     offset, limit, onProgress);
+    } JML_CATCH_ALL {
+        rethrowHttpException(-1, "Execution error: "
```
FinchPowers: -1 -> 500 je crois

sql/builtin_functions.cc
```diff
@@ -1777,6 +1778,30 @@ BoundFunction norm(const std::vector<BoundSqlExpression> & args)
...
+DEFINE_STRUCTURE_DESCRIPTION(ParseJsonOptions);
+
+ParseJsonOptionsDescription::
+ParseJsonOptionsDescription()
+{
+    addAuto("ignoreErrors", &ParseJsonOptions::ignoreErrors,
+            "If true, errors in the JSON are ignored and the element with "
+            "an error will be silently ignored.  If false (the default), "
+            "a JSON format error will lead to the function failing with "
+            "an exception.");
```
FinchPowers: La fonction `try` a été créée pour ça. Je ne crois pas qu'on devrait avoir l'équivalent en paramètre de fonction.

sql/builtin_functions.cc
```diff
@@ -1800,55 +1825,32 @@ BoundFunction parse_json(const std::vector<BoundSqlExpression> & args)
...
                         parseJson(parser, val.getEffectiveTimestamp(),
-                                  encode);
+                                  options.arrays);
                 }
                 catch(std::exception & e) {
-                    if(ignoreErrors) {
+                    if(options.ignoreErrors) {
                         RowValue rv;
                         rv.emplace_back(make_tuple(Path("__parse_json_error__"), CellValue(true), val.getEffectiveTimestamp()));
                         return ExpressionValue(std::move(rv));
```
FinchPowers: Il manque le else throw il me semble.

sql/expression_value.cc
```diff
@@ -4431,6 +4431,21 @@ extractJson() const
     return result;
 }
 
+void
+ExpressionValue::
+extractImpl(void * obj, const ValueDescription & desc) const
+{
+    // TODO: don't go through JSON for simple atoms and structures
```
FinchPowers: Do we have a ticket # for that?

sql/sql_expression.cc
```diff
@@ -3609,6 +3610,50 @@ parse(ML::Parse_Context & context, bool allowUtf8) {
...
+    if (val.isEmbedding() || val.isAtom() || val.empty()) {
+        // Don't deconstruct an embedding or an atom; either pass or not
+        // based upon the timestamp (there is only a single timestamp)
+        return passValue(val.getEffectiveTimestamp());
+    }
+
+    StructValue kept;
+    kept.reserve(val.rowLength());
+
+    // TODO: we should do two passes to first mark and then extract
```
FinchPowers: TODO: ticket #

FinchPowers:
Summary:
* double checks on rowName = rowHash
* create tickets for TODOs
* `try` vs `ignoreError` in ignoreErrors
* status -1 -> status 500

+1 for the code

server/bound_queries.cc
jeremybarnes: It's OK.  There is a conversion operator; it's a supported way of filling in a row hash when only a row name is available.

server/bound_queries.cc
jeremybarnes: will remove

server/bound_queries.cc
jeremybarnes: Same; it's OK (and doesn't change the semantics of the code)

server/bound_queries.cc
jeremybarnes: -1 means keep the existing code.  I've made it clearer in the code.

sql/builtin_functions.cc
jeremybarnes: In that case, you should adjust the DOD for the try function to include replacing this functionality, and include it there.  This is just a refactoring that keeps the same options as before; modifying the interface requires coordination with customers and should be handled in a separate ticket.

sql/expression_value.cc
jeremybarnes: MLDB-1820; added

sql/sql_expression.cc
jeremybarnes: I don't think this needs a ticket number, but it does merit a TODO as if ever it causes performance issues we have an idea how to fix it.

sql/builtin_functions.cc
FinchPowers: @mailletf À ta discrétion: tu peux ouvrir un billet pour retirer cette fonctionnalité d'ici maintenant que nous avons le `try`.

sql/builtin_functions.cc
jeremybarnes: it's there

PR #538, opened 2016-07-18
Reduced value_description.h inclusions in headers (MLDB-1822)

jeremybarnes:
Improves compile time (marginally) and helps keep our interfaces uncluttered.


FinchPowers:
+1

PR #542, opened 2016-07-18
MLDB-1827 fix NULL propagation for base math operators

FinchPowers:


sql/sql_expression_operations.cc
```diff
@@ -1171,8 +1171,12 @@ static CellValue binaryModulus(const CellValue & la, const CellValue & ra)
 struct BinaryPlusOp {
     static CellValue apply(const CellValue & l, const CellValue & r)
     {
-        if (l.empty() || r.empty())
+        if (l.empty()) {
             return l;
+        }
+        if (r.empty()) {
+            return r;
+        }
```
FinchPowers: @jeremybarnes Any timestamp logic to take into account here? Should it return the max ts between l and r?

jeremybarnes: Yes

jeremybarnes: What all of these should be is:

```
if (l.empty() || r.empty()) {
    return ExpressionValue::null(calcTs(l, r));
}
```

FinchPowers: I updated the test and there was nothing to do after all. Ts are not part of cell values.

jeremybarnes: In that case, the logic is more convoluted than necessary.  We can just do if (l.empty() or r.empty()) return CellValue();

jeremybarnes:
Consider simplifying the null logic.  +1 after that.

PR #547, opened 2016-07-19
Fixed SELECT in column expression (MLDBFB-646)

jeremybarnes:
Allows the `SELECT` part of `COLUMN EXPR (...)` to be used.

@simlmx FYI
@jpilaul would be worth adding to the tutorial on COLUMN EXPR

```
select column expr(select {a: value() * 10, b: value() * 20} ) named 'res' from (select x:1, y:2)
```

```
[
         [ "x.a", 10, "-Inf" ],
         [ "x.b", 20, "-Inf" ],
         [ "y.a", 20, "-Inf" ],
         [ "y.b", 40, "-Inf" ]
      ]
```

### DIFF
--- a/sql/binding_contexts.cc
+++ b/sql/binding_contexts.cc
@@ -186,6 +186,21 @@ doGetFunction(const Utf8String & tableName,
                 std::make_shared<PathValueInfo>()};
     }
 
+    if (functionName == "value") {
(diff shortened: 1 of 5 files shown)


FinchPowers:
Wow. +1.

PR #548, opened 2016-07-20
MLDB-1829 jseval rows

jeremybarnes:
- Fixes crashes with JS contexts being mixed within jseval
- Adds an ExpressionValue type to JS that mirrors the real one
- Allows for arguments to be passed through in non-simplified form in jseval by adding a '!' to the variable names parameter, so that we can get full fidelity passing through jseval


jeremybarnes:
Test failure is runner_test, which is completely unrelated and known spurious.

container_files/public_html/doc/builtin/lang/Javascript.md
```diff
@@ -196,6 +196,29 @@ MLDB's atomic types are represented in Javascript as follows:
...
+- `at()` will return a new ExpressionValue with the timestamp
+  modified to happen at the given point in time.  This can be
+  used to put timestamps back on values which have been processed
+  with Javascript code.
+- `columns()` returns an object with the column names as keys and
+  their ExpressionValue values as values.  Note that this method
+  only un-nests by one level.
+
+ExpressionValue objects are primarily used by the `jseval` function
+with simplified arguments off.
```
guyd: Consider adding a link to the description of _simplified arguments_.

container_files/public_html/doc/builtin/sql/ValueExpression.md
```diff
@@ -649,13 +649,21 @@ The SQL function `jseval` allows for the inline definition of functions using Ja
     any SQL expressions and will be bound to the parameters passed in to the
     function.
 
+There are two ways that values in arguments can be represented in Javascript:
+simplified (the default), and a non-simplified representation that is accessed
+by adding a `!` character to the parameter name argument (for example, `!x,y`
+instead of `x,y`).
```
FinchPowers: Ouf. Mes soucis se froncent. Utiliser le "!" ce qui veut généralement dire "not"? Je ne dis pas que c'est mauvais, je me demande si on est certain que c'est la bonne chose à faire.

container_files/public_html/doc/builtin/sql/ValueExpression.md
```diff
@@ -695,3 +703,16 @@ log to the console to aid debugging. Documentation for this object can be found
...
+If the first character of the argument string is `!`, then non-simplified
+arguments are used.  These are harder to work with in Javascript, but allow
+for the entire set of values in MLDB to be represented, especially structured
+values or those with repeated columns or multiple timestamps per value.
+
+For example, the following query yields the same as `(SELECT x:1, y:2)`,
+in other words it doesn't mess around with the values:
+
+```sql
+SELECT jseval('return row;', '!row', {*}) AS * FROM (SELECT x:1, y:2)"
```
guyd: This example seems artificial.  If it is, I would consider adding an example that is more realistic (e.g. the use case that forces us to add the non-simplified argument).

container_files/public_html/doc/builtin/sql/ValueExpression.md
FinchPowers: Une alternative serait un nom de fonction différent. `jseval<something>`.

plugins/lang/js/mldb_js.cc
```diff
@@ -79,6 +79,137 @@ registerMe()
...
+registerMe()
+{
+    using namespace v8;
+
+    HandleScope scope;
+
+    auto fntmpl = CreateFunctionTemplate("ExpressionValue");
+    auto objtmpl = fntmpl->PrototypeTemplate();
+
+    objtmpl->Set(String::New("toJs"), FunctionTemplate::New(Methods::toJs));
```
guyd: Might be more forgiving for the user to register also the function `tojs` (without a capital j) since this is the only function name containing a capital.

container_files/public_html/doc/builtin/sql/ValueExpression.md
simlmx: Moi aussi j'ai sourcillé en voyant le "!"...

guyd:
I reviewed briefly the code with only minor comments.  I don't no the js code that well.

I agree with @FinchPowers that the exclamation mark might be confusing.

FinchPowers:
J'ai fait une expérience.

Dataset
```
{'rowName' : 'row1', 'columns' : [['x', 'data', 0]]
```

Requête
```
SELECT jseval('return row;', 'row', {*}) FROM ds
```

Résultat
```
2016-07-20 09:48:09.426 javascript [
   {
      "columns" : [
         [ "jseval('return row;', 'row', {*}).0.0.0", "x", "NaD" ],
         [ "jseval('return row;', 'row', {*}).0.1", "data", "NaD" ],
         [
            "jseval('return row;', 'row', {*}).0.2",
            {
               "ts" : "1970-01-01T00:00:00Z"
            },
            "NaD"
         ]
      ],
      "rowHash" : "86065feec3521acc",
      "rowName" : "row1"
   }
]
```

Il me semble que chaque colonne est explosée en 3, soit le nom, la valeur et la date. Est-ce bien ce que l'on souhaite?

Aussi, je pense que le nom de la première colonne est incorrect.  Il devrait se terminer par `).0.0` et non pas par `).0.0.0`
```
"jseval('return row;', 'row', {*}).0.0.0"
"jseval('return row;', 'row', {*}).0.1"
"jseval('return row;', 'row', {*}).0.2"
```

container_files/public_html/doc/builtin/sql/ValueExpression.md
jeremybarnes: The goal was to avoid breaking existing uses of jseval, without cluttering up the syntax.  Should we use `jseval2`?  Or `jsexec`?  Any other thoughts?  Because it's more complex to write a js function with ExpressionValues as inputs, I didn't want to make this the default way of doing things.

container_files/public_html/doc/builtin/sql/ValueExpression.md
jeremybarnes: It's an artificial example, but it's also that only way to return a real structure from jseval.  We can add a better use-case.

jeremybarnes:
@FinchPowers that behaviour is exactly the same as master; there is no change to the behaviour without an exclamation point.  Structures are unfortunately broken out as [ [ key, [ value, ts ] ], ... ] and this behaviour is maintained to avoid breaking existing jseval scripts.  If you use `'!row'` instead of `'row'`, you get much more reasonable behaviour, but I didn't want to break existing code and so you need to opt in using the '!'.

container_files/public_html/doc/builtin/sql/ValueExpression.md
simlmx: Why not just add an optional parameter?
jseval('return row;', 'row', {*}, fancyStructuredThingy=true)
(with right syntax...)

FinchPowers:
Merci pour l'explication.

container_files/public_html/doc/builtin/sql/ValueExpression.md
FinchPowers: The js way:
```
jseval('return row;', 'row', {*}, {fancyStructuredThingy :true})
```

Je pense que c'est la meilleure solution jusqu'à maintenant.

jeremybarnes: That doesn't work because it is possible to pass more than one argument to jseval

FinchPowers: That's right... From the doc

> As many argument as are listed in part 2, in the same order. These can be any SQL expressions and will be bound to the parameters passed in to the function.

Then we are back to the alternate function name.

PR #549, opened 2016-07-20
Mldbfb 634 join structured column

steadtler:
Added python converter for rows of rows

Normal executor:

Needed to check both the simplified prefix and the initial prefix because normal tables dont return the table name, but in joins we do return the name of the child table as part of the column name.

Pipeline executor:

Needed to pass the table name in the getallcolumns (wildcard) so that the join knows which of its child table is concerned

Needed to evaluate each atom by the wildcard filter


plugins/lang/python/mldb_python_converters.h
```diff
@@ -122,6 +120,71 @@ struct CellValueConverter
...
+    static void construct(PyObject* obj_ptr, void* storage)
+    {
+        auto pyToPathElement = [] (PyObject* ptr)
+        {
+            if (PyInt_Check(ptr)) {
+                int val = boost::python::extract<int>(ptr);
+                return PathElement(std::to_string(val));
+            }
+            else if (PyFloat_Check(ptr)) {
+                double val = boost::python::extract<double>(ptr);
```
FinchPowers: Bien vu. ;)

mailletf: :P

sql/execution_pipeline_impl.cc
```diff
@@ -105,47 +106,35 @@ doGetAllColumns(std::function<ColumnName (const ColumnName &)> keep,
...
 
-            const ExpressionValue & rowContents
-            = row.values.at(fieldOffset + ROW_CONTENTS);
+    auto exec = [=] (const SqlRowScope & rowScope, const VariableFilter & filter) -> ExpressionValue
+    {
+        auto & row = rowScope.as<PipelineResults>();
 
-            RowValue result;
+        const ExpressionValue & rowContents
+        = row.values.at(fieldOffset + ROW_CONTENTS);
```
FinchPowers: indent

sql/execution_pipeline_impl.cc
```diff
@@ -617,61 +607,82 @@ doGetColumn(const ColumnName & columnName, int fieldOffset)
...
         rightPrefix = right->as();
-    
-    auto leftOutput = left->doGetAllColumns(keep, leftFieldOffset(fieldOffset));
-    auto rightOutput = right->doGetAllColumns(keep, rightFieldOffset(fieldOffset));
+
+    bool useLeft = tableName.empty() || tableName == leftPrefix;
+    bool useRight = tableName.empty() || tableName == rightPrefix;
+
+    if (useLeft)
+        cerr << "use left" << endl;
```
FinchPowers: use the logging interface

FinchPowers:
cerr à enlever / remplacer et +1

PR #552, opened 2016-07-20
Fixed tabular dataset (MLDB-1837)

jeremybarnes:
Fixes an issue with the initAt() function over Tabular datasets whereby it could end up at an invalid iterator position if it started off between chunks.

One character fix.


FinchPowers:
No test case?

jeremybarnes:
Test case is external, and it's a clear boundary condition bug.  Probably not worth the effort for a test case.

PR #555, opened 2016-07-21
MLDB-1834 cleaner err msg for SELECT {*}

FinchPowers:


server/dataset_context.cc
```diff
@@ -129,6 +129,14 @@ doGetColumn(const Utf8String & tableName,
     throw HttpReturnException(400, "Cannot read column \"" + columnName.toUtf8String() + "\" with no dataset.");
 }
 
+GetAllColumnsOutput
+SqlExpressionMldbScope::
+doGetAllColumns(const Utf8String & tableName,
+                function<ColumnName (const ColumnName &)> keep)
+{
+    throw HttpReturnException(400, "Cannot get all columns with no dataset.");
```
jeremybarnes: It should say "cannot use wildcards with no from clause".

steadtler: Note that this scope is used in many contexts, some of which a dataset is present. Also, doGetAllColumns is used for all wildcards, e.g.: x.prefix*, not just "all columns"

jeremybarnes: Also, the doGetColumn below should say "cannot read column "<columnName>" with no from clause"

mailletf:
+1

PR #556, opened 2016-07-21
Mldbfb 650 named aggregators

steadtler:
Allow use of aggregators in NAMED expressions
Fix use of aggregators in HAVING expressions

This required a small refactoring to have a smart ptr on the having and named expressions.


simlmx:
J'ai pas tout compris mais ça a l'air bien joli.
+1 de mon coté.

sql/sql_expression.h
```diff
@@ -1053,14 +1053,6 @@ struct SqlExpression: public std::enable_shared_from_this<SqlExpression> {
     */
     virtual bool isIdentitySelect(SqlExpressionDatasetScope & context) const;
 
-    /** Find any children that is an aggregator call
```
guyd: Any reason why we deleted that comment?

PR #560, opened 2016-07-22
Reduced Path allocations constructing from PathElement (MLDB-1845)

jeremybarnes:
Simple fix to avoid using a std::string as a temporary.


guyd:
+1

PR #561, opened 2016-07-22
[MLDB-1844] Tokenize arguments are now camel case

mailletf:


sql/tokenize.cc
```diff
@@ -38,28 +36,52 @@ TokenizeOptionsDescription()
...
             "return bigrams and trigrams, where tokens are joined by "
             "underscores. For example, "
-            "`tokenize('Good day world', {splitchars:' ', ngram_range:[2,3]})`"
+            "`tokenize('Good day world', {splitChars:' ', ngramRange:[2,3]})`"
             "will return the row `{'Good_day': 1, 'Good_day_world': 1, 'day_world': 1}`");
+
+    onUnknownField = [] (TokenizeOptions * options,
+                         JsonParsingContext & context)
+    {
+        if(context.fieldName() == "min_token_length") {
```
jeremybarnes: Aren't we better to warn the user and then accept the argument under the old name?

sql/tokenize.cc
```diff
@@ -38,28 +36,52 @@ TokenizeOptionsDescription()
...
+            context.exception("The 'splitchars' argument has been "
+                    "renamed to 'splitChars'");
+        }
+        else if(context.fieldName() == "quotechar") {
+            context.exception("The 'quotechar' argument has been "
+                    "renamed to 'quoteChar'");
+        }
+        else {
+            context.exception("Unknown field '" + context.fieldName()
+                    + " parsing import.text configuration");
```
guyd: import.text -> Tokenize option

guyd:
+1 once the tests pass

sql/tokenize.cc
mailletf: We don't really have a mechanism to warn them .. ? I could silently accept the params but apart from adding something to the returned payload, which would probably not be seen, I'm not sure what we can do?

PR #562, opened 2016-07-22
Handle dynamic columns in new executor (MLDB-1843)

jeremybarnes:
Fixes problems with dynamic schemas in the new executor's table expressions.


guyd:
+1

PR #563, opened 2016-07-22
Don't create unnecessary paths on forEachAtom (MLDB-1846)

jeremybarnes:
Small optimization to reduce allocations in a very heavily used codepath.


guyd:
+1

PR #565, opened 2016-07-22
More fixes for types and new executor (MLDB-1843)

jeremybarnes:
Fixes a typo causing problems with structured selects in the new executor, and a type system bug that was marking an object with fixed fields but variable subfields as having no schema variability.


steadtler:
+1

PR #573, opened 2016-07-28
Mldb 1865 dist tables rt update

mailletf:


plugins/dist_table_procedure.cc
```diff
@@ -553,6 +601,81 @@ DistTableFunction(MldbServer * owner,
...
+}
+
+void
+DistTableFunction::
+persist(const Url & modelFileUrl) const
+{
+    // get exclusive access
+    DistTablesMap copiedMap;
+    {
+        boost::unique_lock<boost::shared_mutex> uniqueLock(_access);
```
EricRobert: should be a shared_lock here

FinchPowers:
+1

PR #576, opened 2016-07-29
Mldb 1831 slow join

guyd:
This is fixing the sub-optimal implementation of equijoin in the pipeline executor.  The execution in the pipeline executor is now comparable with the execution in the standard executor (from the benchmark it is about twice slower but considering that the pipeline executor is not multithreaded I believe this is pretty much the best we can do.).

The implementation is getting a little hard to follow and it would probably benefit from some refactoring.


sql/execution_pipeline_impl.cc
```diff
@@ -1132,41 +1124,33 @@ take()
...
-            return l;
-        };
-
+            
         if (lField == rField) {
             // Got a row!
             //cerr << "*** got row match on " << jsonEncode(lField) << endl;
 
+            // return a copy since we are buffering the original left value
+            auto result = shared_ptr<PipelineResults>(new PipelineResults(**l));
```
steadtler: use make_shared?

sql/execution_pipeline_impl.cc
```diff
@@ -1176,14 +1160,16 @@ take()
...
+                l = takeFromBuffer(l);
                 continue;
             }
 
-            l = left->take();
-            if (l) {
-                ExpressionValue nextLField =  l->values.back().getColumn(0, GET_ALL);
+            l = takeFromBuffer(l);
+
+            bool updateFirstDup = false;
```
steadtler: I guess Dup is for duplicate? Its ok to use the full word...

steadtler:
+1 after minor comments

PR #578, opened 2016-07-29
Test cleanup

guyd:


guyd:
Merging right-away since I got a previous +1.

PR #580, opened 2016-08-01
MLDB-1868 path performance regressions

jeremybarnes:
A set of micro-optimizations that take the columnCount() portion of the Reddit benchmark from around 4 seconds to around 1.5 seconds, removing most of the performance loss associated with moving to ExpressionValues.

Primarily this involves reducing the amount of sorting, now that the ExpressionValue itself is always sorted internally, as well as optimizations for less copying in the Transposed dataset, increasing the set of conditions under which memcmp() can be used in path comparisons, and avoiding temporary values when parsing paths in the sparse matrix dataset.

### DIFF
--- a/builtin/transposed_dataset.cc
+++ b/builtin/transposed_dataset.cc
@@ -108,6 +108,11 @@ struct TransposedDataset::Itl
         return col;
     }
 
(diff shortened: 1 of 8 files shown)


steadtler:
+1

PR #586, opened 2016-08-03
Better flattening/unflattening of paths (MLDBFB-666)

jeremybarnes:
Sorts out some of the confusion around using `COLUMN EXPR (AS parse_path(columnName())` by adding `flatten_path` and `unflatten_path` functions that do what would be expected of them and can be used as a matched pair.

Also adds more efficient implementations of the `columnPathElement()` and `columnPathLength()` functions, and fixes an issue where the `AS` expression was applied twice in some dynamic `COLUMN EXPR` contexts.

### DIFF
--- a/container_files/public_html/doc/builtin/sql/ValueExpression.md
+++ b/container_files/public_html/doc/builtin/sql/ValueExpression.md
@@ -332,6 +332,18 @@ will change values on each row under consideration. See the [Intro to Datasets](
   which may be used for example as the result of a `NAMED` clause.  This is the
   inverse of `stringify_path` (above).
 - `path_element(path, n)` will return element `n` of the given `path`.
+- `path_length(path)` will return the number of elements in the given `path`.
+- `flatten_path(path)` will return a path with a single element that encodes
+  the entire `path` passed in, in the same manner as `stringify_path`.  This
+  is useful where a series of nested values need to be turned into a flat set
+  of columns for another function or a vector aggregator.  By using
+  `COLUMN EXPR (AS flatten_path(columnPath()))` an entire object can be
+  flattened in this manner.
+- `unflatten_path(path)` is the inverse of `flatten_path`.  It requires that
+  the input path have a single element, and will turn it back into a variable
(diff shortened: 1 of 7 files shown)


sql/binding_contexts.cc
```diff
@@ -186,7 +189,44 @@ doGetFunction(const Utf8String & tableName,
                 std::make_shared<PathValueInfo>()};
     }
 
+    if (functionName == "columnPathElement") {
```
steadtler: I would add a test for those two in the pipeline executor. At least to be sure wheter it works or not. +1 after that.

jeremybarnes: That's out of scope of this ticket, but I created MLDB-1874 to deal with it.

container_files/public_html/doc/builtin/sql/ValueExpression.md
```diff
@@ -332,6 +332,18 @@ will change values on each row under consideration. See the [Intro to Datasets](
...
 - `path_element(path, n)` will return element `n` of the given `path`.
+- `path_length(path)` will return the number of elements in the given `path`.
+- `flatten_path(path)` will return a path with a single element that encodes
+  the entire `path` passed in, in the same manner as `stringify_path`.  This
+  is useful where a series of nested values need to be turned into a flat set
+  of columns for another function or a vector aggregator.  By using
+  `COLUMN EXPR (AS flatten_path(columnPath()))` an entire object can be
+  flattened in this manner.
+- `unflatten_path(path)` is the inverse of `flatten_path`.  It requires that
+  the input path have a single element, and will turn it back into a variable
```
guyd: have -> has

container_files/public_html/doc/builtin/sql/ValueExpression.md
```diff
@@ -332,6 +332,18 @@ will change values on each row under consideration. See the [Intro to Datasets](
...
+  sized path.  Using `COLUMN EXPR (AS unflatten_path(columnPath()))` an entire
```
guyd: sized -> size?  not sure I understand what this sentence means otherwise.

PR #587, opened 2016-08-03
MLDB-1873 encoding unknown column

jeremybarnes:
Fixes path ordering problems.

NOTE: the test refers to a non-existant file and is broken, but the fix makes it get further than it was.


guyd:
+1 for the code.

PR #589, opened 2016-08-04
Added eval_sql function (MLDB-1876)

jeremybarnes:
This is an internal function that is used by the lidar plugin.  It allows for functionality to be implemented in SQL and C++ parameters to be bound in seamlessly, enabling things like

```c++
SqlBindingScope scope;
uint8_t alpha = 0;  // set the transparency of the pixel
ExpressionValue singlePixelPng = evalSql(scope, "tf_EncodePng([[[0,0,0,$1]]])", alpha);
```

### DIFF
--- a/sql/eval_sql.cc
+++ b/sql/eval_sql.cc
@@ -1,0 +1,70 @@
+/** eval_sql.cc                                                     -*- C++ -*-
+    Jeremy Barnes, 1 August 2016
+    This file is part of MLDB. Copyright 2016 Datacratic. All rights reserved.
+
+    Lambda function written in SQL.
+*/
+
+#include "eval_sql.h"
+#include "binding_contexts.h"
+
+namespace Datacratic {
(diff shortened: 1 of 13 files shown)


sql/eval_sql.cc
```diff
@@ -0,0 +1,68 @@
+/** eval_sql.h                                                     -*- C++ -*-
+    Jeremy Barnes, 1 August 2016
+    Lambda function written in SQL.
```
steadtler: copyright notice?

sql/eval_sql.h
```diff
@@ -0,0 +1,100 @@
+/** eval_sql.h                                                     -*- C++ -*-
+    Jeremy Barnes, 1 August 2016
```
steadtler: copyright notice?

steadtler:
neat. +1 after minor comments.

PR #590, opened 2016-08-04
Allow inference and use of type attributes for TF builtins (MLDB-1877)

jeremybarnes:
Fixes various embedding type inference and maintenance problems to allow TF functions that don't accept doubles to work properly.

### DIFF
--- a/tensorflow/MLDB-1736-tensorflow-builtins.js
+++ b/tensorflow/MLDB-1736-tensorflow-builtins.js
@@ -49,4 +49,13 @@ var op = mldb.get("/v1/plugins/tensorflow/routes/ops/EncodePng");
 
 mldb.log(op);
 
+var png = mldb.query("SELECT tf_EncodePng([[[0,0,0,0]]], {T: { type: 'DT_UINT8'}}) AS png NAMED 'png'");
+mldb.log(png);
+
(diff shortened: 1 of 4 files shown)


steadtler:
+1

tensorflow/MLDB-1736-tensorflow-builtins.js
```diff
@@ -49,4 +49,13 @@ var op = mldb.get("/v1/plugins/tensorflow/routes/ops/EncodePng");
...
 mldb.log(op);
 
+var png = mldb.query("SELECT tf_EncodePng([[[0,0,0,0]]], {T: { type: 'DT_UINT8'}}) AS png NAMED 'png'");
+mldb.log(png);
+
+var png = mldb.query("SELECT tf_EncodePng([[[1,1,1,0]]], {}) AS png NAMED 'png'");
+mldb.log(png);
+
+var png = mldb.query("SELECT tf_EncodePng([[[1,1,1,0]]]) AS png NAMED 'png'");
+mldb.log(png);
```
FinchPowers: Pas d'assert sur le résultat?

jeremybarnes: It implicitly assets that there is no exception.

PR #592, opened 2016-08-04
Added ValueDescription for std::array (MLDB-1881)

jeremybarnes:
Trivial change to allow std::array to be used in ValueDescriptions.

### DIFF
--- a/types/array_description.h
+++ b/types/array_description.h
@@ -1,0 +1,135 @@
+/** array_description.h                                        -*- C++ -*-
+    Jeremy Barnes, 21 August 2015
+    Copyright (c) 2015 Datacratic Inc.  All rights reserved.
+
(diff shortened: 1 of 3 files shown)


types/array_description.h
```diff
@@ -0,0 +1,135 @@
...
+
+    // Constructor to create a partially-evaluated array description.
+    ArrayDescription(ConstructOnly)
+    {
+    }
+
+    std::shared_ptr<const ValueDescriptionT<T> > inner;
+
+    virtual void parseJsonTyped(std::array<T, N> * val,
+                                JsonParsingContext & context) const JML_OVERRIDE
```
steadtler: Shouldnt we just use override now?

steadtler:
+! with a small question

types/array_description.h
jeremybarnes: MLDB-1882

PR #593, opened 2016-08-04
Allow all JSON-compatible types to pass to ValueFunction (MLDB-1880)

jeremybarnes:
Replaces exception fallback with one that goes through JSON to allow all types to be converted.  This means that the ValueFunction interface is now able to support most types, rather than just ExpressionValues and CellValues.


steadtler:
+1

PR #594, opened 2016-08-05
Replace getallcolumn function callback by a structure similar to boun…

steadtler:
…dexpression thats going to be a bit more flexible

### DIFF
--- a/sql/sql_expression.h
+++ b/sql/sql_expression.h
@@ -495,6 +495,39 @@ struct GetAllColumnsOutput {
     std::shared_ptr<RowValueInfo> info;
 };
 
+/*****************************************************************************/
+/* COLUMN FILTER                                                             */
+/*****************************************************************************/
+
+/** Filter that rejects, accepts and or/modify a columns name.
+*/
+
+struct ColumnFilter {
+    typedef std::function<ColumnName (const ColumnName &)> Exec;
+   
+    ColumnFilter() 
+    {
+
+    }    
+
+    ColumnFilter(Exec exec)
+        : exec(std::move(exec))
+    {
+    }
+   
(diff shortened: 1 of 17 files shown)


sql/sql_expression.h
```diff
@@ -495,6 +495,39 @@ struct GetAllColumnsOutput {
...
+/** Filter that rejects, accepts and or/modify a columns name.
+*/
+
+struct ColumnFilter {
+    typedef std::function<ColumnName (const ColumnName &)> Exec;
+   
+    ColumnFilter() 
+    {
+
+    }
```
FinchPowers: Ça sert à quoi de le déclarer s'il est vide?

FinchPowers:
+1

sql/sql_expression.h
steadtler: Pour ne pas que le compilateur se plaigne que tu n'utilise pas l'autre.

FinchPowers: Ah ben, je ne savais pas qu'il s'en plaindrait... :)

sql/sql_expression.h
```diff
@@ -495,6 +495,39 @@ struct GetAllColumnsOutput {
...
+    operator bool () const { return !!exec; }
+
+    Exec exec;
+
+    ColumnName operator () (const ColumnName & columnName) const
+    {
+        if (exec)
+            return exec(columnName);
+        else
+            return columnName;
```
guyd: Consider creating a static member of type ColumnFilter named _identity_ to do just that.  This way the code is very explicit and we can prevent errors using a ColumnFilter that has wrongly not been initialized.

jeremybarnes:
Should be a const reference to the column filter.

PR #595, opened 2016-08-05
getallcolumn refactor changes as per PR

steadtler:


sql/sql_expression.h
```diff
@@ -504,23 +504,31 @@ struct GetAllColumnsOutput {
...
 
-    ColumnFilter(Exec exec)
-        : exec(std::move(exec))
+    ColumnFilter(Exec exec_)
+        : exec(std::move(exec_))
     {
+        init = true;
+    }
+
+    static ColumnFilter identity() {
```
guyd: I was thinking of something like that:
.h
`static ColumnFilter identity`
.cc
```
ColumnFilter::identity = ColumnFilter(
[](const ColumnName & columnName) {
   return columnName;
});
```
Everywhere else
`ColumnFilter::identity`
And I would have removed the unnecessary default constructor and the init member.

guyd:
Minor comment which is more a manner of taste.
+1

sql/sql_expression.h
steadtler: But one of the reason for this refactor is that I want to know when the filter does nothing...

guyd: Not sure I understand how one way or the other makes it easier to know that the filter does nothing.

PR #596, opened 2016-08-05
generic output ds share their description string

FinchPowers:
Trivial refactoring.

### DIFF
--- a/core/procedure.h
+++ b/core/procedure.h
@@ -33,7 +33,7 @@ struct Procedure;
 
 typedef EntityType<Procedure> ProcedureType;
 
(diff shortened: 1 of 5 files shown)


core/procedure.h
```diff
@@ -33,6 +33,10 @@ struct Procedure;
 
 typedef EntityType<Procedure> ProcedureType;
 
+static std::string GENERIC_OUTPUT_DS_DESC =
```
guyd: constexpr const char * would be a better type I think because the string should be constant and shared.

guyd:
Another option is to have a base class for config with an output dataset.  This way you keep the description close to where it is needed.

guyd:
+1

PR #601, opened 2016-08-10
Mldb 1850 postgresql

steadtler:
1st pass postgresql plugin open for comments. Had to change a few things to be able to handle simple queries in postgresql.dataset without having to implement the RowMatrix and ColumnIndex interfaces.


core/dataset.h
```diff
@@ -562,6 +562,11 @@ struct Dataset: public MldbEntity {
     virtual std::vector<ColumnName>
     getColumnNames(ssize_t offset = 0, ssize_t limit = -1) const;
 
+    /** Return wheter or not all columns names and info are known.
```
guyd: wheter -> whether
columns -> column

postgresql/MLDB-1850-postgresql.py
```diff
@@ -0,0 +1,141 @@
...
+mldb.log("Re-Query")
+res = mldb.query("select * from out")
+
+mldb.log(res)
+
+expected = [["_rowName","a","b","c"],
+            ["row_0","alfalfa",1,3.5],
+            ["row_1","brigade",2,5.7]]
+
+assert res == expected
```
guyd: assertEqual would give better error message

postgresql/MLDB-1850-postgresql.py
```diff
@@ -0,0 +1,141 @@
...
+        'query': 'select * from mytable order by a'
+    }
+})
+
+res = mldb.query("select query_from_postgres()")
+
+mldb.log(res)
+
+expected = [["_rowName","query_from_postgres().a","query_from_postgres().b","query_from_postgres().c"],
+            ["result","alfalfa",1,3.5]]
```
guyd: How come there is only one line returned here?

postgresql/MLDB-1850-postgresql.py
```diff
@@ -0,0 +1,141 @@
...
+
+assert res == expected
```
guyd: assertEqual

postgresql/MLDB-1850-postgresql.py
```diff
@@ -0,0 +1,141 @@
...
+
+res = mldb.query("select b from postgresqldataset")
+
+mldb.log(res)
+
+expected = [["_rowName","b"],
+            ["brigade",2],
+            ["alfalfa",1]]
+
+assert res == expected
```
guyd: assertEqual

postgresql/MLDB-1850-postgresql.py
```diff
@@ -0,0 +1,141 @@
...
+
+res = mldb.query("select * from postgresqldataset")
+
+mldb.log(res)
+
+expected = [["_rowName","a","b","c"],
+            ["brigade","brigade",2,5.7],
+            ["alfalfa","alfalfa",1,3.5]]
+
+assert res == expected
```
guyd: assertEqual

postgresql/MLDB-1850-postgresql.py
```diff
@@ -0,0 +1,141 @@
...
+
+mldb.script.set_return("success")
```
guyd: The file would be easier to read and debug if it was split into individual tests.

postgresql/doc/Postgresql.md
```diff
@@ -0,0 +1,48 @@
+# Postgresql plugin documentation
+
+This plugin allows MLDB to interact with a PostgreSQL database.
+For more information about PostgreSQL: <https://www.postgresql.org/>
+
+## PostgreSQL import procedure
+
+This procedure allows to import data from a PostgreSQL database into 
+a MLDB dataset.
```
guyd: I never know if we should use _a_ or _an_ before an acronym starting with a vowel.

postgresql/doc/Postgresql.md
```diff
@@ -0,0 +1,48 @@
+# Postgresql plugin documentation
```
guyd: I would point somewhere here to the documentation of credentials and give instruction that credentials must be set before using any of these function/procedure.

postgresql/postgresql_plugin.cc
```diff
@@ -0,0 +1,832 @@
+/* postgresql_plugin.cc
+   Mathieu Marquis Bolduc, 16 July 2016
+   Copyright (c) 2015 Datacratic Inc.  All rights reserved.
```
guyd: 2015 -> 2016

postgresql/postgresql_plugin.cc
```diff
@@ -0,0 +1,832 @@
...
+
+#include "mldb/core/plugin.h"
+#include "mldb/core/procedure.h"
+#include "mldb/core/function.h"
+#include "mldb/core/dataset.h"
+#include "mldb/soa/credentials/credentials.h"
+#include "mldb/types/structure_description.h"
+#include "mldb/types/any_impl.h"
+
+#include <postgresql/libpq-fe.h>
```
guyd: Any reason why this is imported as a system header?

postgresql/postgresql_plugin.cc
```diff
@@ -0,0 +1,832 @@
...
+    }
+    else if (postgrestype == 701) {
+        return CellValue(atof(PQgetvalue(res, i, j)));
+    }
+    else {
+        return CellValue(PQgetvalue(res, i, j));
+    }
+}
+
+pg_conn* startPostgresqlConnection(string databaseName, string host, int port)
```
guyd: Why not use const & on the two first parameters?

postgresql/postgresql_plugin.cc
```diff
@@ -0,0 +1,832 @@
...
+    auto creds = getCredential("postgresql", host);
+
+    string connString;
+    connString += "dbname=" + databaseName
+               + " host=" + host
+               + " port=" + std::to_string(port)
+               + " user=" + creds.id
+               + " password=" + creds.secret;
+
+    POSTGRESQL_VERBOSE(cerr << connString << endl;)
```
guyd: Logging credentials is risky.  I would not output the password here.

postgresql/postgresql_plugin.cc
```diff
@@ -0,0 +1,832 @@
...
+
+    auto conn = PQconnectdb(connString.c_str());
+    if (PQstatus(conn) != CONNECTION_OK)
```
guyd: Can we be more specific about the error?  It would be particularly helpful to know if it is a connection problem or a credential one.

postgresql/postgresql_plugin.cc
```diff
@@ -0,0 +1,832 @@
...
+
+DECLARE_STRUCTURE_DESCRIPTION(PostgresqlDatasetConfig);
+DEFINE_STRUCTURE_DESCRIPTION(PostgresqlDatasetConfig);
+
+PostgresqlDatasetConfigDescription::
+PostgresqlDatasetConfigDescription()
+{
+    addField("databaseName", &PostgresqlDatasetConfig::databaseName, "Name of the database to connect to.");
+    addField("port", &PostgresqlDatasetConfig::port, "Port of the database to connect to.", 1234);
+    addField("host", &PostgresqlDatasetConfig::host, "Address of the database to connect to ");
```
guyd: First two descriptions are terminated with a period but the last three.

postgresql/postgresql_plugin.cc
```diff
@@ -0,0 +1,832 @@
...
+    addField("tableName", &PostgresqlDatasetConfig::tableName, "Name of the table to be recorded into");
+    addField("primaryKey", &PostgresqlDatasetConfig::primaryKey, "Primary key used to access progress table");
```
guyd: progress!?
I guess we mean PostgreSQL.  This is the only place in the description that we specify that this is for PostgreSQL.  I think the context is clear enough and we don't need to mention it here.

postgresql/postgresql_plugin.cc
```diff
@@ -0,0 +1,832 @@
...
+    }
+
+    pg_conn* startConnection() const
+    {
+        return startPostgresqlConnection(config_.databaseName, config_.host, config_.port);
+    }
+
+    virtual Any getStatus() const
+    {
+        return string("ok");
```
guyd: I would create an MLDBFB ticket to return a better status.  I feel this is particularly important here since many things can go wrong.

postgresql/postgresql_plugin.cc
```diff
@@ -0,0 +1,832 @@
...
+    }
+
+    virtual void recordRowItl(const RowName & rowName,
+                              const std::vector<std::tuple<ColumnName, CellValue, Date> > & vals)
+    {
```
guyd: Should those method throw?

postgresql/postgresql_plugin.cc
```diff
@@ -0,0 +1,832 @@
...
+    }
+
+    /** Commit changes to the database.  Default is a no-op. */
+    virtual void commit()
+    {
+    }
+
+    virtual std::pair<Date, Date> getTimestampRange() const override
+    {
+        throw HttpReturnException(400, "getTimestampRange : Postgresql dataset is read-only");
```
guyd: I believe the capitalization should be PostgreSQL.

postgresql/postgresql_plugin.cc
```diff
@@ -0,0 +1,832 @@
...
+    }
+};
+
+DECLARE_STRUCTURE_DESCRIPTION(PostgresqlRecorderDatasetConfig);
+DEFINE_STRUCTURE_DESCRIPTION(PostgresqlRecorderDatasetConfig);
+
+PostgresqlRecorderDatasetConfigDescription::
+PostgresqlRecorderDatasetConfigDescription()
+{
+    addField("databaseName", &PostgresqlRecorderDatasetConfig::databaseName, "Name of the database to connect to.");
```
guyd: We can use a base config with the PostgreSQL _coordinates_ and derive from it for the other configs.  This would avoid the string duplications and make maintenance easier.

postgresql/postgresql_plugin.cc
```diff
@@ -0,0 +1,832 @@
...
+        : Dataset(owner)
+    {
+        config_ = config.params.convert<PostgresqlRecorderDatasetConfig>();
+
+        if (config_.createTable) {            
+
+            auto conn = startConnection();
+
+            //Drop table if exist
+            string dropDescription = "DROP TABLE IF EXISTS " + config_.tableName;
```
guyd: Isn't this dangerous.  I would make it configurable in the config and set it to false by default.

postgresql/postgresql_plugin.cc
```diff
@@ -0,0 +1,832 @@
...
+                 string errorMsg(PQresultErrorMessage(res));
+                 PQclear(res);
+                 PQfinish(conn);
+                 throw HttpReturnException(400, "Could not drop PostgreSQL table:  ", errorMsg);
+            }
+
+            POSTGRESQL_VERBOSE(cerr << "table " << config_.tableName << " dropped!" << endl;)
+            PQclear(res);
+
+            string tableDescription = "CREATE TABLE " + config_.tableName + " (" + ")";
```
guyd: Do we have plan to support more of the CREATE TABLE syntax?

postgresql/postgresql_plugin.cc
```diff
@@ -0,0 +1,832 @@
...
+        }
+    }
+
+    void alterColumns(pg_conn* conn, const std::vector<std::tuple<ColumnName, CellValue, Date> > & vals)
+    {
+        for (auto& p : vals) {
+            ColumnName column = std::get<0>(p);
+            if (insertedColumns.count(column) == 0){
+                insertedColumns.insert(column);
+                string alterString = "ALTER TABLE " +
```
guyd: Can we alter a table once data has been stored in it?

postgresql/postgresql_plugin.cc
```diff
@@ -0,0 +1,832 @@
...
+                PQclear(res);
+            }
+        }
+    }
+
+    void insertRowPostgresql(pg_conn* conn, const std::vector<std::tuple<ColumnName, CellValue, Date> > & vals) 
+    {
+        alterColumns(conn, vals);
+
+        string insertString = "INSERT INTO " + config_.tableName + " (";
```
guyd: There is a lot of string manipulation on each insertion, I wonder if we should not have a preallocated buffer to avoid reallocation.  May be in a second pass.

postgresql/postgresql_plugin.cc
```diff
@@ -0,0 +1,832 @@
...
+    }
+
+    /** Commit changes to the database.  Default is a no-op. */
+    virtual void commit() override
+    {
+    }
+
+    virtual std::pair<Date, Date> getTimestampRange() const override
+    {
+        throw HttpReturnException(400, "Postgresql recorder dataset is record-only");
```
guyd: PostgreSQL

postgresql/postgresql_plugin.cc
```diff
@@ -0,0 +1,832 @@
...
+    string databaseName;
+    int port;
+    string host;
+    string postgresqlQuery;
+
+    /// The output dataset.  Rows will be dumped into here via insertRows.
+    PolyConfigT<Dataset> outputDataset;
+
+    PostgresqlImportConfig() {
+        databaseName = "database";
```
guyd: This does not seem a sensible default.  I would keep the empty string.

postgresql/postgresql_plugin.cc
```diff
@@ -0,0 +1,832 @@
...
+
+};
+
+DECLARE_STRUCTURE_DESCRIPTION(PostgresqlImportConfig);
+DEFINE_STRUCTURE_DESCRIPTION(PostgresqlImportConfig);
+
+PostgresqlImportConfigDescription::
+PostgresqlImportConfigDescription()
+{
+    addParent<ProcedureConfig>();
```
guyd: The addParent should be done at the end if I recall correctly so that the runOnCreation flag appear last like with all other procedures.

postgresql/postgresql_plugin.cc
```diff
@@ -0,0 +1,832 @@
...
+
+    addField("databaseName", &PostgresqlImportConfig::databaseName, "Name of the database to connect to.");
+    addField("port", &PostgresqlImportConfig::port, "Port of the database to connect to.", 1234);
+    addField("host", &PostgresqlImportConfig::host, "Address of the database to connect to ");
+    addField("postgresqlQuery", &PostgresqlImportConfig::postgresqlQuery, "Query to run in postgresql to get rows");
+
+    addField("outputDataset", &PostgresqlImportConfig::outputDataset,
+             "Output dataset configuration.  This may refer either to an "
+             "existing dataset, or a fully specified but non-existing dataset "
+             "which will be created by the procedure.", PolyConfigT<Dataset>().withType("sparse.mutable"));
```
guyd: Consider adding validation of the database name here to make sure it is not empty and follow any other constraints forced by PostgreSQL.

postgresql/postgresql_plugin.cc
```diff
@@ -0,0 +1,832 @@
...
+    pg_conn* startConnection(const PostgresqlImportConfig & config_) const
+    {
+        return startPostgresqlConnection(config_.databaseName, config_.host, config_.port);
+    }
+
+    virtual Any getStatus() const override
+    {
+        Json::Value result;
+        result["ok"] = true;
+        return result;
```
guyd: Here also I would add more _flesh_ to the status.

postgresql/postgresql_plugin.cc
```diff
@@ -0,0 +1,832 @@
...
+
+    virtual RunOutput run(const ProcedureRunConfig & run,
+                          const std::function<bool (const Json::Value &)> & onProgress) const override
+    {        
+        RunOutput result;
+
+        auto runProcConf = applyRunConfOverProcConf(procedureConfig, run);
+
+        // Connect to Postgresl database
+        auto conn = startConnection(runProcConf);
```
guyd: Consider using an object that will close the `conn` and one that will clear `res` once they go out of scope.

postgresql/postgresql_plugin.cc
```diff
@@ -0,0 +1,832 @@
...
+struct PostgresqlQueryFunctionConfig
+{
+    string databaseName;
+    int port;
+    string host;
+
+    string query;
+
+    PostgresqlQueryFunctionConfig() {
+        databaseName = "database";
```
guyd: Same here.  Not a sensible default.  Add database name validation in the description class.

postgresql/postgresql_plugin.cc
```diff
@@ -0,0 +1,832 @@
...
+        port = 1234;
```
guyd: Isn't the default PostgreSQL port 5432?  I would make it a constexpr instead of copying the default in every config.

postgresql/postgresql_plugin.cc
```diff
@@ -0,0 +1,832 @@
...
+    {
+        PostgresqlQueryFunctionApplier applier(this, functionConfig);
+        return applier.info;
+    }
+};
+
+static RegisterDatasetType<PostgresqlRecorderDataset, PostgresqlRecorderDatasetConfig>
+regPostgresqlRecorderDataset(postgresqlPackage(),
+                 "postgresql.recorder",
+                 "Dataset type that record to a Postgresql database",
```
guyd: PostgreSQL

server/dataset_context.cc
```diff
@@ -431,64 +431,98 @@ doGetAllColumns(const Utf8String & tableName,
...
-                return ColumnName();
+    bool allWereKept = true;
+    bool noneWereRenamed = true;
+    std::unordered_map<ColumnHash, ColumnName> index;
+    std::vector<KnownColumn> columnsWithInfo;
+    SchemaCompleteness schema = SCHEMA_OPEN;
+
+    if (keep.exec) {
+
+        cerr << "has exec " << endl;
```
guyd: to remove

guyd:
I very much like it.  The code is clear and easy to read.  Added minor comments.

postgresql/postgresql_plugin.cc
jraby: We opted to use libpq from the system packages instead on bundling it.

postgresql/MLDB-1850-postgresql.py
steadtler: Thats the requirements for now, for the function.

postgresql/MLDB-1850-postgresql.py
```diff
@@ -0,0 +1,141 @@
+#
+# MLDB-1850-postgresql.py
+# Datacratic, 2016
```
FinchPowers: C'est sensé être ton nom avec la date exacte. Aussi il manque de structure à ce test. As-tu pensé à utiliser le test generator? Il fait tout ça pour toi.

FinchPowers:
```
%> cd mldb/testing
%> python generate_test.py <nom du test>
```

postgresql/postgresql_plugin.cc
steadtler: Yes, the details of the error are appended to the exception error message.

postgresql/postgresql_plugin.cc
steadtler: Some of them are datasets configs, procedure configs, function configs.. they dont have the same base class.

postgresql/postgresql_plugin.cc
steadtler: Yes but its not terrible efficient. I think we'll go on a per-need basis for the syntax. We could spend all year supporting all of postgresql syntax.

postgresql/postgresql_plugin.cc
steadtler: Im no PostgreSQL expert, but it looks like it

postgresql/postgresql_plugin.cc
steadtler: To paraphrase @mailletf , performance is not a concern for the 1st pass

steadtler:
Fixed most PR comments. Note that the test stays manual and wont be part of the regular test suite because we can't start/stop a postgresql database easily on demand.

guyd:
+1

PR #602, opened 2016-08-10
clang compat

FinchPowers:
Changes required toward having clang compilation succeeds. There are at least 1withstanding error.

```
In file included from ./mldb/ext/tensorflow/tensorflow/core/lib/core/threadpool.cc:19:
In file included from mldb/ext/tensorflow/third_party/eigen3/unsupported/Eigen/CXX11/Tensor:1:
In file included from build/x86_64/include/external/eigen_archive/eigen-eigen-d02e6a705c30/unsupported/Eigen/CXX11/Tensor:57:
In file included from build/x86_64/include/external/eigen_archive/eigen-eigen-d02e6a705c30/unsupported/Eigen/CXX11/ThreadPool:53:
build/x86_64/include/external/eigen_archive/eigen-eigen-d02e6a705c30/unsupported/Eigen/CXX11/src/ThreadPool/EventCount.h:183:10: error: private field 'pad_' is not used [-Werror,-Wunused-private-field]
    char pad_[128];
         ^
```


builtin/merged_dataset.cc
```diff
@@ -214,8 +215,10 @@ struct MergedDataset::Itl
 
         virtual const RowName & rowName(RowName & storage) const
         {
-            uint64_t hash = (*it).first;
-            return source->getRowName(RowHash(hash));
+            throw HttpReturnException(500, "FIXME");
```
FinchPowers: FIXME or remove me.

plugins/sparse_matrix_dataset.cc
```diff
@@ -159,8 +159,10 @@ struct SparseMatrixDataset::Itl
         }
 
         virtual const RowName & rowName(RowName & storage) const {
-            uint64_t i = internalStream->current();
-            return source->getRowNameTrans(RowHash(i), *trans);
+            throw HttpReturnException(500, "FIXME");
```
FinchPowers: FIXME or remove me.

plugins/dist_table_procedure.h
```diff
@@ -185,10 +185,12 @@ struct DistTableProcedure: public Procedure {
...
     DistTableProcedureConfig procConfig;
 
-    static void persist(const Url & modelFileUrl, DistTableMode mode, 
+    static void persist(const Url & modelFileUrl, DistTableMode mode,
                         const DistTablesMap & distTablesMap);
 
-    static const int DIST_TABLE_PERSIST_VERSION = 2;
+    static constexpr int DIST_TABLE_PERSIST_VERSION() {
+        return 2;
+    }
```
FinchPowers: Counter for error
```
~/workspace/mldb/build/x86_64/lib/libmldb_builtin_plugins.so: undefined reference to `Datacratic::MLDB::DistTableProcedure::DIST_TABLE_PERSIST_VERSION'
```

ext/tensorflow.mk
```diff
@@ -248,7 +248,9 @@ ifeq ($(toolchain),gcc5)
 TENSORFLOW_WARNING_FLAGS := -Wno-reorder -Wno-return-type -Wno-overflow -Wno-overloaded-virtual -Wno-parentheses -Wno-maybe-uninitialized -Wno-array-bounds -Wno-unused-function -Wno-unused-variable -Wno-uninitialized -Wno-comment
 endif
 ifeq ($(toolchain),clang)
-TENSORFLOW_WARNING_FLAGS := -Wno-unused-const-variable -Wno-unused-private-field -Wno-tautological-undefined-compare -Wno-missing-braces -Wno-absolute-value -Wno-unused-function -Wno-inconsistent-missing-override -Wno-constant-conversion
+# -Wno-error should not be used but so far it seems like the only way to mute
+# warning: braces around scalar initializer
+TENSORFLOW_WARNING_FLAGS := -Wno-unused-const-variable -Wno-unused-private-field -Wno-tautological-undefined-compare -Wno-missing-braces -Wno-absolute-value -Wno-unused-function -Wno-inconsistent-missing-override -Wno-constant-conversion -Wno-unused-variable -Wno-error
```
FinchPowers: J'ai ajouté -Wno-unused-variable et -Wno-error. Je sais que techniquement -Wno-error écrase tous les -Wno- mais au moins ça laisse une trace des autres quand on voudra les enlever.

plugins/dist_table_procedure.h
jeremybarnes: this can be fixed with constexpr.  You need to actually define static variables once in the .cc file.

plugins/sparse_matrix_dataset.cc
jeremybarnes: what is the error message here?

builtin/merged_dataset.cc
jeremybarnes: What is the erro rmessage here?

jeremybarnes:
-Wno-unused-private-field

    char pad_[128];

ext/tensorflow.mk
jeremybarnes: No, don't add -Wno-error.

plugins/dist_table_procedure.cc
```diff
@@ -505,11 +505,10 @@ persist(const Url & modelFileUrl, DistTableMode mode,
 {
     filter_ostream stream(modelFileUrl);
     ML::DB::Store_Writer store(stream);
-    store << DistTableProcedure::DIST_TABLE_PERSIST_VERSION << mode << distTablesMap;
+    store << DistTableProcedure::DIST_TABLE_PERSIST_VERSION() << mode << distTablesMap;
```
jeremybarnes: Fix this with constexpr in the .h file, not by making it a function.

ext/tensorflow.mk
jeremybarnes: Ah, that's OK for the flags for an external include.

plugins/dist_table_procedure.h
```diff
@@ -185,10 +185,12 @@ struct DistTableProcedure: public Procedure {
 
     DistTableProcedureConfig procConfig;
 
-    static void persist(const Url & modelFileUrl, DistTableMode mode, 
+    static void persist(const Url & modelFileUrl, DistTableMode mode,
                         const DistTablesMap & distTablesMap);
 
-    static const int DIST_TABLE_PERSIST_VERSION = 2;
```
jeremybarnes: You just need static constexpr int DIST_TABLE_PERSIST_VERSION = 2;

For older versions of GCC, sometimes you need enum { DIST_TABLE_PERSIST_VERSION=2 };

plugins/sparse_matrix_dataset.cc
jeremybarnes: What is the error?

server/column_scope.cc
```diff
@@ -215,7 +215,8 @@ runIncrementalT(const std::vector<BoundSqlExpression> & exprs,
                     
                 size_t numRows = endOffset - startOffset;
 
-                Val results[exprs.size()];
```
jeremybarnes: Nope, can't do.  It's too slow and this is very performance sensitive.  Come and see me and I'll show you how to address.

server/column_scope.cc
```diff
@@ -286,7 +287,8 @@ runIncrementalT(const std::vector<BoundSqlExpression> & exprs,
     // Apply the expression to everything
     auto doRow = [&] (size_t first, size_t last)
         {
-            Val results[exprs.size()];
```
jeremybarnes: Same here

server/column_scope.cc
```diff
@@ -286,7 +287,8 @@ runIncrementalT(const std::vector<BoundSqlExpression> & exprs,
...
+            unique_ptr<Val[]> resultsPtr(new Val[exprs.size()]);
+            Val * results = resultsPtr.get();
```
jeremybarnes: Same here

types/compact_vector_description.h
```diff
@@ -28,64 +28,67 @@ struct CompactVectorDescription
     {
     }
 
-    virtual void parseJson(void * val, JsonParsingContext & context) const
+    void parseJson(void * val, JsonParsingContext & context) const override
```
jeremybarnes: keep the virtual

types/compact_vector_description.h
```diff
@@ -28,64 +28,67 @@ struct CompactVectorDescription
...
     {
         compact_vector<T, I, S, Sf, P, A> * val2 = reinterpret_cast<compact_vector<T, I, S, Sf, P, A> *>(val);
         return parseJsonTyped(val2, context);
     }
 
-    virtual void parseJsonTyped(compact_vector<T, I, S, Sf, P, A> * val, JsonParsingContext & context) const
```
jeremybarnes: keep the virtual

types/compact_vector_description.h
```diff
@@ -28,64 +28,67 @@ struct CompactVectorDescription
...
+    void parseJsonTyped(compact_vector<T, I, S, Sf, P, A> * val,
+                        JsonParsingContext & context) const override
     {
         this->parseJsonTypedList(val, context);
     }
 
-    virtual void printJson(const void * val, JsonPrintingContext & context) const
```
jeremybarnes: etc

types/utility_descriptions.h
```diff
@@ -39,11 +39,11 @@ struct BridgedValueDescription: public ValueDescription {
     virtual void * getArrayElement(void * val, uint32_t element) const override;
     virtual const void * getArrayElement(const void * val, uint32_t element) const override;
     virtual const ValueDescription &
-    getArrayElementDescription(const void * val, uint32_t element) const;
+        getArrayElementDescription(const void * val, uint32_t element) const override;
```
jeremybarnes: spacing

types/utility_descriptions.h
```diff
@@ -39,11 +39,11 @@ struct BridgedValueDescription: public ValueDescription {
...
     virtual void setArrayLength(void * val, size_t newLength) const override;
     virtual size_t getTupleLength() const override;
-    virtual std::vector<std::shared_ptr<const ValueDescription> >
-    getTupleElementDescriptions() const override;
+    virtual std::vector<std::shared_ptr<const ValueDescription>>
```
jeremybarnes: Spacing, and the > > is nicer on emacs users

ext/tensorflow.mk
jeremybarnes: OK, on second thoughts it should be OK since it's in an external submodule.

types/compact_vector_description.h
FinchPowers: Why? It's redundant.

> From the Standard: If a virtual member function vf is declared in a class Base and in a class Derived, derived directly or indirectly from Base, a member function vf with the same name, parameter-type-list (8.3.5), cv-qualification, and ref-qualifier (or absence of same) as Base::vf is declared, then Derived::vf is also virtual (whether or not it is so declared) and it overrides Base::vf.

types/utility_descriptions.h
FinchPowers: Oh? I thought it was an old habit from when compilers confused it with `operator >>`

builtin/merged_dataset.cc
FinchPowers: `error: returning reference to local temporary object [-Werror,-Wreturn-stack-address]`

plugins/dist_table_procedure.h
FinchPowers: I already tried the `constexpr int DIST_TABLE_PERSIST_VERSION = 2;` version but it still generates a ton of messages.

The enum version works fine though. 👍

plugins/dist_table_procedure.h
FinchPowers: How do you define it in the .cc file if it's already defined in the .h?

plugins/sparse_matrix_dataset.cc
FinchPowers: `error: returning reference to local temporary object [-Werror,-Wreturn-stack-address]error: returning reference to local temporary object [-Werror,-Wreturn-stack-address]`

server/column_scope.cc
FinchPowers: If we use the placement new technique, the ctor and dtor of `<typename Val> won't be called. Isn't that a risk?

Also, the unique_ptr pattern is also used on line 207 so if we do something I suggest we fix both.

FinchPowers: ...unless we then call uninitialized_fill right after, wrap the "thing" it in a RAII structure that will call delete on all of the allocated elements?

plugins/sparse_matrix_dataset.cc
jeremybarnes: Ah, so this has found a real and serious error.  You need to return storage = source->getRowNameTrans(...)

types/compact_vector_description.h
jeremybarnes: Sure, it's redundant, just like comments and parameter names.  The first is redundant in the parameter names, but which is better?

```
void transferMoney(int,int,int)
void transferMoney(int fromAccountNumber, int toAccountNumber, int amountInCents)
```

For virtual, it is important for a programmer to know which methods are virtual, as any of the following actions could have unexpected and difficult to debug side-effects:
- renaming a method
- changing a signature
- deleting a method
- cleaning up redundant versions of a method

Without the redundant virtual keyword, a programmer needs to find the original .h file and look to see if a method is virtual before modifying it.  Of course they won't, and we will end up with bugs.  With the virtual keyword there, it is clearly documented as being virtual and so extra care needs to be taken.

FinchPowers: I mean it's redundant now that we have the `override` specifier.

FinchPowers: Ok got it.

server/column_scope.cc
jeremybarnes: Placement new does call the constructor, and you need to call the destructor manually.  std::uninitialized_fill basically calls placement new over a range.

server/column_scope.cc
```diff
@@ -145,6 +145,29 @@ static CellValue extractVal(const CellValue & val, CellValue *)
     return val;
 }
 
+namespace {
+    // Made to init and hold a non POD array of dynamic size in a memory safe
+    // way.
+    template<typename T>
+    struct Holder {
+        T * holded;
```
jeremybarnes: held

jeremybarnes: I would do it more like:

```
template<typename T, size_t MAX_STACK_ENTRIES=4096/sizeof(T)>
struct PossiblyDynamicBuffer {
    PossiblyDynamicBuffer(size_t sz)
        : size_(sz)
    {
        if (onStack()) {
            std::uninitialized_fill(stackEntries, stackEntries + size_, T());
        }
        else {
            heapEntries = new T[size_];
        }
    }

    ~PossiblyDynamicBuffer()
    {
        if (onStack()) {
            for (size_t i = 0; i < size_;  ++i)
                stackEntries[i].~T();
        }
        else {
            delete[] heapEntries;
        }
    }

    size_t size_;

    bool onStack() const { return size <= MAX_STACK_ENTRIES; }

    T * data() const { return onStack() ? stackEntries: heapEntries; }
    size_t size() const { return size_; }
    
    union {
        T stackEntries[MAX_STACK_ENTRIES];
        T * heapEntries;
    };
};
```

server/column_scope.cc
```diff
@@ -145,6 +145,52 @@ static CellValue extractVal(const CellValue & val, CellValue *)
     return val;
 }
 
+namespace {
+    // Made to init and hold a non POD array of dynamic size in a memory safe
+    // way.
+    template<typename T, size_t MAX_STACK_ENTRIES=4096/sizeof(T)>
```
guyd: I don't get why you divide by the size of the object here.  This controls the number of entries not the number of bytes, right?

FinchPowers: Exactement. Il faut que le nombre soit statique pour que ça compile, donc ici le nombre statique c'est 4kb / size of T.

guyd: Je pense que je comprends, tu divises par sizeof(T) parce que tu veux que le maximum de bytes par défaut soit 4K et non pas variable avec la grosseur de l'objet.

FinchPowers: Bingo.

guyd: I believe this will cause an overrun if the objects do not fall on the 4K boundary.  Can you try with an object of say 96 bytes?

FinchPowers: Testé. Ça prend le plancher donc c'est valide.

jeremybarnes: the number of objects is passed, not the size

guyd: Yes and it all make sense to me finally! :-)

jeremybarnes:
+1

FinchPowers:
Note: This is compatible with clang 3.6 with libstd not based on gcc 6.1.1

PR #605, opened 2016-08-11
Allow build to start with gcc6 toolchain

jraby:


steadtler:
+1 when tests pass

jeremybarnes:
Should document this in Building.mk

jraby:
@jeremybarnes does that look better?

jraby:
Merged, we can fix the wording later if required.

PR #606, opened 2016-08-12
MLDB-1884 fix null default ts

FinchPowers:


steadtler:
+1 when tests pass

jeremybarnes:
Rationale for it being negative infinity versus notADate()?

FinchPowers:
My current task was limited to making the results consistent with the other "NULL" examples.

```
SELECT 'patate' -> "-Inf"
SELECT NULL LIKE 'abc' -> "-Inf"
```

The current patch makes it so that
```
SELECT NULL -> "-Inf"
```
instead of 1970-...


If we want to make it "not a date" then every where we have -Inf should be updated too and it should be done in a separate task.

jeremybarnes:
OK, I agree that it's better than the previous version.  +1

PR #607, opened 2016-08-12
Fix failure when reruning MLDB-574

FinchPowers:


testing/testing.mk
```diff
@@ -116,7 +116,10 @@ $(eval $(call mldb_unit_test,MLDB-534-svd-function-column-errors.js))
 $(eval $(call mldb_unit_test,MLDB-537-hang-on-put-error.js))
 $(eval $(call mldb_unit_test,MLDB-102-select-formats.js))
 $(eval $(call mldb_unit_test,MLDB-543-column-named-zero.js))
+
+$(shell rm -f tmp/MLDB-574-reddit.sqlite) # MLDB-574 fails if the file is already present
```
guyd: I would put that in the test instead of the build system.

steadtler:
+1 thanks Mich

testing/testing.mk
FinchPowers: À ma conaissance javascript n'a pas accès au filesystem.

PR #609, opened 2016-08-15
Put proper Python paths in error messages (MLDB-1896)

jeremybarnes:
Puts the real filename in stack traces and error messages from Python scripts rather than <string>.  Python's interface for doing this is very strange and so it required rather more logic than should have been necessary.


steadtler:
+1

PR #610, opened 2016-08-15
Compilation and tests pass with gcc6 (MLDB-1895)

jeremybarnes:
make toolchain=gcc6 now compiles and passes all tests


steadtler:
+1

PR #611, opened 2016-08-15
Mldb 1891 case in import

steadtler:
Create a proper ValueInfo for CASE expressions.


sql/expression_value.cc
```diff
@@ -1285,6 +1285,53 @@ getColumn(const PathElement & columnName) const
...
+
+    std::vector<KnownColumn> leftcolumns = left_->getKnownColumns();
+    std::vector<KnownColumn> rightcolumns = right_->getKnownColumns();
+
+    sort(leftcolumns.begin(), leftcolumns.end());
+    sort(rightcolumns.begin(), rightcolumns.end());
+
+    std::vector<KnownColumn> result;
+
+    set_intersection(leftcolumns.begin(),leftcolumns.end(),rightcolumns.begin(),rightcolumns.end(),back_inserter(result));
```
guyd: I wonder why this is the intersection instead of the union.  Do you have a convincing argument why one instead of the other?

steadtler: Yes. The intersection is the set of columns we know for SURE will be selected in the output expression.

sql/expression_value.h
```diff
@@ -1526,6 +1530,60 @@ struct UnknownRowValueInfo: public RowValueInfo {
...
+                                                   Date timestamp)> & write)
+        const;
+
+    virtual bool isCompatible(const ExpressionValue & value) const
+    {
+        return left_->isCompatible(value) && right_->isCompatible(value);
+    }
+
+    virtual SchemaCompleteness getSchemaCompleteness() const {
+        return std::min(left_->getSchemaCompleteness(), right_->getSchemaCompleteness());
```
guyd: This is relying on the order of the enum.  I think 
```
return left_->getSchemaCompleteness() == SCHEMA_CLOSED && right_->getSchemaCompleteness() == SCHEMA_CLOSED ? SCHEMA_CLOSED : SCHEMA_OPEN;
```
is more explicit and less error prone.

sql/expression_value.h
```diff
@@ -1526,6 +1530,60 @@ struct UnknownRowValueInfo: public RowValueInfo {
...
+    }
+
+    virtual SchemaCompleteness getSchemaCompletenessRecursive() const {
+        return std::min(left_->getSchemaCompletenessRecursive(), right_->getSchemaCompletenessRecursive());
```
guyd: same here

sql/expression_value.cc
guyd: Hum.  OK

guyd:
+1  with a minor comment

sql/expression_value.cc
jeremybarnes: That's not true.  It should be the union.

If we know FOR SURE that it's there, then we should set that it's dense.

If it may be there, then we should set that it's not dense.

If either side has dynamic columns (where their names are computed at run-time), then we should say that the schema is open.  Otherwise closed.

jeremybarnes:
-1 until the issue is fixed.
note also that there are the static_type and static_known_columns builtin SQL functions for testing functionality like this.  They should probably be used in the testcase to directly test the output.

steadtler:
@jeremybarnes 

Ok it now does the union and recursively merge the columns.

steadtler:
@jeremybarnes 

and now tested using static_known_columns

FinchPowers:
Est-ce mort?

sql/expression_value.cc
```diff
@@ -1285,6 +1285,86 @@ getColumn(const PathElement & columnName) const
...
 /*****************************************************************************/
+/* OR Expression Value Info                                                  */
+/*****************************************************************************/
+
+/** Expression Value info when we dont know which of two value info we will get 
+    With a Case for Example.
+*/
+
+std::shared_ptr<RowValueInfo> 
+ORExpressionValueInfo::
```
jeremybarnes: Should be Or, not OR

jeremybarnes: And in fact, it's a variant type, so it should be VariantExpressionValueInfo

sql/expression_value.h
```diff
@@ -1526,6 +1530,74 @@ struct UnknownRowValueInfo: public RowValueInfo {
...
+/* OR Expression Value Info                                                  */
+/*****************************************************************************/
+
+/** Expression Value info when we dont know which of two value info we will get 
+    With a Case for Example.
+*/
+
+struct ORExpressionValueInfo: public ExpressionValueInfoT<ExpressionValue> {
+
+    ORExpressionValueInfo(std::shared_ptr<ExpressionValueInfo> left, std::shared_ptr<ExpressionValueInfo> right) {
```
jeremybarnes: We should avoid inlining functions in a class that's included all over the place.

sql/sql_expression_operations.cc
```diff
@@ -2505,14 +2505,22 @@ BoundSqlExpression
...
         boundElse = elseExpr->bind(scope);
+        info = boundElse.info;
+    }
 
     std::vector<std::pair<BoundSqlExpression, BoundSqlExpression> > boundWhen;
 
     for (auto & w: when) {
         boundWhen.emplace_back(w.first->bind(scope), w.second->bind(scope));
+        if (info)
+            info = std::make_shared<ORExpressionValueInfo>(info, boundWhen.back().second.info);
```
jeremybarnes: If left and right are the same type, or if one is a strict subset of the other, we should combine them: for example, `CASE WHEN ... THEN 1 ELSE 0` should return an IntValueInfo, and `CASE WHEN ... THEN 0.5 ELSE 1` should return a float value info.

jeremybarnes:
Nearly there.  To do is one renaming, moving functions out of inline, and some kind of basic functionality to collapse the types when they are the same or one subsumes the other one.

sql/sql_expression_operations.cc
steadtler: I added the minimum and we can add more in a later story. But the very reason we created this info type was to avoid creating and maintaining the whole combinatory tree.

steadtler:
If you have more comments send them in I'll address them afterwards.

PR #613, opened 2016-08-17
remove the randomness in the test

guyd:
On outer join, not all rows contain all the columns.  This is fixing the test but a bigger issue remains  that some of the rows are duplicated.  This is coming as a separate PR.


steadtler:
+1

PR #614, opened 2016-08-17
removed transitory members from AnnotedJoinCondition::Side

guyd:
Trivial simplification of the code to unnecessary AnnotedJoinCondition::Side members.


steadtler:
+1 good stuff.

PR #615, opened 2016-08-17
Mldb 1568 remove dataset query route

steadtler:


guyd:
The small iris dataset does not seem required.
+1 otherwise

PR #616, opened 2016-08-17
Removed ML::atomic_ops (MLDB-1901)

jeremybarnes:
Removes those files and replaces (almost) all remaining uses with std::atomic.  The one place where we still need to use the old functions is in the gc_lock, which needs a 16 byte compare-and-swap.  This is a portability problem and will be replaced with an atomic shared pointer, but in a different ticket.


guyd:
+1

PR #621, opened 2016-08-18
MNIST demo notebook

jpilaul:


guyd:
Overall this is an awesome demo.  There is one error when re-running this cell
screen shot 2016-08-19 at 2 22 26 pm
and I was not sure what's the purpose of the _offset_ field
screen shot 2016-08-19 at 2 28 13 pm

guyd:
When loading the MNIST plugin I got a myriad of errors like these:
screen shot 2016-08-19 at 2 42 41 pm

guyd:
May be the _Let's explore the data_ would be easier to understand if we were describing the dataset a bit and the column 785.

guyd:
I would also try to reduce python code so that user attention is on MLDB code and not boilerplate code.  For example, in this cell:
screen shot 2016-08-19 at 3 06 47 pm

We can directly output MLDB response instead of drilling into it (as I've done).  Also, I am not sure why you are using a vertical_count here instead of the more common `count(*)`.

jpilaul:
This is not the right version of the demo. I will need to resubmit my code. Sorry about that.

guyd:
OK.  Let me know when you have it ready.

jpilaul:
Ok - this is the correct version. I think that you were getting the errors occurred when you loaded the plugin because it needs the files from here:
https://github.com/mldbai/rt-mnist-digit-recogniser

guyd:
Surprisingly the demo plugin does not seem as accurate as the training says it is.  For example, it get digit _7_ almost always wrong.

guyd:
Also, the plugin UI could be more usable.  These are a few things that I felt were upsetting:
- picking from the mnist sample does not clear the drawing box resulting in digit overlays
- the results are show too far from the drawing box so it required scrolling all the time back and forth
- the main usage is draw and submit, therefore the two controls should next to each other.

Note that these are minor and I understand that this is a demo so people will not spent days in front of that screen.  Overall, I found the demo, especially the plugin, very fun and interactive.  Kudos.

jpilaul:
Thanks again for the feedback!
7s are admittedly a little tough. Have you tried centering and resizing on 7s? The issue is that the user drawing will always be a little different from the train set so the "live prediction" may suffer as a result. To solve this issue, I have preprocessed the data and added centering and resizing option.

UI feedback makes a lot of sense - I will improve.

guyd:
Regarding the failed predictions, this seems to happen as often with _7_ sampled from the mnist data.  This is why I thought there was something fishy.

jpilaul:
Thanks for mentioning this. I think that I know what this is. I will submit a quick update on the plug-in to center based on pixel mass instead of box coordinates...

mailletf:
MLDB's classifiers functions  et explain fucntions devraient pas avoir de S.

mailletf:
A pickled version of the dataset is available on the deeplearning.net website http://deeplearning.net/tutorial/gettingstarted.html.
 -> mettre le lien sur le texte comme dans l'autre

mailletf:
"In the original dataset, the labels were seperated." seperated in what? different files?

mailletf:
pour le confusion matrix

```
pd.DataFrame(runResults["confusionMatrix"])\
    .pivot_table(index="actual", columns="predicted", fill_value=0)
```

qui vient d'ici: https://docs.mldb.ai/ipy/notebooks/_tutorials/_latest/Procedures%20and%20Functions%20Tutorial.html

mailletf:
ajouter un "print" à cette ligne? mldb.put('/v1/functions/mnist_explainer', {

mailletf:
"The explain function is called and outputs the weight of each pixels towards the final prediction." (qqchose comme)-> shows how helpful/hurtful each pixel was towards the final prediction. If it is red, it means the model thought it meant something that did not end up being true.

mailletf:
"In the representation of the explain matrix (figure 3), the green pixels help increase the score while the red pixels help decrease the score of the chosen digit." même chose que commentaire précédent.

mailletf:
"check out the on plugins" manque un . à la fin

mailletf:
"      "address": "file:///mldb_data/do_you_knowwwwwzer_buster"" -> changer pour le link github quand on open source le plugin. il faudra décider si on veut nuker l'hsitorique ou non

mailletf:
"/v1/plugins/my_pluginster" -> quelque chose d'autre? :P

mailletf:
Dans le plugin:

- il manque un event javascript quand dessine qui trigger une prédiction quand on lache la souris. genre mouseup ou de quoi.
- les images qu'on peut clicker pour utiliser comme modèle ne sont pas la. as-tu oublié de les comiter? ou ?

mailletf:
+1

PR #622, opened 2016-08-18
MLDB-1735 make ln/log behave like postgresql's

simlmx:
- add support for log(x) and log(b,x) that match postgresql's functions
- make log/ln and sqrt not fail on 0/negative values and return nan/-inf instead
- fix a few typos in the doc
- simplify the numeric-functions test file

### DIFF
--- a/sql/builtin_functions.cc
+++ b/sql/builtin_functions.cc
@@ -778,17 +778,56 @@ registerMod(mod, std::make_shared<IntegerValueInfo>(), "mod");
 double ln(double v)
 {
     if (v <= 0)
-        throw HttpReturnException(400, "the argument of the ln function must"
-                                       " be strictly positive");
+        cerr << "negative value encountered in ln" << endl;
 
     return std::log(v);
 }
 
+// log function consistent with postgresql's
+BoundFunction log(const std::vector<BoundSqlExpression> & args)
+{
+    // log(x) (base 10)
+    if (args.size() == 1) {
+        if (args[0] <= 0)
+            cerr << "non-positive value encountered in log" << endl;
+
+        return {[] (const std::vector<ExpressionValue> & args,
+                    const SqlRowScope & scope) -> ExpressionValue
+                {
+                    ExcAssertEqual(args.size(), 1);
+                    return ExpressionValue(std::log10(args[0].toDouble()),
+                                           args[0].getEffectiveTimestamp());
+                },
+                std::make_shared<Float64ValueInfo>()};
+    // log(base, x)
+    } else if (args.size() == 2) {
+        if (args[0] <= 0 || args[1] <= 0)
+            cerr << "non-positive value encountered in log" << endl;
+
+        return {[] (const std::vector<ExpressionValue> & args,
+                    const SqlRowScope & scope) -> ExpressionValue
+                {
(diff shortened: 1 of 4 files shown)


sql/builtin_functions.cc
```diff
@@ -778,17 +778,64 @@ registerMod(mod, std::make_shared<IntegerValueInfo>(), "mod");
 double ln(double v)
 {
     if (v <= 0)
-        throw HttpReturnException(400, "the argument of the ln function must"
-                                       " be strictly positive");
+        cerr << "negative value encountered in ln" << endl;
```
steadtler: Please remove the cerr, this could happen on every row and kill performance in a very nasty way.

sql/builtin_functions.cc
```diff
@@ -778,17 +778,64 @@ registerMod(mod, std::make_shared<IntegerValueInfo>(), "mod");
...
     return std::log(v);
 }
 
+// log function consistent with postgresql's
+BoundFunction log(const std::vector<BoundSqlExpression> & args)
+{
+    // log(x) (base 10)
+    if (args.size() == 1) {
+        if (args[0] <= 0)
+            cerr << "non-positive value encountered in log" << endl;
```
steadtler: Same

sql/builtin_functions.cc
```diff
@@ -778,17 +778,64 @@ registerMod(mod, std::make_shared<IntegerValueInfo>(), "mod");
...
+                        return ExpressionValue();
+
+                    return ExpressionValue(std::log10(args[0].toDouble()),
+                                           args[0].getEffectiveTimestamp());
+                },
+                std::make_shared<Float64ValueInfo>()};
+    // log(base, x)
+    } else if (args.size() == 2) {
+        if (args[0] <= 0 || args[1] <= 0)
+            cerr << "non-positive value encountered in log" << endl;
```
steadtler: same

sql/builtin_functions.cc
```diff
@@ -778,17 +778,64 @@ registerMod(mod, std::make_shared<IntegerValueInfo>(), "mod");
...
+}
+
+static RegisterBuiltin registerLog(log, "log");
+
 double sqrt(double v)
 {
     if (v < 0)
-        throw HttpReturnException(400, "the argument of the sqrt function must"
-                                       " be non-negative");
+        cerr << "negative value encountered in sqrt" << endl;
```
steadtler: same

testing/MLDB-1336-builtin-checks.py
```diff
@@ -11,9 +11,6 @@ def test_sequence(self):
         with self.assertRaisesRegexp(mldb_wrapper.ResponseException, "Executing builtin function exp: Can't convert value 'a' of type 'ASCII_STRING' to double") as re:
             query = "SELECT exp('a')"
             mldb.query(query)
-        with self.assertRaisesRegexp(mldb_wrapper.ResponseException, "Executing builtin function ln: the argument of the ln function must be strictly positive") as re:
-            query = "SELECT ln(-1)"
```
steadtler: Can we keep the test to check that it DOESNT crash and that it returns a consistent value?

steadtler:
+1 after removing the cerr's.

PR #623, opened 2016-08-18
Various fixes for ARM64 support (MLDB-1903)

jeremybarnes:
These are all code quality improvements.  No effect on the default build, but puts a base in place for arm and arm64 architectures and significantly reduces the number of places where we are tied to the x86_64 architecture.


steadtler:
+1

PR #625, opened 2016-08-18
Implemented plugin macros (MLDB-1904)

jeremybarnes:
Improvements to macros to allow plugins to include installed directories and binaries


jraby:
It reads ok to me.

PR #626, opened 2016-08-19
Fix cuda build

jraby:


jeremybarnes:
Looks OK to me.  We should probably merge MLDB-1904, then rebase this branch on top of it before merging.  +1

PR #628, opened 2016-08-23
MLDB-1906 ts conv2d

jeremybarnes:
Allows the conv2d artifact in Tensorflow to work.

Note that despite the documentation, only the NHWC encoding works, which makes the format of image inputs a little bit funky.  There are other Tensorflow ops that may help to coerce the data into the appropriate format.

Also, despite the documentation the DT_DOUBLE type doesn't work for convolutions.

### DIFF
--- a/tensorflow/tensorflow_plugin.cc
+++ b/tensorflow/tensorflow_plugin.cc
@@ -158,6 +158,52 @@ struct AttrValueDescription: public Datacratic::ValueDescriptionT<AttrValue> {
                 throw HttpReturnException(400, "Unknown JSON AttrValue '" + v.toStringNoNewLine() + "'");
             }
         }
+        else if (context.isArray()) {
+            // Do a list if singly nested
+            Json::Value v = context.expectJson();
+
+            if (v.empty()) {
+                val->mutable_list();
+            }
+            else {
+                switch (v[0].type()) {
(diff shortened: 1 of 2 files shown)


tensorflow/tensorflow_plugin.cc
```diff
@@ -927,6 +973,10 @@ struct TensorflowGraphBase: public Function {
                 break;
             }
         }
+        else if (val.isEmbedding()) {
+            // TODO: infer consistent type for tensor, and convert it
```
mailletf: do we need to throw an exception until this is done?

tensorflow/tensorflow_plugin.cc
```diff
@@ -1849,6 +1886,31 @@ struct TensorflowPlugin: public Plugin {
...
+                        if (it == config.attr.end()) {
+                            // We didn't find the attribute
+                            // Set it to the default value
+                            for (auto & attr: opDef->attr()) {
+                                if (attr.name() == attrName) {
+                                    // found it
+                                    config.inputType = attr.default_value().type();
+                                }
+                            }
+                            // Can't do it
```
mailletf: couple of extra whitespaces prefixing this line /troll

mailletf:
+1

PR #629, opened 2016-08-24
change the error message when encountering a bad line in .cloud_crede…

guyd:
…ntials to not show the line contents which is likely containing credentials.


mailletf:
+1

PR #630, opened 2016-08-24
Mldb 1621 support and test weights usage in classifier.test

simlmx:
- add a general test for classifier.test
   - right now it tests either the output dataset OR the stats returned by the query, whichever let me test that the weights work easily
- make sure it supports weights for all modes (boolean, regression, categorical)
- minor related cleaning and fixes


testing/test_classifier_test_proc.py
```diff
@@ -0,0 +1,312 @@
...
+            'params': {
+                'testingData': "SELECT score, label:{}, weight:{} FROM ds".format(label, weight),
+                'outputDataset': 'out',
+                'mode': mode
+            }
+        }
+
+    def test_boolean_dataset_no_weight(self):
+        res = mldb.post('/v1/procedures', self._get_params('boolean', 'bool_label', '1')).json()
+        # TODO test the output of res
```
guyd: I believe this is already done.

guyd:
Will you update the document accordingly?

plugins/accuracy.cc
```diff
@@ -225,15 +225,15 @@ runBoolean(AccuracyConfig & runAccuracyConf,
         output->commit();
     }
 
-    cerr << "stats are " << endl;
+    // cerr << "stats are " << endl;
 
-    cerr << stats.toJson();
+    // cerr << stats.toJson();
```
mailletf: je laisserais celui la. je m'en sers souvent :P

simlmx: Bof c'est ça qui est retourné ensuite. Utilise ce qui est retourné non? Moi je trouve que ça me pollue mon output!

PR #632, opened 2016-08-24
No longer require 128 bit atomics for gc_lock (MLDB-1909)

jeremybarnes:
This fixes a major portability problem (only x86_64 has 128 bit atomics, whereas basically every interesting platform has 64 bit atomics).

The testing code is pretty comprehensive, but this is a very low-level primitive and so we should keep an eye out for test failures in concurrency.

We now only use std::atomic to implement the functionality; no assembler is required any more.

The changes made were as follows:
1.  The visibleEpoch field was removed, and its value inferred from epoch (saving 32 bits)
2.  The exclusive field was removed, and replaced with a single bit on the "in" counter (saving 32 bits).  Note that this reduces that maximum number of threads that can be simultaneously in a critical section from 65535 to 32767.  It shouldn't be a problem in practice, but it is a limit.
3.  Since we no longer had an exclusive or visibleEpoch field to run a futex over, separate futexes outside of the atomic section were created to handle sleeping and wakeups.  These could in the future be replaced by condition variables with some loss in performance.
4.  std::atomic was used for atomic operations
5.  The code was refactored, to put all of the inline methods in a separate file, remove details of the internal structure from the header, and separate the shared GC lock (for IPC) into a separate file.

### DIFF
--- a/arch/arch.mk
+++ b/arch/arch.mk
@@ -2,7 +2,7 @@
 
 LIBARCH_SOURCES := \
         simd_vector.cc \
-	simd_vector_avx.cc \
+        simd_vector_avx.cc \
         demangle.cc \
 	tick_counter.cc \
 	cpuid.cc \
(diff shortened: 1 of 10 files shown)


steadtler:
+1

PR #633, opened 2016-08-24
MLDB-1903 arm64 compilation

jeremybarnes:
This is a grab-bag of portability and cross-compilation fixes that enables MLDB to build ARM and ARM64 binaries.

Highlights are:
- Use of the GNU multiprecision library for math on 128 bit integers on 64 bit platforms where the compiler doesn't help us out
- Protobuf on cross compile will first compile a host version, then use that to cross-compile a target version
- Lots of refactoring or guarding of intel ISA constructs in the lower level libraries
- Addition of a ports system that knows how to download a development environment from the .debs for the target architecture and use it in the build
- Documentation for how to build with the new architectures

Test failures are make dependencies which should be fixed shortly.


guyd:
Reviewed rapidly without comment.  +1 once the build is green.

PR #634, opened 2016-08-24
[MLDB-1558] removed the arbitrary limit in CellValue::parse so that i…

guyd:
…t supports parsing of very large floats


guyd:
I agree that this is not ideal.  Note that the memcpy is not new but I will see if I can find a decent implementation.

guyd:
@jeremybarnes I've reviewed a few opened source implementations of strtod.  Gay's implementation is the reference but it is somewhat involved.  Modifying it would be risky.  Other simpler implementations are available but they are prone to rounding and precision errors (this is why we moved to the standard strtod in the first place).
Unless we feel this is central for MLDB, I suggest we keep the memcpy for the moment.  Suggestions?  Concerns?

jeremybarnes:
OK, agreed.  Could you copy that comment and a link to the implementation that you found into the source so that we don't lose track?

PR #641, opened 2016-08-31
MLDB-1913 v8 upgrade

jeremybarnes:
Upgrades v8 to the very latest version, removing a system dependency and allowing arm64 builds (the version of v8 we used didn't work on arm64).

This involves the following changes:

1.  Discovered how to build v8 using <strike>scons</strike> <strike>GYP</strike> GN, using the system ICU (which clashes with the one included in v8), and with cross-compilation available.  The results of this work are in the mldbai/v8-cross-build repo, and the output is cached in the v8-cross-build-output repo which is included in MLDB under ext/.
2.  Updated all of the JS code to the latest v8 API.  This was mostly mechanical, and in general the new API is cleaner and better which improved the quality of the JS binding code.
3.   Created a new `Platform` implementation for v8 with access to the thread pool etc inside MLDB, so that it would better integrate.  This should eventually allow us to receive trace messages from V8 directly in MLDB's logging system.
4.  Adjusted the initialization to find the external binary blobs that v8 now needs to be available to the execution.  This may require some adjustments to the packaging.

I will verify that a docker build works before merging.

### DIFF
--- a/plugins/lang/js/dataset_js.cc
+++ b/plugins/lang/js/dataset_js.cc
@@ -90,9 +90,14 @@ recordRow(const v8::FunctionCallbackInfo<v8::Value> & args)
     JsContextScope scope(args.This());
     try {
         Dataset * dataset = getShared(args.This());
-            
-        dataset->recordRow(JS::getArg<RowName>(args, 0, "rowName"),
-                           JS::getArg<std::vector<std::tuple<ColumnName, CellValue, Date> > >(args, 1, "values", {}));
+         
+        auto rowName = JS::getArg<RowName>(args, 0, "rowName");
+        auto values = JS::getArg<std::vector<std::tuple<ColumnName, CellValue, Date> > >(args, 1, "values", {});
+
+        {
+            //v8::Unlocker unlocker(args.GetIsolate());
+            dataset->recordRow(std::move(rowName), std::move(values));
+        }
 
         args.GetReturnValue().Set(args.This());
     } HANDLE_JS_EXCEPTIONS(args);
@@ -105,9 +110,14 @@ recordRows(const v8::FunctionCallbackInfo<v8::Value> & args)
     JsContextScope scope(args.This());
     try {
         Dataset * dataset = getShared(args.This());
-            
-        dataset->recordRows(JS::getArg<std::vector<std::pair<RowName, std::vector<std::tuple<ColumnName, CellValue, Date> > > > >(args, 0, "rows", {}));
+        
+        auto rows = JS::getArg<std::vector<std::pair<RowName, std::vector<std::tuple<ColumnName, CellValue, Date> > > > >(args, 0, "rows", {});
 
+        {
+            //v8::Unlocker unlocker(args.GetIsolate());
+            dataset->recordRows(std::move(rows));
+        }
+        
         args.GetReturnValue().Set(args.This());
     } HANDLE_JS_EXCEPTIONS(args);
 }
@@ -119,9 +129,14 @@ recordColumn(const v8::FunctionCallbackInfo<v8::Value> & args)
     JsContextScope scope(args.This());
     try {
         Dataset * dataset = getShared(args.This());
-            
-        dataset->recordColumn(JS::getArg<ColumnName>(args, 0, "columnName"),
-                              JS::getArg<std::vector<std::tuple<ColumnName, CellValue, Date> > >(args, 1, "values", {}));
+        
+        auto columnName = JS::getArg<ColumnName>(args, 0, "columnName");
+        auto column = JS::getArg<std::vector<std::tuple<ColumnName, CellValue, Date> > >(args, 1, "values", {});
+
+        {
+            //v8::Unlocker unlocker(args.GetIsolate());
+            dataset->recordColumn(std::move(columnName), std::move(column));
+        }
 
         args.GetReturnValue().Set(args.This());
     } HANDLE_JS_EXCEPTIONS(args);
@@ -135,7 +150,8 @@ recordColumns(const v8::FunctionCallbackInfo<v8::Value> & args)
     try {
         Dataset * dataset = getShared(args.This());
             
-        dataset->recordColumns(JS::getArg<std::vector<std::pair<ColumnName, std::vector<std::tuple<ColumnName, CellValue, Date> > > > >(args, 0, "columns", {}));
+        auto columns = JS::getArg<std::vector<std::pair<ColumnName, std::vector<std::tuple<ColumnName, CellValue, Date> > > > >(args, 0, "columns", {});
+        dataset->recordColumns(std::move(columns));
 
         args.GetReturnValue().Set(args.This());
     } HANDLE_JS_EXCEPTIONS(args);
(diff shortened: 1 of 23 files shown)


plugins/lang/js/dataset_js.cc
```diff
@@ -26,7 +26,8 @@ v8::Handle<v8::Object>
 DatasetJS::
 create(std::shared_ptr<Dataset> dataset, JsPluginContext * context)
 {
-    auto obj = context->Dataset->GetFunction()->NewInstance();
+    v8::Isolate* isolate = v8::Isolate::GetCurrent();
```
guyd: Do we need to free that pointer like it is done below?

plugins/lang/js/function_js.cc
```diff
@@ -24,7 +24,8 @@ v8::Handle<v8::Object>
 FunctionJS::
 create(std::shared_ptr<Function> function, JsPluginContext * context)
 {
-    auto obj = context->Function->GetFunction()->NewInstance();
+    v8::Isolate* isolate = v8::Isolate::GetCurrent();
+    auto obj = context->Function.Get(isolate)->GetFunction()->NewInstance();
```
guyd: Same question here, do we need to free the pointer.

plugins/lang/js/js_common.cc
```diff
@@ -7,27 +7,185 @@
...
+
+    /**
+     * Schedules a task to be invoked on a foreground thread wrt a specific
+     * |isolate|. Tasks posted for the same isolate should be execute in order of
+     * scheduling. The definition of "foreground" is opaque to V8.
+     */
+    virtual void CallOnForegroundThread(Isolate* isolate, Task* task) override
+    {
+        cerr << "CallOnForegroundThread" << endl;
+        std::terminate();
```
guyd: Do we really want to termine MLDB here?  Wouldn't an exception be enough?

plugins/lang/js/js_common.cc
```diff
@@ -7,27 +7,185 @@
...
+     * |isolate| after the given number of seconds |delay_in_seconds|.
+     * Tasks posted for the same isolate should be execute in order of
+     * scheduling. The definition of "foreground" is opaque to V8.
+     */
+    virtual void CallDelayedOnForegroundThread(Isolate* isolate, Task* task,
+                                               double delay_in_seconds) override
+    {
+        cerr << "CallDelayedOnForegroundThread" << endl;
+        cerr << "delay_in_seconds = " << delay_in_seconds << endl;
+        std::terminate();
```
guyd: Same here, do we want to terminate.

plugins/lang/js/js_common.cc
```diff
@@ -7,27 +7,185 @@
...
+     **/
+    virtual const uint8_t* GetCategoryGroupEnabled(const char* name) override
+    {
+        static uint8_t no = 0;
+        return &no;
+    }
+
+    /**
+     * Gets the category group name of the given category_enabled_flag pointer.
+     * Usually used while serliazing TRACE_EVENTs.
```
guyd: serliazing -> serializing

plugins/lang/js/js_common.cc
```diff
@@ -7,27 +7,185 @@
...
+     **/
+    virtual const char*
+    GetCategoryGroupName(const uint8_t* category_enabled_flag) override
+    {
+        static const char dummy[] = "dummy";
```
guyd: May be _category group name_ would be less puzzling.

plugins/lang/js/js_common.cc
```diff
@@ -75,17 +233,39 @@ init(bool forThisThreadOnly)
...
     };
     int v8_argc = sizeof(v8_argv) / sizeof(const char *);
 
     v8::V8::SetFlagsFromCommandLine(&v8_argc, (char **)&v8_argv, false);
+
+    // TODO: linuxisms...
+    char exePath[PATH_MAX];
+    ssize_t pathLen = readlink("/proc/self/exe", exePath, PATH_MAX);
+    if (pathLen == -1)
+        throw HttpReturnException(400, "Couldn't determine v8 library path");
```
guyd: The message is misleading.  We should say that we cannot find our own path.  And I feel this should be a 500.  Not sure how a user could fix that.

plugins/lang/js/dataset_js.cc
jeremybarnes: Where is the isolate freed?  It should never be freed.  It is like that because Google's guidelines never have non-const references, and so it's a pointer.

plugins/lang/js/function_js.cc
jeremybarnes: no

plugins/lang/js/js_common.cc
jeremybarnes: v8 is compiled with exceptions and RTTI off, and this is called from there, so it will have the same effect.

plugins/lang/js/js_common.cc
jeremybarnes: Yes... although we may end up simply implementing those methods.

plugins/lang/js/js_common.cc
jeremybarnes: Copied from the v8 source code

plugins/lang/js/js_common.cc
jeremybarnes: Again, copied from the v8 skeleton.

plugins/lang/js/js_common.cc
jeremybarnes: This would only ever occur if the /proc filesystem wasn't mounted, which is basically an OS error.  Agreed that we could improve the error message.

plugins/lang/js/dataset_js.cc
guyd: Since I don't know the V8 interface, I thought that `EscapableHandleScope` was some kind of guard to free allocated resources.  It is now clear that it is not!

plugins/lang/js/js_common.cc
guyd: Got it.

plugins/lang/js/js_common.cc
```diff
@@ -251,12 +253,18 @@ V8Init(MldbServer * server)
...
-    //    throw HttpReturnException(500, "Couldn't initialize ICU data for v8");
-    //}
-
-    v8::V8::InitializeExternalStartupData("build/x86_64/lib/libv8.so");
+    // TODO: linuxisms...
+    char exePath[PATH_MAX];
+    ssize_t pathLen = readlink("/proc/self/exe", exePath, PATH_MAX);
+    if (pathLen == -1)
+        throw HttpReturnException
+            (400, "Couldn't path to the MLDB executable; "
```
guyd: _Couldn't find path_...

guyd:
+1

jeremybarnes:
Confirmed that the docker build includes the required data files.

jeremybarnes:
The latest version is confirmed to build and start a viable Docker container, and run the jseval tutorial successfully.  So I think that we're safe to merge.

PR #646, opened 2016-09-02
MLDB-1923 clang 3.8 compilation

jeremybarnes:
Switches the default version of clang to 3.8 and fixes the warning messages that it found (mostly around spurious std::move constructs).  This allows compilation and all tests to pass with toolchain=clang.

@jean it requires clang 3.8 to be installed on the build machines.

### DIFF
--- a/arch/simd_vector.cc
+++ b/arch/simd_vector.cc
@@ -37,7 +37,9 @@ void vec_scale(const float * x, float k, float * r, size_t n)
 {
     size_t i = 0;
 
-    if (false) ;
+        if (false)
+        ;
+
 #if JML_INTEL_ISA
     else {
         v4sf kkkk = vec_splat(k);
@@ -71,7 +73,9 @@ void vec_add(const float * x, const float * y, float * r, size_t n)
 {
     size_t i = 0;
 
-    if (false) ;
+        if (false)
+        ;
+
 #if JML_INTEL_ISA
     else {
         //cerr << "unoptimized" << endl;
@@ -110,7 +114,9 @@ void vec_prod(const float * x, const float * y, float * r, size_t n)
 {
     size_t i = 0;
 
-    if (false) ;
+        if (false)
+        ;
+
 #if JML_INTEL_ISA
     else {
         //cerr << "unoptimized" << endl;
@@ -150,7 +156,9 @@ void vec_prod(const float * x, const double * y, float * r, size_t n)
     size_t i = 0;
 
 #if 0 // TODO: do
-    if (false) ;
+        if (false)
+        ;
+
     else {
         //cerr << "unoptimized" << endl;
         
@@ -366,7 +374,9 @@ void vec_scale(const double * x, double k, double * r, size_t n)
 {
     size_t i = 0;
 
-    if (false) ;
+        if (false)
+        ;
+
 #if JML_INTEL_ISA
     else {
         v2df kk = vec_splat(k);
@@ -564,7 +574,9 @@ void vec_minus_sse2(const float * x, const float * y, float * r, size_t n)
 {
     size_t i = 0;
 
-    if (false) ;
+        if (false)
+        ;
+
     else {
         //cerr << "unoptimized" << endl;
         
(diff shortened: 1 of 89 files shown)


jraby:
thx for the heads up. clang-3.8 should be available on the build machines.

arch/simd_vector.cc
```diff
@@ -37,7 +37,9 @@ void vec_scale(const float * x, float k, float * r, size_t n)
 {
     size_t i = 0;
 
-    if (false) ;
+        if (false)
+        ;
```
FinchPowers: Ça sert à quoi un `if (false) ;`?

jeremybarnes: It avoids syntax errors when you have a long if with different parts enabled by the preprocessor.  It means that you have all `else if` branches so it doesn't matter which ones are disabled.  Otherwise you get a syntax error if you disable the first.

rest/rest_request_router.h
```diff
@@ -281,9 +281,31 @@ struct RestRequestParsingContext {
...
+        ObjectEntry(ObjectEntry && other) noexcept
+            : obj(other.obj), type(other.type), deleter(std::move(other.deleter))
+        {
+            other.obj = nullptr;
+            other.deleter = nullptr;
+        }
+
+        ObjectEntry & operator = (ObjectEntry && other) noexcept
+        {
+            ObjectEntry newMe(std::move(other));
```
FinchPowers: Pkoi cet objet temporaire? Pourquoi ne pas caller `sap(other);` directement?

sql/builtin_functions.cc
```diff
@@ -3058,11 +3058,12 @@ BoundFunction tryFct(const std::vector<BoundSqlExpression> & args)
...
             ExcAssertEqual(boundArgs.size(), 1);
             try {
-                return storage = std::move(boundArgs[0](row, GET_LATEST));
+                return storage = boundArgs[0](row, GET_LATEST);
             }
             catch (const std::exception & exc) {
-                return storage = std::move(ExpressionValue(
-                    ML::getExceptionString(), Date::negativeInfinity()));
+                return storage
+                = ExpressionValue(ML::getExceptionString(),
```
FinchPowers: `=` should be on the line above.

sql/expression_value.h
```diff
@@ -1401,7 +1401,7 @@ struct AnyValueInfo: public ExpressionValueInfoT<ExpressionValue> {
 /// For an embedding
 struct EmbeddingValueInfo: public ExpressionValueInfoT<ML::distribution<CellValue, std::vector<CellValue> > > {
     EmbeddingValueInfo(StorageType storageType = ST_ATOM)
-        : EmbeddingValueInfo({-1}, storageType)
+        : EmbeddingValueInfo(std::vector<ssize_t>({-1}), storageType)
```
FinchPowers: `ssize_t -1` ça fait quoi ça (sachant que size_t est unsigned)? Max ssize_t ?

rest/rest_request_router.h
jeremybarnes: It's a common pattern to write a move and copy constructor that's noexcept without repeating code or writing a version of the move assignment that needs to clean up:
- We first move the object into a different copy, which means that the move constructor is what's called.  The move constructor is easier to write as it never needs to clean anything up, and it's noexcept.
- We then swap the new me with *this, which is again noexcept.
- Finally, we destroy newMe, and leave other to be cleaned up by the calling code.  Thus the cleanup is done automatically; no need to write code to clean up.

Since none of these can throw an exception, we have a noexcept move assignment without needing to know any details of how things are implemented.

testing/mldb_config_persistence_test.cc
```diff
@@ -61,7 +61,7 @@ BOOST_AUTO_TEST_CASE( test_plugin_loading )
         }
         
         BOOST_CHECK_EQUAL(proxy.get("/v1/plugins").jsonBody(),
-                          Json::Value({ "jsplugin" }));
+                          Json::Value({ Json::Value("jsplugin") }));
```
FinchPowers: Has this been accidentally added to this PR?

sql/builtin_functions.cc
jeremybarnes: Disagree.  It's much easier to read with operators as the first thing on the line.

sql/expression_value.h
jeremybarnes: ssize_t is signed (size_t is unsigned, ssize_t is short for "signed size_t".  Here, it signifies "I don't know the size of that dimension".

sql/builtin_functions.cc
FinchPowers: I agree to disagree then :p

FinchPowers:
+1

PR #650, opened 2016-09-02
Added the learningRate field to tsne.train

jpilaul:
learningRate was added (equivalent of "eta" in ml/tsne/tsne.h)


mailletf:
+1

PR #653, opened 2016-09-06
Fixed crashes on long running JS scripts (MLDB-1925)

jeremybarnes:
JS scripts that perform a GC cycle need GC to happen, which is done via this callback.  This implements the missing functionality to allow for the GC to actually happen.


steadtler:
+1

PR #657, opened 2016-09-08
Wait for jconfig.h before building the lib

jraby:
Should fix :

```
 In file included from ./mldb/ext/jpeg/jaricom.c:17:0:
 ./mldb/ext/jpeg/jinclude.h:20:55: fatal error: jconfig.h: No such file or directory
```

The original code didn't seem to add a dependency, but it might be magic beyond my gmake powers.
Replaced with an obvious construct.

@jeremybarnes  can you double check please?

### DIFF
--- a/ext/jpeg.mk
+++ b/ext/jpeg.mk
@@ -16,7 +16,7 @@ $(CWD)/jconfig.h:
 	cd mldb/ext/jpeg && ./configure
 endif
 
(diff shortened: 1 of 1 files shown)


jraby:
Hmm that's strange,  it fixes `make -j $(nproc) libjpeg`, which didn't work from a clean checkout (`git clean -ff -xd -e virtualenv -e .make_hash_cache`).
But it is still broken when building `libtensorflow`.  There's something fishy...

PR #658, opened 2016-09-08
[MLDB-1928] rename f to f1Score

simlmx:
Simply rename the confusing "f" to "f1Score" (and "bestF" to "bestF1Score") in the classifier.test output.


mailletf:
+1

PR #659, opened 2016-09-08
protobuf dependency fixes

jraby:
- Add dependency on protobuf headers for pb.h and pb.cc files
 - install protobuf in proper directory when $BIN $LIB and INC are changed
   (docker builds)

### DIFF
--- a/ext/protobuf.mk
+++ b/ext/protobuf.mk
@@ -30,9 +30,6 @@ $(4)/protoc: $(if $(call sne,$(1),$(HOSTARCH)),$(HOSTBIN)/protoc)
 	&& ./autogen.sh > configure-log.txt 2>&1 \
 	&& TMP=$(PWD)/$(TMP) ./configure \
 		--prefix $(PWD)/$(BUILD)/$(1) \
-		--libdir=$(PWD)/$(2) \
-		--includedir=$(PWD)/$(3) \
-		--bindir=$(PWD)/$(4) \
 		--program-suffix="" \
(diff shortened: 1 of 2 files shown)


ext/tensorflow.mk
```diff
@@ -119,7 +119,7 @@ $(INC)/external/eigen_archive/eigen-eigen-$(TENSORFLOW_EIGEN_MERCURIAL_HASH):
 
 # To create a protobuf file, we compile the input file.  Same for the .h
 # file.
-$(CWD)/%.pb.cc $(CWD)/%.pb.h:		$(CWD)/%.proto $(HOSTBIN)/protoc
+$(CWD)/%.pb.cc $(CWD)/%.pb.h:		$(CWD)/%.proto $(HOSTBIN)/protoc $(INC)/google/protobuf
```
jeremybarnes: Should be | $(INC)/google/protobuf as directory timestamps change whenever their contents change and we don't want that to cause a full rebuild.

ext/protobuf.mk
```diff
@@ -30,6 +30,9 @@ $(4)/protoc: $(if $(call sne,$(1),$(HOSTARCH)),$(HOSTBIN)/protoc)
 	&& ./autogen.sh > configure-log.txt 2>&1 \
 	&& TMP=$(PWD)/$(TMP) ./configure \
 		--prefix $(PWD)/$(BUILD)/$(1) \
+		--libdir=$(PWD)/$(2) \
```
jeremybarnes: The problem with this is that it spams masses of unnecessary files into the docker build.  Anything that is needed but was missing should be copied into the tree after build by the rules below, rather than done like this.

PR #662, opened 2016-09-12
Update sample.local.mk

FinchPowers:


jml-build/sample.local.mk
```diff
@@ -1,12 +1,4 @@
-# to speed up repeated builds, use ccache
-CXX := ccache g++
-CC := ccache gcc
-FC := ccache gfortran
+COMPILER_CACHE:=ccache
 
-# to get colorized output with ccache, put colorccache on your path and use
-# CXX := colorccache g++
-# CC := colorccache gcc
```
jeremybarnes: Why turn off tcmalloc?

jraby: It is already enabled in `mldb/Makefile`: https://github.com/mldbai/mldb/blob/master/Makefile#L5

jeremybarnes: +1

PR #664, opened 2016-09-13
Fixed path ordering with unicode and digits (MLDB-1936)

jeremybarnes:
This fixes an issue where the optimized and non-optimized orderings were different, so that paths with and without digits would not have a total ordering.  This most often manifested itself in the classifier training procedure, where "encoding unknown column" errors would occur.


FinchPowers:
+1

PR #665, opened 2016-09-13
MLDB-1927 import text named new default

FinchPowers:
From
```
lineNumber()
```

To
```
dataFileUrl() + '.' + lineNumber()
```

### DIFF
--- a/plugins/importtext_procedure.cc
+++ b/plugins/importtext_procedure.cc
@@ -77,7 +77,7 @@ ImportTextConfigDescription::ImportTextConfigDescription()
     addField("named", &ImportTextConfig::named,
              "Row name expression for output dataset. Note that each row "
              "must have a unique name.",
-             SqlExpression::parse("lineNumber()"));
+             SqlExpression::parse("dataFileUrl() + '.' + lineNumber()"));
     addField("timestamp", &ImportTextConfig::timestamp,
              "Expression for row timestamp.",
              SqlExpression::parse("fileTimestamp()"));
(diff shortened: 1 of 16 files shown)


plugins/importtext_procedure.cc
```diff
@@ -77,7 +77,7 @@ ImportTextConfigDescription::ImportTextConfigDescription()
     addField("named", &ImportTextConfig::named,
              "Row name expression for output dataset. Note that each row "
              "must have a unique name.",
-             SqlExpression::parse("lineNumber()"));
+             SqlExpression::parse("dataFileUrl() + '.' + lineNumber()"));
```
jeremybarnes: Should be [ dataFileUrl(), lineNumber() ]

FinchPowers: I'm not clear on your suggestion. My 2 attemps don't work.
* `dataFileUrl(), lineNumber()` fails with `dataFileUrl(), lineNumber():1:14 ('dataFileUrl()>>>,<<< lineNumber('): expected eof`
* `[dataFileUrl(), lineNumber()]` fails with `Cannot convert value of type 'embedding' to 'atomic value' (Utf8 string)`

Why isn't the way I did it ok?

simlmx: Faut que tu créees un ColumnPath ou qqch du genre non?

guyd: File path can be very long.  I am concerned that a fair portion of the imported dataset memory footprint will be comprised of the repeated file path.  It seems to me that this is a problem specific to the merge dataset.  Why not change the merge dataset to prefix/suffix the row names of the datasets?  Something like`[]-[12]`for rows in the right dataset and `[729]-[]` for rows in the left dataset.

FinchPowers: Suite de la discussion dans le billet.

FinchPowers:
Scrappé en faveur d'une meilleure solution.

PR #666, opened 2016-09-13
Dynamic load Tensorflow kernels for CPU (MLDB-1939)

jeremybarnes:
Our build for Tensorflow had a hard-coded option to use avx operations, which are not supported on older 64 bit CPUs.  By the same token, it was unable to make use of avx2 and FMA instructions available on new 64 bit CPUs.

This patch modifies the building of the Tensorflow kernels to build a separate library optimized for each processor architecture, and then select the appropriate one at runtime.

Note that it adds quite a long time to the build process, since there are four different options built.  Adding `TF_KERNEL_VARIANTS:=avx` or `TF_KERNEL_VARIANTS:=sse42` or `TF_KERNEL_VARIANTS:=avx2` to local.mk will cause only a single one to be built, improving compile times.

### DIFF
--- a/ext/tensorflow.mk
+++ b/ext/tensorflow.mk
@@ -338,18 +338,74 @@ TENSORFLOW_KERNEL_CC_FILES:=$(shell find $(CWD)/tensorflow/core/kernels -name "*
 
 TENSORFLOW_KERNEL_CC_BUILD:=$(sort $(TENSORFLOW_KERNEL_CC_FILES:$(CWD)/%=%))
 
-# For each target architecture we set the flags here
-TF_ARCH_FLAGS_x86_64:=-mavx
-
-$(eval $(call set_compile_option,$(TENSORFLOW_KERNEL_CC_BUILD) $(TENSORFLOW_PROTOBUF_BUILD),$(TENSORFLOW_COMPILE_FLAGS) $(TF_ARCH_FLAGS_$(ARCH))))
-
+# Dependencies for the kernel files... basically everything else needs to
+# be built first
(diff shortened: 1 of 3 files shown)


mailletf:
We should probably add the TF_KERNEL_VARIANTS flag explanation to the build document

FinchPowers:
We should come up with a solution to autoset `F_KERNEL_VARIANTS:=...` properly.
A lesser alternative would be to document that into sample.local.mk.

jeremybarnes:
@FinchPowers for a release, we need all variants.  Otherwise it won't work on lesser machines than the one it was compiled on, and will run more slowly than necessary on more capable machines.  The release build should be the default IMO to ensure that it's properly tested by the build bot.  For local development, we can get away with a single one; there it's OK to autoset it.  This will save build time for developers.

FinchPowers:
Is there a command to run to know which one to use?

jeremybarnes:
I added some documentation to sample.local.mk

jeremybarnes:
Anyone feel like hitting +1?

FinchPowers:
J'Attend que ma compilation termine... :p

FinchPowers:
Sauf que là Jenkins va prendre 20 minutes à chaque fois... Je ne pense pas qu'on veut ça.

FinchPowers:
+1. On va regarder si ccache fait la job ou une alternative pour que les tests demeurent rapides.

jeremybarnes:
Jenkins a pris 5 mins le deuxime build... je crois que ccache marche.

PR #670, opened 2016-09-15
Mldb 1933 subselect flatten

steadtler:
Allow sub dataset to keep structured data (expression values instead of cell values)
Allows structured queries that return non-flattened data
Add `STRUCTURED` keyword to column expr to select non-flattened columns


builtin/sub_dataset.cc
```diff
@@ -47,48 +47,56 @@ SubDatasetConfigDescription()
...
     {
         SqlExpressionMldbScope mldbContext(owner);
 
-        std::vector<MatrixNamedRow> rows
-            = queryFromStatement(statement, mldbContext);
+        std::vector<NamedRowValue> rows;
+        auto pair = queryFromStatementExpr(statement, mldbContext);
 
-        init(std::move(rows));
+        columnInfo = std::get<1>(pair);
```
jeremybarnes: missing a std::move() here

builtin/sub_dataset.cc
```diff
@@ -248,9 +283,11 @@ struct SubDataset::Itl
     {
         std::vector<std::tuple<RowName, CellValue> > result; 
 
-        for (auto row : subOutput)
+        for (const auto& row : subOutput)
         {
-            for (auto c : row.columns)
+            auto flattened = row.flatten();
```
jeremybarnes: We should use flattenDestructive() here, since we don't need the input.

builtin/sub_dataset.cc
```diff
@@ -357,6 +394,33 @@ getRowStream() const
...
+    }
+
+    return Dataset::getKnownColumnInfo(columnName);
+}
+
+std::vector<ColumnName> 
+SubDataset::
+getFlattenedColumnNames() const
+{
+    return itl->fullFlattenColumnNames;
```
jeremybarnes: should be `fullyFlattenedColumnNames`

builtin/transposed_dataset.cc
```diff
@@ -54,7 +54,7 @@ struct TransposedDataset::Itl
         : dataset(dataset),
           matrix(dataset->getMatrixView()),
           index(dataset->getColumnIndex()),
-          columnCount(matrix->getColumnNames().size())
+          columnCount(dataset->getFlattenedColumnNames().size())
```
jeremybarnes: We should add `Dataset::getFlattenedColumnCount()` here so that we don't have to create and copy the whole list, only to keep just its length.

jeremybarnes: Ah, looks like you implemented it but then didn't use it here.

server/analytics.cc
```diff
@@ -36,7 +36,7 @@ namespace MLDB {
 /** Equivalent to SELECT (select) FROM (dataset) WHEN (when) WHERE (where), and each matching
     row is passed to the aggregator.
 */
-bool iterateDataset(const SelectExpression & select,
```
jeremybarnes: This needs to continue to return whether it finished or not.  The ExpressionValueInfo should be returned as a separate element in a pair.

server/analytics.cc
```diff
@@ -48,11 +48,15 @@ bool iterateDataset(const SelectExpression & select,
...
-        .execute(processor, offset, limit, onProgress);
+    BoundSelectQuery query(select, from, alias, when, where, orderBy, calc);
+
+    query.execute(processor, offset, limit, onProgress);
+
+    return query.selectInfo;
+
 }
 
-bool iterateDatasetExpr(const SelectExpression & select,
```
jeremybarnes: Again, these bools have been removed several times and we've had to put them back each time.  We know from experience that we need them in.

server/dataset_context.cc
```diff
@@ -422,14 +422,15 @@ doGetBoundParameter(const Utf8String & paramName)
...
+doGetAllColumnsInternal(const Utf8String & tableName,
+                    const ColumnFilter& keep,
+                    bool atoms)
 {
     if (!tableName.empty()
-        && std::find(childaliases.begin(), childaliases.end(), tableName)
-            == childaliases.end()
-        && tableName != alias)
-        throw HttpReturnException(400, "Unknown dataset " + tableName);
+    && std::find(childaliases.begin(), childaliases.end(), tableName)
```
jeremybarnes: Indentation is screwed up

sql/expression_value.cc
```diff
@@ -867,6 +901,13 @@ getKnownColumns() const
     return columns;
 }
 
+/*std::vector<KnownColumn>
```
jeremybarnes: SHould just remove probably.

sql/expression_value.h
```diff
@@ -225,6 +225,10 @@ struct ExpressionValueInfo {
     /// a row.
     virtual std::vector<KnownColumn> getKnownColumns() const;
 
+    /// Return the set of known atoms for a row.  Default throws that it's not
```
jeremybarnes: I don't think this should be here.  It does type analysis, and so should be in ExpressionValueInfo.

jeremybarnes:
Still some work to do on this one.

builtin/transposed_dataset.cc
steadtler: Implemented it afterwards... will change it.

builtin/sub_dataset.cc
steadtler: But that means I have to copy the row. Wont that potentially be costlier?

sql/expression_value.h
steadtler: ... thats where it is

PR #671, opened 2016-09-15
Re-enable merge ds internal sorting

FinchPowers:


builtin/merged_dataset.cc
```diff
@@ -79,9 +79,12 @@ struct MergedDataset::Itl
                   [] (std::shared_ptr<Dataset> p1,
                       std::shared_ptr<Dataset> p2)
                   {
-                      return false;
```
jeremybarnes: Please add a virtual method `getRowCount()` to `Dataset` with a default implementation that forwards to `getMatrixView()->getRowCount()`, and sort on that highest to lowest.  We don't care about the column counts at all.  We don't want to use `getRowNames()` because it's a very expensive operation.

testing/MLDB-963-when-in-WHEN.py
```diff
@@ -28,6 +29,7 @@ def setUpClass(cls):
         ds1.record_row(str(row_count - 1), [['x', "9", same_time_tomorrow],
                                             ['y', "9", same_time_tomorrow]])
         ds1.commit()
+        time.sleep(1)
```
jeremybarnes: This is not a valid fix to any problem that I can think of...

core/dataset.h
```diff
@@ -685,6 +685,8 @@ struct Dataset: public MldbEntity {
        This will return the name that the row has in the table with this alias*/
     virtual RowName getOriginalRowName(const Utf8String& tableName,
                                        const RowName & name) const;
+
+    virtual size_t getRowCount() const;
```
jeremybarnes: Minor nitpick: should be uint64_t, since on 32 bits an external dataset may have more than 4 billion rows.

testing/MLDB-963-when-in-WHEN.py
```diff
@@ -14,7 +14,7 @@ class WhenInWhen(unittest.TestCase):
 
     @classmethod
     def setUpClass(cls):
-        same_time_tomorrow = now + datetime.timedelta(days=1)
+        same_time_tomorrow = now + datetime.timedelta(seconds=3599)
```
jeremybarnes: That's the same time next hour, not tomorrow :)

jeremybarnes:
+1 once you've considered my final (minor) comments and made sure that the test failure isn't related.

testing/MLDB-963-when-in-WHEN.py
FinchPowers: Aha oups.

FinchPowers:
* Fixed size_t -> uint64_t
* Fixed timing issue with the test

PR #673, opened 2016-09-15
small fixes and improvements to tensorflow plugin

steadtler:
Some incorrect or missing type conversion, found while working on a demo.

### DIFF
--- a/sql/expression_value.cc
+++ b/sql/expression_value.cc
@@ -1771,7 +1771,10 @@ StorageType
 ExpressionValue::
 getEmbeddingType() const
 {
-    return embedding_->storageType_;
+    if (type_ == Type::EMBEDDING)
+        return embedding_->storageType_;
+    
+    throw HttpReturnException(500, "Querying embedding type on non-embedding value");
 }
 
 ExpressionValue
--- a/sql/expression_value.cc
+++ b/sql/expression_value.cc
@@ -1767,6 +1767,13 @@ embedding(Date ts,
     return result;
 }
 
+StorageType
+ExpressionValue::
+getEmbeddingType() const
+{
+    return embedding_->storageType_;
+}
+
 ExpressionValue
 ExpressionValue::
 superpose(std::vector<ExpressionValue> vals)
--- a/sql/expression_value.h
+++ b/sql/expression_value.h
@@ -889,7 +889,8 @@ struct ExpressionValue {
     */
     void convertEmbedding(void * buf, size_t len, StorageType bufType) const;
 
-    /** Return the storage type of the embedding 
+    /** Return the storage type of the embedding
+        Will throw an error if not an embedding.
     */
     StorageType getEmbeddingType() const;
 
--- a/sql/expression_value.h
+++ b/sql/expression_value.h
@@ -889,6 +889,10 @@ struct ExpressionValue {
     */
     void convertEmbedding(void * buf, size_t len, StorageType bufType) const;
 
+    /** Return the storage type of the embedding 
+    */
+    StorageType getEmbeddingType() const;
+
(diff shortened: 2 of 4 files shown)


tensorflow/tensorflow_plugin.cc
```diff
@@ -854,54 +859,60 @@ struct TensorflowGraphBase: public Function {
             }
             
             switch (type) {
-            case tensorflow::DT_FLOAT: {
```
jeremybarnes: Noise... what is the actual change here?

sql/expression_value.cc
```diff
@@ -1767,6 +1767,13 @@ embedding(Date ts,
     return result;
 }
 
+StorageType
+ExpressionValue::
+getEmbeddingType() const
+{
+    return embedding_->storageType_;
```
jeremybarnes: Too dangerous: this will segfault if it's not an embedding.  We should also make it work for a non-embedding type, although that can happen later.

sql/expression_value.h
```diff
@@ -889,6 +889,10 @@ struct ExpressionValue {
     */
     void convertEmbedding(void * buf, size_t len, StorageType bufType) const;
 
+    /** Return the storage type of the embedding
```
jeremybarnes: Comment about when it's appropriate to call it.

tensorflow/tensorflow_plugin.cc
```diff
@@ -1175,6 +1193,35 @@ struct TensorflowGraphBase: public Function {
...
+                return DT_UINT8;
+            case ST_INT8:
+                return DT_INT8;
+            case ST_INT16:
+                return DT_INT16;
+            case ST_UINT16:
+                return DT_UINT16;
+            case ST_BLOB:
+                return DT_STRING;
+            default:
```
jeremybarnes: Dates should convert to doubles.

tensorflow/tensorflow_plugin.cc
steadtler: I added UINT16

tensorflow/tensorflow_plugin.cc
```diff
@@ -974,7 +985,11 @@ struct TensorflowGraphBase: public Function {
             }
         }
         else if (val.isEmbedding()) {
-            // TODO: infer consistent type for tensor, and convert it
+            auto storageType = val.getEmbeddingType();
+            if (storageType == ST_ATOM) {
+                storageType = ST_FLOAT64; //TODO not sure its possible to get consistent type then
```
FinchPowers: I can't really +1 with such a comment... :p

tensorflow/tensorflow_plugin.cc
jeremybarnes: OK, please avoid re-indenting the whole lot to make a single small change.

tensorflow/tensorflow_plugin.cc
FinchPowers: Then it's not a TODO but a simple comment.

steadtler: I see it as a big improvement over the previous TODO! I'll add something to tensorflow.md, not quite sure how to word it.

tensorflow/tensorflow_plugin.cc
steadtler: sorry, had issue with an unmatched brace, and the indentation change didnt appear in git diff.

tensorflow/tensorflow_plugin.cc
FinchPowers: C'est moi et ma phobie des TODO. Si tu veux tu peux en parler avec @simlmx :p

PR #674, opened 2016-09-16
fixing numerical errors - numerical values are slightly different dep…

guyd:
…ending on the platform we run the test on.


mailletf:
+1

PR #676, opened 2016-09-19
Edlib unit test integration

FinchPowers:
@jeremybarnes looks good or makes you sick? ;)


mailletf:
do the test pass without the git-lfs files?

FinchPowers:
Oui.

jeremybarnes:
It doesn't make me sick, but I feel that unit testing of the libraries that we depend on is of very limited value.  I'm much rather we rewrite the unit tests as MLDB tests so that we exercise the interface... unless we're modifying the edlib code itself that is.

FinchPowers:
Yes I'm going to modify edlib to support utf-8.

jeremybarnes:
OK +1 then

PR #678, opened 2016-09-21
Fixed protobuf sched_yield failure in configure (MLDB-1953)

jeremybarnes:
Fixes build failures when TMP is an absolute path, but keeps it working when TMP is a relative path.


FinchPowers:
+1

PR #679, opened 2016-09-21
MLDB-1954 clang

jeremybarnes:
The latest instalment in making it possible to develop MLDB on a laptop on a plane...


FinchPowers:
+1

PR #684, opened 2016-09-26
modified the test to fail gracefully if pymongo is not installed

guyd:


guyd:
pymongo is part of the python dependencies.  There is no need for the extra check.

PR #687, opened 2016-09-28
MLDB-1958 fix outer join on false

FinchPowers:


testing/MLDB-1624-more-join-test.py
```diff
@@ -2081,26 +2081,49 @@ def test_equijoin_with_larger_table(self):
...
+                    [ "[]-[row2]", None, 1]]
+        res = mldb.query("""
+            SELECT * FROM (SELECT * FROM ds1) AS s1
+            OUTER JOIN (SELECT * FROM ds2) AS s2 ON false
+        """)
+        self.assertTableResultEquals(res, expected)
+
+        res = mldb.query("""
+            SELECT * FROM (SELECT * FROM ds1) AS s1
+            OUTER JOIN (SELECT * FROM ds2) AS s2 ON s1.rowName() = 'wwwwwwwww'
```
guyd: Can you also try with a more complex ON clause so that it uses the pipeline executor?  Something like
```
s1.rowName() = 'wwwwwwwww' AND s2.rowName() = 'wwwwwwwww'
```
should do the trick.

builtin/joined_dataset.cc
```diff
@@ -398,12 +399,20 @@ struct JoinedDataset::Itl
                     //return all rows
                     sideCondition = SqlExpression::TRUE;
 
+                    shared_ptr<SqlExpression> sideWhere;
```
guyd: Can you add a comment here stating why we need this extra step?  This code is getting harder to understand everyday! :-(

PR #692, opened 2016-09-30
Mldb 408 task cancellation

guyd:
- added the code to cancel a task using a PUT on a run with the cancellation payload
- adapted Bucketize to allow cancellation
- simplified the data members of the BackgroundTask.  It now comprises one bool and the state of the task.
- adapted some of the dataset iteration code to forward cancellation to callers

The changes of the bucketize procedure will have to be applied to all procedures.  I am planning to do this in a separate PR.


FinchPowers:
Ouch. We will need to update all functions? The same thing we *did* when progress was added? Is there no other way?

rest/rest_collection.h
```diff
@@ -135,14 +142,16 @@ struct BackgroundTaskBase {
     void setHandle(int64_t handle)
     {
         std::unique_lock<std::mutex> guard(mutex);
+        std::cerr << "new task with handle " << handle << std::endl;
```
FinchPowers: rm cerr

testing/MLDB-408-task-cancellation.py
```diff
@@ -0,0 +1,65 @@
...
+                    'id' : 'output',
+                    'type' : 'sparse.mutable'
+                },
+                'percentileBuckets': {
+                    'b1': [0, 25],
+                    'b2': [25, 50],
+                    'b3': [50, 75],
+                    'b4': [75, 100]
+                },
+                'runOnCreation' : False
```
FinchPowers: Shouldn't it be running ?

rest/rest_collection.h
guyd: will do

testing/MLDB-408-task-cancellation.py
guyd: If you look at the function above, it starts the procedure asynchronously so that we can later cancel it.

core/procedure.cc
```diff
@@ -38,6 +38,7 @@ ProcedureRunConfigDescription()
     nullAccepted = true;
 
     addField("id", &ProcedureRunConfig::id, "ID of run");
+    addField("state", &ProcedureRunConfig::state, "State of the run");
```
jeremybarnes: This shouldn't be here.  It's not a configuration parameter.  That's why I don't think we should POST to the same route.

plugins/bucketize_procedure.cc
```diff
@@ -224,7 +228,9 @@ run(const ProcedureRunConfig & run,
         DEBUG_MSG(logger) << "Bucket " << mappedRange.first << " from "
                           << lowerBound << " to " << higherBound;
 
-        parallelMap(lowerBound, higherBound, applyFct);
+        if (!parallelMapHaltable(lowerBound, higherBound, applyFct)) {
+            DEBUG_MSG(logger) << BucketizeProcedureConfig::name << " procedure was cancelled";
```
jeremybarnes: Procedure was cancelled notification should be handled by the caller, not in each procedure.

plugins/sql_functions.cc
```diff
@@ -628,7 +628,10 @@ run(const ProcedureRunConfig & run,
             .execute({recordRowInOutputDataset, true /*processInParallel*/},
                      runProcConf.inputData.stm->offset,
                      runProcConf.inputData.stm->limit,
-                     onProgress);
+                     onProgress) )
+            {
+                DEBUG_MSG(logger) << TransformDatasetConfig::name << " procedure was cancelled";
```
jeremybarnes: Shouldn't we also exit here?  Maybe by throwing a MLDB::Cancellation exception?

rest/rest_collection.cc
```diff
@@ -670,11 +670,12 @@ void
 BackgroundTaskBase::
 cancel()
 {
-    bool wasCancelled = this->cancelled.exchange(true);
```
jeremybarnes: There were a lot of race conditions causing crashes in this code which were fixed with a big amount of effort.  I think we need a stress test to ensure that they stay fixed; this code is very sensitive.

core/procedure.cc
guyd: It is actually a PUT and why would the state not be part of the _run_ configuration?  I can see a run be running, cancelled, paused and so on.

rest/rest_collection.cc
```diff
@@ -752,6 +768,7 @@ getState() const
     case State::_executing:
         return L"executing";
     default:
+        ExcAssert(!"update the BackgroundTaskBase::getState");
```
jeremybarnes: Can you add an extra comment to ensure that we know what to do if this triggers?

plugins/bucketize_procedure.cc
guyd: You mean the logger message or some other notification?

I am not sure if there are a single place where I can log the message.  I will take a look.

rest/rest_collection.h
```diff
@@ -108,6 +108,8 @@ getDefaultDescription(RestCollectionChildEvent<Key, Value> * = 0);
 /** Base class for background tasks. */
 
 struct BackgroundTaskBase {
+    /// _initializing and _executing are transient states while 
+    /// _cancelled, _finished and _error are final states 
     enum State {_initializing, _executing, _cancelled, _finished, _error};
```
jeremybarnes: We use CAPS for enum constants.  leading underscores are reserved for the OS and compiler.  Please rename to allcaps.

rest/rest_collection_impl.h
```diff
@@ -625,30 +628,33 @@ finishedBackgroundJob(Key key,
 
     std::unique_lock<std::mutex> guard(task->mutex);
 
-    if (task->cancelled) {
+    if (task->state == BackgroundTaskBase::State::_cancelled ||
+        task->state == BackgroundTaskBase::State::_error) {
+
+        // there are no object created
```
jeremybarnes: There *is* no object created

rest/rest_collection.cc
guyd: I've noticed indeed that it was sensitive.   This is why I added a few assert.  I am 100% for the stress test.  Doing it now!

rest/rest_collection.cc
guyd: Isn't exactly what the message of the assert is doing?

rest/rest_collection_impl.h
```diff
@@ -1503,7 +1509,7 @@ addPutRoute()
                     { "EntityPath", jsonEncodeStr(path) }
                 };
 
-                connection.sendHttpResponse(201, jsonEncodeStr(status),
```
jeremybarnes: Are you sure?  Only if we've asked for async should it be 202.

rest/rest_collection.h
guyd: Will do.

server/procedure_run_collection.cc
```diff
@@ -44,10 +45,126 @@ initRoutes(RouteManager & manager)
 
     ExcAssert(manager.getKey);
 
-    manager.addPutRoute();
```
jeremybarnes: Why copy out the whole boilerplate below rather than use this method?

rest/rest_collection_impl.h
guyd: This is precisely the async case.

testing/MLDB-409-transform-dataset.js
```diff
@@ -64,7 +64,8 @@ var transform_config = {
             named: "rowName() + '_transformed'",
             limit: 3
         },
-        outputDataset: { id: 'transformed', type: 'sparse.mutable' }
+        outputDataset: { id: 'transformed', type: 'sparse.mutable' },
+        runOnCreation : false
```
jeremybarnes: Why are these tests being modified?

plugins/sql_functions.cc
guyd: We clearly need to exit here.  Good catch.

rest/rest_collection.cc
jeremybarnes: Only you know the meaning of that message, I don't understand it :) .  Is it a logic error?  Unimplemented code?  If so, what does it need to do?  etc.

server/procedure_run_collection.cc
guyd: The boilerplace was copied and modified to handle the update of a procedure run.  This is clearly not ideal.  I will try to minimize the code duplication.

rest/rest_collection.cc
guyd: OK.  I will be more explicit.

testing/MLDB-409-transform-dataset.js
guyd: runOnCreation is true by default.  This test was running the procedure twice because it explicitly creates a procedure run after creating the procedure.  This was not causing any problem but I decided to fix it anyway.

core/cancellation_exception.h
```diff
@@ -0,0 +1,31 @@
...
+
+#include "mldb/arch/exception.h"
+#include "mldb/types/any.h"
+#include "mldb/types/string.h"
+
+namespace MLDB {
+
+struct CancellationException: public Exception {
+    CancellationException(const Utf8String & message)
+        : Exception(message.rawData()), message(message)
```
jeremybarnes: This will cause a hanging pointer, as we don't control the lifetime of the message parameter.  It will be tough to fix.

core/cancellation_exception.h
```diff
@@ -0,0 +1,31 @@
...
+
+#pragma once
+
+#include "mldb/arch/exception.h"
+#include "mldb/types/any.h"
+#include "mldb/types/string.h"
+
+namespace MLDB {
+
+struct CancellationException: public Exception {
```
jeremybarnes: This should derive from SilentException, its methods should not be inline (there should be a .cc file), the details should be handled opaquely with a unique_ptr<Itl> and we should not need to include the any or string headers.  All this to make sure that the ABI is as tidy as possible, as this will be an API header.

jeremybarnes:
I added some comments, mostly minor.  +1 once they're addressed.

testing/MLDB-408-task-cancellation.py
```diff
@@ -0,0 +1,92 @@
...
+    def run_procedure_async(self, name, config):
+        mldb.put("/v1/procedures/" + name, config)
+        response = mldb.post_async("/v1/procedures/" + name + "/runs")
+        mldb.log(response)
+        return response.headers['Location']
+        
+    def run_and_cancel(self, name, config):
+        location = self.run_procedure_async(name, config)
+
+        resp = mldb.put(location + "/state", {'state': 'cancelled'})
```
FinchPowers: @guyd un peu tard pour mon commentaire, mais il serait préférable (plus REST) que le put soit sur location direct, sans `+ "/state"`. C'est mis en évidence par le fait que state est là deux fois, dans l'URL et comme clé.

guyd: Jeremy et moi avons discuté de ça dans un commentaire plus haut.

PR #693, opened 2016-09-30
Mldb 1391 classifier proc err on bad cfg

FinchPowers:
Validates the classifier configuration payloads.
* The validation can be wrong if dotted keys are used. The problem lies with the `Configuration` class which allows them, making validation "impossible" in its current state. We could either refuse dotted keys or use a Path system. (A test cases demonstrates the situation.)
* End values starting with an underscore are ignored, which allows commenting. (This pattern is used in `classifiers.json`.)

### DIFF
--- a/jml/utils/string_functions.cc
+++ b/jml/utils/string_functions.cc
@@ -240,21 +240,4 @@ trim(const string & other)
     return result;
 }
 
-std::string
-join(const std::vector<std::string> strings)
-{
-    stringstream ss;
-    bool first = true;
-    for (const auto & s: strings) {
-        if (JML_UNLIKELY(first)) {
-            first = false;
-        }
-        else {
-            ss << " ";
-        }
-        ss << s;
-    }
-    return ss.str();
-}
-
 } // namespace ML
--- a/jml/utils/string_functions.cc
+++ b/jml/utils/string_functions.cc
@@ -240,5 +240,21 @@ trim(const string & other)
     return result;
 }
 
+std::string
+join(const std::vector<std::string> strings)
+{
+    stringstream ss;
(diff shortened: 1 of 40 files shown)


plugins/classifier.cc
```diff
@@ -633,23 +649,8 @@ run(const ProcedureRunConfig & run,
...
-                                  runProcConf.configurationFile :
-                                  "/opt/bin/classifiers.json");
-        classifierConfig = jsonDecodeStream<ML::Configuration>(stream);
-    }
-
     timer.restart();
 
-    std::shared_ptr<ML::Classifier_Generator> trainer
-        = ML::get_trainer(runProcConf.algorithm,
-                          classifierConfig);
```
FinchPowers: Moved this code higher in the function to get validation before the query runs.

jml/utils/string_functions.cc
```diff
@@ -240,5 +240,21 @@ trim(const string & other)
     return result;
 }
 
+std::string
+join(const std::vector<std::string> strings)
```
jeremybarnes: This should take a reference, and ideally we wouldn't add another function in string_functions.cc (which we want to remove).  Use boost::string instead.

jeremybarnes:
This is OK, but I think it would be much better to simply convert the JML configuration mechanism to ValueDescription and then delete the Configuration class (which was written before JSON even existed).

I'm OK with doing it in two steps if we don't think it's worth investing the time now.

jml/utils/string_functions.cc
FinchPowers: So many things to know... sure no problem! :)

PR #699, opened 2016-09-30
fixing the doc after the removal of a namespace

guyd:


guyd:
@mailletf merge when you want

PR #703, opened 2016-10-04
Mldb 1812 row name to row path

FinchPowers:
RowName type to RowPath, ColumnName to ColumnPath.
I did not go thoroughly through all variables and function names.


builtin/filtered_dataset.cc
```diff
@@ -109,17 +109,17 @@ struct FilteredDataset::Itl
 
     /** MatrixView */
 
-    virtual RowName getRowName(const RowHash & rowHash) const
```
jeremybarnes: Should also be renamed to getRowPath

builtin/filtered_dataset.cc
```diff
@@ -109,17 +109,17 @@ struct FilteredDataset::Itl
...
+    virtual RowPath getRowName(const RowHash & rowHash) const
     {
         return matrixView->getRowName(rowHash);
     }
 
-    virtual ColumnName getColumnName(ColumnHash column) const
```
jeremybarnes: etc

PR #706, opened 2016-10-04
Updated support email

mailletf:


jraby:
Your changes look good to me. I added a few more tweaks, please double check.

PR #709, opened 2016-10-05
Lidar improvements

steadtler:
Various fixes and improvements that I needed for the lidar demo.

Include: Generic extract expressions
image wrapper
volume wrapper
coerce to embedding
other fixes


plugins/embedding.cc
```diff
@@ -1334,6 +1336,269 @@ regNearestNeighborsFunction(builtinPackage(),
...
+}
+
+static RegisterFunctionType<ImageWrapperFunction, ImageWrapperFunctionConfig>
+regImageWrapperFunction(builtinPackage(),
+                            "image.wrapper",
+                            "Wraps access to a 2d embedding",
+                            "functions/ImageWrapperFunction.md.html");
+
+/*****************************************************************************/
+/* Get N Neighbor function                                                   */
```
steadtler: Im open to a different name

container_files/public_html/doc/builtin/functions/ImageWrapperFunction.md
```diff
@@ -0,0 +1,29 @@
+# Image Wrapper Function
```
jeremybarnes: Image wrapper is not a great user facing name.  Maybe read pixels?

core/value_function.cc
```diff
@@ -49,6 +49,11 @@ toValueInfo(std::shared_ptr<const ValueDescription> desc)
                 info.desc = field;
                 std::tie(info.info, info.fromInput, info.toOutput)
                     = toValueInfo(field.description);
+
+                if (!info.info)
+                    throw HttpReturnException
+                            (500, "No info for field " + info.fieldName.toUtf8String());
```
jeremybarnes: Let's add the field.description to the exception so that if this exception triggers we know which type we need to implement.  Also, this check should probably be in toValueInfo and we can document that it won't return null pointers.

plugins/embedding.cc
```diff
@@ -1334,6 +1336,269 @@ regNearestNeighborsFunction(builtinPackage(),
...
+/* Image Wrapper function                                                    */
+/*****************************************************************************/
+
+DEFINE_STRUCTURE_DESCRIPTION(ImageWrapperFunctionConfig);
+
+ImageWrapperFunctionConfigDescription::
+ImageWrapperFunctionConfigDescription()
+{
+    addField("expression",
+             &ImageWrapperFunctionConfig::expression,
```
jeremybarnes: We do need to document the parameters here.  We can then auto-generate the documentation and it's sure to stay up to date with the code.

plugins/embedding.cc
```diff
@@ -1334,6 +1336,269 @@ regNearestNeighborsFunction(builtinPackage(),
...
+ImageWrapperInput()
+{
+}
+
+DEFINE_STRUCTURE_DESCRIPTION(ImageWrapperInput);
+
+ImageWrapperInputDescription::
+ImageWrapperInputDescription()
+{
+    addField("x", &ImageWrapperInput::x,
```
jeremybarnes: and here

plugins/embedding.cc
```diff
@@ -1334,6 +1336,269 @@ regNearestNeighborsFunction(builtinPackage(),
...
+    addField("y", &ImageWrapperInput::y,
+             "");
+}
+
+DEFINE_STRUCTURE_DESCRIPTION(ImageWrapperOutput);
+
+ImageWrapperOutputDescription::
+ImageWrapperOutputDescription()
+{
+    addField("value", &ImageWrapperOutput::value,
```
jeremybarnes: etc

plugins/embedding.cc
```diff
@@ -1334,6 +1336,269 @@ regNearestNeighborsFunction(builtinPackage(),
...
+ImageWrapperOutput
+ImageWrapperFunction::
+applyT(const ApplierT & applier_, ImageWrapperInput input) const
+{   
+    ImageWrapperOutput output;    
+
+    //We clamp, we do not currently provide interpolation
+    int x = input.x.coerceToInteger().toInt();
+    int y = input.y.coerceToInteger().toInt();
+    x = boost::algorithm::clamp(x, 0, shape[0]-1);
```
jeremybarnes: Instead of clamping, we should return null if the pixel is outside of the image.

plugins/embedding.cc
jeremybarnes: Maybe ProximatePixels?

plugins/embedding.cc
```diff
@@ -1334,6 +1336,269 @@ regNearestNeighborsFunction(builtinPackage(),
...
+{
+    std::unique_ptr<ImageWrapperFunctionApplier> result
+        (new ImageWrapperFunctionApplier(this));
+ 
+    return std::move(result);
+}
+
+static RegisterFunctionType<ImageWrapperFunction, ImageWrapperFunctionConfig>
+regImageWrapperFunction(builtinPackage(),
+                            "image.wrapper",
```
jeremybarnes: image.readPixels?

Also, the "right" way for this to work in SQL would be to have a dataset function `embedding_dataset()` just like `row_table()` that allowed it to be done as a subselect.  For later.

plugins/sparse_matrix_dataset.cc
```diff
@@ -482,6 +482,7 @@ struct SparseMatrixDataset::Itl
             return { hash, 24 };
         }
         default:
+            cerr << "WIERD TYPE: " << val.cellType() << endl;
```
jeremybarnes: Let's make this error message easier for a user to action, by changing the ExcAssert(false) to an throw HttpReturnException(...)

plugins/sql_functions.cc
```diff
@@ -366,7 +366,7 @@ SqlExpressionFunction(MldbServer * owner,
 
         // 4.  Our required input is known by the binding context, as it records
         //     what was read.
-        info.input = innerScope->inputInfo;
```
jeremybarnes: This should use the ExpressionValueInfo::toRow() function, not a dynamic pointer cast.  (blocker)

plugins/sql_functions.cc
```diff
@@ -483,7 +483,7 @@ getFunctionInfo() const
 
     // 4.  Our required input is known by the binding context, as it records
     //     what was read.
-    result.input = outerScope.inputInfo;
+    result.input = std::dynamic_pointer_cast<RowValueInfo>(outerScope.inputInfo);
```
jeremybarnes: Same here.

server/column_scope.h
```diff
@@ -1,9 +1,7 @@
 /** column_scope.h                                                 -*- C++ -*-
     Mathieu Marquis Bolduc, 11 Mars 2016
     Copyright (c) 2016 Datacratic Inc.  All rights reserved.
-
```
jeremybarnes: Why remove this whitespace?  IMO it makes it easier to read...

jeremybarnes: In fact this whole thing is a spurious change and should be removed.

sql/binding_contexts.h
```diff
@@ -350,7 +350,7 @@ struct SqlExpressionExtractScope: public SqlBindingScope {
 
     /// Input variables, for when they are known.  Will be null
     /// when the input is unknown.
-    std::shared_ptr<RowValueInfo> inputInfo;
```
jeremybarnes: I don't understand why this can't be a rowValueInfo?

sql/builtin_functions.cc
```diff
@@ -3275,6 +3275,24 @@ BoundFunction fetcher(const std::vector<BoundSqlExpression> & args)
 }
 static RegisterBuiltin registerFetcherFunction(fetcher, "fetcher");
 
+BoundFunction isconst(const std::vector<BoundSqlExpression> & args)
```
jeremybarnes: should be static_is_constant() to make it clear that we're testing the static analysis

sql/expression_value.cc
```diff
@@ -2608,6 +2608,68 @@ getNestedColumn(const ColumnName & columnName, const VariableFilter & filter) co
     else return *val;
 }
 
+ExpressionValue
+ExpressionValue::
+coerceToEmbedding() const
+{
+    if (type_ == Type::EMBEDDING)
+        return *this;
+    else if (type_ != Type::STRUCTURED)
```
jeremybarnes: In Tensorflow, a scalar is simply an embedding of dimension size zero.  We should consider if this is what we want to do.

sql/expression_value.cc
```diff
@@ -2608,6 +2608,68 @@ getNestedColumn(const ColumnName & columnName, const VariableFilter & filter) co
...
+
+    //Start with finding the shape of the embedding
+    size_t numElements = 0;
+    DimsVector shape;
+    auto atomCount = [&] (const Path & columnName,
+                                   const Path & prefix,
+                                   const CellValue & val,
+                                   Date ts)
+                    {
+                        Path fullColumnName = prefix + columnName;
```
jeremybarnes: If prefix is empty, we should not perform this operation which is expensive.

sql/expression_value.cc
```diff
@@ -2608,6 +2608,68 @@ getNestedColumn(const ColumnName & columnName, const VariableFilter & filter) co
...
+                        if (shape.size() < fullColumnName.size())
+                            shape.resize(fullColumnName.size());
+
+                        for (size_t i = 0; i < fullColumnName.size(); ++i) {
+                            shape[i] = std::max(shape[i], (size_t)(fullColumnName[i].toIndex()+1));
```
jeremybarnes: This function should fail if the rows don't have a consistent size, as it's not an embedding in that case.

sql/expression_value.cc
```diff
@@ -2608,6 +2608,68 @@ getNestedColumn(const ColumnName & columnName, const VariableFilter & filter) co
...
+                            index += fullColumnName[i].toIndex() * compoundShape[i];
+                        }
+                        values[index] = val;
+                        return true;
+                    };
+
+    forEachAtom(onAtom);
+
+    //create the embedding
+    return ExpressionValue(values, Date::negativeInfinity(), shape);
```
jeremybarnes: In a future iteration we should infer the correct type and avoid creating a temporary.

sql/sql_expression.cc
```diff
@@ -1562,6 +1549,22 @@ parse(ML::Parse_Context & context, int currentPrecedence, bool allowUtf8)
             }
         }
 
+        if (context.match_literal('['))
```
jeremybarnes: We want this to bind extremely tightly, so I would put this at the top of the while(true) loop, ie before the IS NULL part.

tensorflow/tensorflow_plugin.cc
```diff
@@ -992,7 +996,7 @@ struct TensorflowGraphBase: public Function {
             return castToTypedTensor(val, storageToDatatype(storageType));
         }
 
-        cerr << "trying to cast " << jsonEncode(val) << " to tensor" << endl;
+        //cerr << "trying to cast " << jsonEncode(val) << " to tensor" << endl;
         throw HttpReturnException(500, "Unable to cast value to tensor");
```
jeremybarnes: Let's add a little more information (some idea of what kind of expression) to help in debugging down the road.

tensorflow/tensorflow_plugin.cc
```diff
@@ -1411,7 +1421,7 @@ struct TensorflowGraphBase: public Function {
         functionScope.inferInput();
         
         FunctionInfo result;
-        result.input = std::move(functionScope.inputInfo);
+        result.input = std::move(std::dynamic_pointer_cast<RowValueInfo>(functionScope.inputInfo));
```
jeremybarnes: Again, no dynamic_pointer_cast... use toRow().

sql/binding_contexts.h
steadtler: Because it can be an embedding now too. We can put it back after we make EmbeddingValueInfo derive from RowValueInfo.

plugins/embedding.cc
steadtler: The issue with that is that tensorflow/the tensorflow plugin can't deal with null values. The classic solution is to add an option for padding, either 0, null, repeat (which is whats done here), wrapover, etc. Later?

plugins/sparse_matrix_dataset.cc
steadtler: Sorry, I though I cleaned that up, maybe I forgot to push it.

sql/expression_value.cc
steadtler: I say we go on a per-need basis.

steadtler:
Verbal + 1 from JB

PR #712, opened 2016-10-07
MLDB-1972 fft builtin

jeremybarnes:
This function makes a major set of improvements to the embedding infrastructure, allowing for us to import an FFT library.

The changes broadly are:
- Tests can now specify which file they are implemented in, and so can be implemented in other languages than c++.  This was used to unit test the pffft library.
- reshape() now requires a 3rd argument when resizing the number of elements describing what any created elements should be.
- we now try harder to coerce to embeddings in functions that only can operate on this representation.  It should mean that there are less cases where rows and embeddings act differently.
- added a system for functions that implement builtin constants, and allowed for the constantness of a function result to be communicated back
- fixed issues with `eval_sql` (many of the builtin signal processing functions are implemented with it) which would crash when one eval_sql function was implemented with another.
- added a `concat()` and `slice()` function to manipulate embeddings
- created generic functionality to create an ExpressionValueInfo from a ValueDescription
- fixed several small bugs and incorrect type deductions along the way
- added the pffft FFT library as the best 1d FFT library based upon code size, compatible license and reasonable performance
- Made the `EmbeddingValueInfo` class derive from `RowValueInfo`, since embeddings are an implementation detail of rows
- Added a bunch of signal processing functions implemented on top of this framework
- Added documentation and testcases


sql/builtin_signal_functions.cc
```diff
@@ -0,0 +1,307 @@
...
+        type = PFFFT_COMPLEX;
+    else throw HttpReturnException(400, "FFT type must be either 'real' "
+                                   "or 'complex'; got '" + typeStr
+                                   + "'.");
+    
+
+
+    DimsVector dimsVector = args[0].getEmbeddingShape();
+
+    cerr << "fft " << directionStr << " " << typeStr << " " << dimsVector << endl;
```
steadtler: cerr's

sql/expression_value.h
```diff
@@ -1430,8 +1542,77 @@ struct AnyValueInfo: public ExpressionValueInfoT<ExpressionValue> {
...
+        return true;
+    }
+};
+
 /// For an embedding
-struct EmbeddingValueInfo: public ExpressionValueInfoT<ML::distribution<CellValue, std::vector<CellValue> > > {
+struct EmbeddingValueInfo: public RowValueInfo {
     EmbeddingValueInfo(StorageType storageType = ST_ATOM)
         : EmbeddingValueInfo(std::vector<ssize_t>({-1}), storageType)
     {
```
steadtler: Missing isEmbedding / couldBeEmbedding ?

testing/MLDB-1972-fft.js
```diff
@@ -0,0 +1,66 @@
+// This file is part of MLDB. Copyright 2016 Datacratic. All rights reserved.
+
```
steadtler: No checks?

sql/expression_value_conversions.cc
```diff
@@ -568,5 +568,541 @@ convertEmbeddingImpl(void * to,
...
+        defaultConstructBuffer((CellValue *)buffer, numElements);
+        return;
+    case ST_BOOL:
+        defaultConstructBuffer((bool *)buffer, numElements);
+        return;
+    case ST_TIMESTAMP:
+        defaultConstructBuffer((Date *)buffer, numElements);
+        return;
+    case ST_TIMEINTERVAL:
+        throw HttpReturnException(500, "Can't store time intervals");
```
steadtler: Why though?

jeremybarnes: We don't have a TimeInterval class that we can use as an underlying type, and it already wasn't supported.  I agree we should address this, but it's out of scope of this ticket.  (MLDB-1988)

testing/MLDB-1972-fft.js
jeremybarnes: Will add.

sql/expression_value.h
jeremybarnes: They are there.

PR #714, opened 2016-10-11
check that the data is actually dense in getColumnDense

steadtler:


core/dataset.cc
```diff
@@ -405,7 +405,15 @@ getColumnDense(const ColumnPath & column) const
     }
 
     for (auto& name : rowNames) {
-        result.push_back(values.find(name)->second.first);
+
+        auto iter = values.find(name);
+        if (iter == values.end())
+            throw MLDB::Exception(("Row " + name.toUtf8String() +
+                                   " does not have column " +
+                                     column.toUtf8String()).rawString() +
```
mailletf: in THE dense data index

PR #717, opened 2016-10-12
Fix some clang build errors

FinchPowers:
Remaining error with... TF.
```
./mldb/ext/tensorflow/tensorflow/cc/framework/scope.cc:25:8: error: constructor for 'tensorflow::Scope' must explicitly initialize the const member 'colocation_constraints_'
Scope::Scope(Graph* graph, Status* status, Scope::NameMap* name_map)
       ^
mldb/ext/tensorflow/tensorflow/cc/framework/scope.h:245:36: note: 'colocation_constraints_' declared here
  const std::unordered_set<string> colocation_constraints_;
```

### DIFF
--- a/sql/expression_value_conversions.cc
+++ b/sql/expression_value_conversions.cc
@@ -865,7 +865,7 @@ uint64_t storageBufferBytes(const DimsVector & size, StorageType storage)
     size_t sz = 1;
     for (auto & s: size)
         sz *= s;
-    return storageBufferBytes(size, storage);
+    return storageBufferBytes(sz, storage);
 }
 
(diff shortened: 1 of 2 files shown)


sql/expression_value_conversions.cc
```diff
@@ -865,7 +865,7 @@ uint64_t storageBufferBytes(const DimsVector & size, StorageType storage)
     size_t sz = 1;
     for (auto & s: size)
         sz *= s;
-    return storageBufferBytes(size, storage);
+    return storageBufferBytes(sz, storage);
```
FinchPowers: Thank you clang, that looked like an infinite loop. Is this ok now?

PR #730, opened 2016-10-22
Allow batch predictions (MLDB-2022)

jeremybarnes:
This adds the `/v1/functions/<functionname>/batch` endpoint for batch predictions, and generally cleans up the prediction interface to MLDB.

In order to enable an improved user experience, the user function infrastructure has been modified to be more like builtin functions.  In particular, inputs and outputs no longer need to be rows, and much of the work necessary for positional arguments has been done.

The `row_dataset` has been modified to return one row per *column*, and an `atom_dataset` construct added with semantics similar to the original.  The types of these datasets have been improved, with inference of the `value` type and the `column` type is now path, not string.

The `sql.expression` object has been improved to allow `raw` and `autoInput` parameters to be passed, bypassing the requirement for a row on output and input respectively.

Lots of test updates and documentation improvements have also been added.


testing/MLDB-2022-multiple-prediction-example.js
```diff
@@ -0,0 +1,127 @@
...
+assertEqual(fn.responseCode, 201);
+
+mldb.log(fn);
+
+var res = mldb.get('/v1/functions/score_one/batch',
+                   { input: [[1,2,3],[4,5],[6],[]] });
+
+mldb.log(res);
+
+var expected = [ 6, 9, 6, 0 ];
```
mailletf: the last 0 should be a null no?

container_files/public_html/doc/builtin/WorkingWithRest.md
```diff
@@ -23,3 +23,179 @@ Beyond that, if MLDB is running on a server named `<host>` on port `<port>` then
...
+which will return us
+
+````
+{ "one": 6, "two": 9, "three": 6, "four": 0 } 
+````
+
+with the function being applied to each member of the object.
+
+
+mldb.log(res);
```
FinchPowers: Coquille?

container_files/public_html/doc/builtin/sql/FromExpression.md
```diff
@@ -110,6 +120,42 @@ x       |  1
 y       |  2
 z       |  "three"
 
+whereas the following query
+
+```sql
+select * from row_dataset({w: {x: 1, y:1, y:2, z: 'three'}})
```
FinchPowers: Il y a un mélange de requêtes en SQL case et d'autres toutes en minuscules. Pas grave, mais inconstant.

plugins/sql_functions.cc
```diff
@@ -343,6 +346,22 @@ SqlExpressionFunctionConfigDescription()
              "This can lead to faster batch queries, at the expense of a "
              "possibly high per-query overhead for individual queries.",
              false);
+    addField("raw", &SqlExpressionFunctionConfig::raw,
+             "If true, then the output will be raw (just the result "
+             "of the expression will be returned; it will not be turned "
+             "into a row and the name of the output will be ignored).  If "
+             "false (defult), then the output will be structured into "
```
FinchPowers: defult -> default

sql/table_expression_operations.cc
```diff
@@ -645,15 +646,53 @@ bind(SqlBindingScope & context) const
...
+    
+    switch (style) {
+    case COLUMNS:
+        if (inputInfo->getSchemaCompleteness() == SCHEMA_CLOSED) {
+            // Try to infer a type for the column column
+            auto knownColumns = inputInfo->getKnownColumns();
+            if (!knownColumns.empty()) {
+                valueInfo = knownColumns[0].valueInfo;
+                for (size_t i = 1;  i < knownColumns.size();  ++i) {
+                    valueInfo = ExpressionValueInfo
```
FinchPowers: Unsure (I don't know what getCovering does.) Is it intended to loop to assign to `valueInfo` every time? Couldn't it just take the last value instead of looping?

sql/table_expression_operations.cc
```diff
@@ -645,15 +646,53 @@ bind(SqlBindingScope & context) const
...
+        break;
+    case ATOMS:
+        if (inputInfo->getSchemaCompleteness() == SCHEMA_CLOSED) {
+            // Try to infer a type for the column column
+            auto knownAtoms = inputInfo->getKnownAtoms();
+            if (!knownAtoms.empty()) {
+                valueInfo = knownAtoms[0].valueInfo;
+                for (size_t i = 1;  i < knownAtoms.size();  ++i) {
+                    valueInfo = ExpressionValueInfo
+                        ::getCovering(valueInfo, knownAtoms[i].valueInfo);
```
FinchPowers: Same here.

testing/MLDB-1012_nested_function_calls.py
```diff
@@ -79,14 +79,14 @@
...
-    "params": {"expression": "recurse({input})[output] as output"}
-})
-
 # MLDB-1251
 mldb.log("MLDB-1251")
 try:
+    mldb.put("/v1/functions/recurse", {
+        "type": "sql.expression",
+        "params": {"expression": "recurse({input})[output] as output"}
+    })
```
FinchPowers: This tested that the query failed. From the change I guess that it now tests that the put function fails instead. So it doesn't test the same thing.

testing/MLDB-2022-multiple-prediction-example.js
jeremybarnes: I thought so, but that's what it gives. We can revisit in another issue.

container_files/public_html/doc/builtin/sql/FromExpression.md
jeremybarnes: fixed

container_files/public_html/doc/builtin/WorkingWithRest.md
jeremybarnes: oups

plugins/sql_functions.cc
jeremybarnes: fixed

sql/table_expression_operations.cc
jeremybarnes: `getCovering` takes two types and returns a single type that can represent the values contained in either of them.  It's correctly used.  I added a comment.

testing/MLDB-1012_nested_function_calls.py
jeremybarnes: The test actually did check that the binding didn't cause an infinite loop; it's just that the binding now happens earlier than before, and so the test didn't test where the infinite loop was happening anymore.  It's testing that the same exception was thrown from the same function, so it's basically the same as before.

FinchPowers:
+1

container_files/public_html/doc/builtin/WorkingWithRest.md
```diff
@@ -23,3 +23,175 @@ Beyond that, if MLDB is running on a server named `<host>` on port `<port>` then
 ## Calling the API over HTTP from Python with `pymldb`
 
 If you are using the built-in [Notebook interface](Notebooks.md) or want to work with MLDB from Python, you can install [`pymldb`](Notebooks.md), which gives you access to an MLDB-specific library to interact with the API over HTTP, while hiding the details of HTTP from you. The ![](%%nblink _tutorials/Using pymldb Tutorial) will show you how to use `pymldb`.
+
+## Exposing functions are a REST API
```
guyd: are -> as

sql/table_expression_operations.cc
```diff
@@ -697,30 +738,60 @@ bind(SqlBindingScope & context) const
...
 
-                auto onAtom = [&] (const Path & columnName,
-                                   const Path & prefix,
-                                   const CellValue & val,
-                                   Date ts)
-                {
-                    NamedRowValue row;
-                    row.rowHash = row.rowName = ColumnPath(to_string(n++));
+                if (style == ATOMS) {
+                    cerr << "atom join on " << jsonEncode(row) << endl;
```
guyd: debug output

PR #731, opened 2016-10-23
Removed Id class (MLDB-2027)

jeremybarnes:
Removes the legacy Id class entirely from MLDB, reducing code size and getting rid of some history from the codebase.

The class has been moved to the enterprise version, where it is needed to maintain compatibility with legacy binary file formats.

Note that this will mean that stats tables and distribution tables created with previous versions of MLDB can no longer be loaded by this version of MLDB.


plugins/matrix.cc
```diff
@@ -411,9 +411,9 @@ invertFeatures(const ClassifiedColumns & columns,
                 //<< " sparse " << bucketSparseIndexes[i].size() << endl;
 
                 if (bucketDiscreteIndexes[i].size() > 0) {
-                    result[i + numContinuousColumns].discreteValues
-                        .insert(bucketDiscreteIndexes[i].begin(),
-                                bucketDiscreteIndexes[i].end());
+                    for (auto & rowNumber: bucketDiscreteIndexes[i])
```
mailletf: is this for this PR ? or comited by mistake in it?

server/credential_collection.cc
```diff
@@ -131,12 +132,19 @@ initRoutes(RouteManager & manager)
     manager.addDeleteRoute();
 }
 
+uint64_t highwayHash(const uint64_t* key, const char* bytes,
```
mailletf: no implementation?

plugins/matrix.cc
jeremybarnes: it's for this PR.  The implementation of Insert used Id.

PR #732, opened 2016-10-23
Implemented missing getColumnRowCount() in sparse.mutable to fix SVD test speed (MLDB-2028)

jeremybarnes:
Band-aid patch to fix the immediate test speed problem by adding a missing override and avoiding a very naive computation (iterating the whole dataset on every row to count the number of rows in each column in the row).  This should eventually be fixed at a deeper level by allowing the column selection in `COLUMN EXPR` to be cached when it's constant over the dataset.

Also some minor optimizations to avoid copying and excess lookups in the SVD setup code were added.

It takes the runtime for `MLDB-498-svd-apply-function.js` from 3 minutes to 4 seconds.

### DIFF
--- a/plugins/sparse_matrix_dataset.cc
+++ b/plugins/sparse_matrix_dataset.cc
@@ -706,6 +706,25 @@ struct SparseMatrixDataset::Itl
         return { columnName, std::make_shared<AnyValueInfo>(), COLUMN_IS_SPARSE };
     }
 
+    virtual uint64_t getColumnRowCount(const ColumnPath & column) const override
+    {
+        auto trans = getReadTransaction();
+
+        uint64_t result = 0;
+
+        uint64_t lastRow = -1;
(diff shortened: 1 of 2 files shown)


plugins/sparse_matrix_dataset.cc
```diff
@@ -706,6 +706,25 @@ struct SparseMatrixDataset::Itl
...
+    virtual uint64_t getColumnRowCount(const ColumnPath & column) const override
+    {
+        auto trans = getReadTransaction();
+
+        uint64_t result = 0;
+
+        uint64_t lastRow = -1;
+        auto onEntry = [&] (const BaseEntry & entry)
+            {
+                if (entry.rowcol != lastRow)
```
mailletf: why could this be the same??

guyd:
+1

PR #733, opened 2016-10-23
Removed boost_thread as dependency (MLDB-2029)

jeremybarnes:
... it's not needed anymore (we use the standard library)


FinchPowers:
+1

PR #736, opened 2016-10-26
adding trace level logging

guyd:


FinchPowers:
+1
docker test set to manual in master so we can ignore it here

PR #738, opened 2016-10-27
Added JSONTestSuite tests (MLDB-2032)

jeremybarnes:
This adds a large JSON test suite, and fixes the handling of UTF-16 surrogate pairs (which are the way that characters out of the basic multilingual plane are handled).  It also makes handling of extremely large numbers more robust (if they are too big for a long, they become a double, and if they are too big for a double, they become infinity rather than crashing as before).

The test failure is spurious.


FinchPowers:
Why is it not using BOOST for the tests as usual?
+1

PR #742, opened 2016-11-02
Cleaned up protobuf dependency interactions (MLDB-2042)

jeremybarnes:
This fixes up various protobuf dependency issues that were causing the docker builds to fail.  Most importantly, it avoids putting protobuf or tensorflow includes under $(INC), which changes from docker to non-docker builds, and hence allows for the work to be re-used.

Authors of plugins that use protobuf (I'm thinking @steadtler) will need to read the comments in protobuf.mk and adjust accordingly so that their plugins will continue to build.


steadtler:
+1, I'll keep in mind next time we update mldb in the plugin.

FinchPowers:
Malade! :) +1

PR #744, opened 2016-11-03
MLDB-2043 tabular big int

jeremybarnes:
Fixes problems with Bit_Extractor<uint64_t> causing data corruption in tabular datasets.  Drags along part of a refactoring of the frozen columns.


jeremybarnes:
(Test failure is spurious)

PR #753, opened 2016-11-25
Fix cuda

jraby:
Multiple tweaks to allow WITH_CUDA builds to work on x86_64.

### DIFF
--- a/arch/cuda.cc
+++ b/arch/cuda.cc
@@ -18,7 +18,7 @@
 using namespace std;
 
 
-namespace ML {
+namespace MLDB {
 
 namespace {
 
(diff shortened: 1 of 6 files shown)


jeremybarnes:
OK, but don't touch the generic arch options.  Add the extra includes into tensorflow.mk instead.

jraby:
I'll do that.

These changes were based on aarch64:

  - https://github.com/mldbai/mldb/blob/master/jml-build/arch/aarch64.mk#L3
  - https://github.com/mldbai/mldb/blob/master/jml-build/arch/aarch64.mk#L13

FinchPowers:
Moi aussi!



2016-11-28 11:02 GMT-05:00 jraby <notifications@github.com>:

PR #754, opened 2016-11-25
Make ARM build again (MLDB-2062)

jeremybarnes:
This fixes various small issues and allows a clean build for ARM using the following sequence, which matches the sequence documented in Building.mk:

```
nice make -j24 -k build_tools
nice make -j24 -k port_deps ARCH=arm
nice make -j24 -k mldb_runner mldb_plugins ARCH=arm
```

It also updates the port_deps process to install dependencies of the packages listed in ports.mk,
reducing the amount of work required to add a package (it's no longer necessary to specify each of its dependencies manually).

Finally it fixes build ordering of the protobuf include files to make sure that the Tensorflow build happens with everything in the right order.


guyd:
+1

PR #757, opened 2016-11-28
Another round of cuda mk fixes

jraby:


jraby:
Not needed.
cuda.mk is legacy and shouldn't be used for now.

PR #758, opened 2016-11-28
MLDB-1916 zstandard compression

jeremybarnes:
This adds the zstandard compression protocol (`.zst`) to MLDB for both compression and decompression.  That protocol dominates gzip in both speed and compression accuracy, which is especially helpful when performing streaming mobile data capture on a device with limited CPU capacity.

Rather than hard-coding like the others, this patch also makes compression protocols pluggable which should ease the development of further protocols in the future.

Finally, the full set of tests were re-enabled in `filter_streams_test` to ensure test coverage.


guyd:
reviewed briefly +1

PR #759, opened 2016-11-29
rm libarch_cuda and fix pb dependency for TF gpu files.

jraby:


jeremybarnes:
We should remove the source files in arch referred to by cuda.mk too
+1 once done

jraby:
I'm not sure I understand what you mean. `mldb/arch/cuda.mk` or `mldb/jml-build/cuda.mk` ?

Or do you simply want to rm LIBARCH_CUDA_SOURCES  (`arch/cuda.cc`)  ?

jeremybarnes:
+1

PR #760, opened 2016-11-30
MLDBFB-724 fix classifier segfault

FinchPowers:
Crash occurred when `boundDataset.dataset->config_` was nulptr.
May not be the proper fix.

### DIFF
--- a/plugins/classifier.cc
+++ b/plugins/classifier.cc
@@ -535,26 +535,30 @@ run(const ProcedureRunConfig & run,
 
     int nx = numRows;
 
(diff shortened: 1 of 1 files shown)


jeremybarnes:
This feels more like a bandaid.  Can you look quickly at why config_ wasn't set, and fix it there if not?

jeremybarnes:
Actually we should add a getId() function to MldbEntuty that will return sing kind of default if config is not set and use that instead of config_

FinchPowers:
Fixed in #770

PR #762, opened 2016-12-05
Explicit dependency on parent dir creation

jraby:
Should fix this error:

```
/bin/cp: cannot create regular file ‘build/x86_64/tmp/docker-mldb/opt/lib/date_timezone_spec.csv’: No such file or directory
```


jeremybarnes:
+1

PR #763, opened 2016-12-05
Mldb 2063 sparse mutable encode val path

guyd:


jeremybarnes:
I would consider encoding a path that's just a number (a *lot* of them are), ie where tryAsIndex() is not -1, as a separate encoding that doesn't pass through strings.

+1 apart from that.

guyd:
Makes perfect sense.  This is why I bumped the tag version to 30 because I thought that we might have path that require a different encoding.

PR #766, opened 2016-12-07
Mldb 2040 pipeline join refactor

steadtler:
Refactor both the equi and cross join pipeline executor to better handle left, right and full outer joins.

We cannot output a row as an outer row until its been tested against every single other row it could match with.
For left or right cross joins, we choose the proper side to loop on first to be efficient (for example if its a right join, we loop on right first). This way we avoid having to cache all rows. For a full cross join, we dont have a choice but to cache all rows on one side and remember if they've ever been output.

For equi joins, we already had a sliding window row cache for the left side, so we now also keep track of whether each row was ever matched, and selectively output an outer row then it gets removed from the cache. This will also help with memory in many cases, as the cache was not emptied until the end, while not its size will be kept at a minimum.

Finally, we use the left and right pipeline scopes to determine how many empty elements to add when we output outer rows.

### DIFF
--- a/sql/execution_pipeline_impl.cc
+++ b/sql/execution_pipeline_impl.cc
@@ -901,6 +901,12 @@ bind() const
 
 /*****************************************************************************/
 /* CROSS JOIN EXECUTOR                                                       */
+/* We cannot output a row as an outer row until its been tested against      */
+/* every single other row it could match with.                               */
+/* For left or right cross joins, we choose the proper side to loop          */
+/* on first to be efficient                                                  */
+/* (for example if its a right join, we loop on right first).                */
+/* This way we avoid having to cache all rows.                               */
 /*****************************************************************************/
     
 JoinElement::CrossJoinExecutor::
@@ -993,7 +999,7 @@ take()
 
                 return result;
             }
-	        wasOutput = false;
+            wasOutput = false;
             l = this->left->take();
         }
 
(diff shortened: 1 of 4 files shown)


sql/execution_pipeline_impl.h
```diff
@@ -339,18 +339,47 @@ struct JoinElement: public PipelineElement {
...
+        const Bound * parent;
+        std::shared_ptr<ElementExecutor> root, left, right;
+        virtual std::shared_ptr<PipelineResults> take();
+
+        std::shared_ptr<PipelineResults> r;
+        typedef std::list<std::pair<std::shared_ptr<PipelineResults>, bool > > bufferType;
+        bufferType bufferedLeftValues;
+        bufferType::iterator l;
+        bool firstSpin;
+        bool wasOutput;
```
guyd: It's not clear to me why we need `wasOutput` in addition to the flag in the buffer to tell if the row was outputted or not?

guyd:
You give a good explanation of the optimization with the left or right outer join in the description above, we should also have it in the code.

sql/execution_pipeline_impl.cc
```diff
@@ -922,113 +927,276 @@ std::shared_ptr<PipelineResults>
...
-            ExpressionValue where = rEmbedding.getColumn(1, GET_ALL);
-            if (!where.asBool()) {
+        if (!l || !r)
+            return nullptr;
+
+        // Got a row!       
+        ExpressionValue lEmbedding = l->values.back();
+        ExpressionValue rEmbedding = r->values.back();
+
+	    auto result = l;
```
guyd: Alignment is wrong in this area too.

sql/execution_pipeline_impl.h
guyd: Got it after discussion with Mathieu.  I suggest we rename the value to `rightRowWasOutputted`.

guyd:
+1 with very minor comment about alignment and missing comments

PR #767, opened 2016-12-07
Cross compilation fixes

jraby:
- Use `-o APT::Architecture=$ARCH` in apt cmds instead of relying on `dpkg --add-architecture`
- Doc update
- Tweak test to avoid `-Werror=overflow` in a test which does an overflow on purpose.


guyd:
Ca me semble bon.  +1

PR #768, opened 2016-12-07
Mldb 2074 empty outer join

steadtler:
Create tests and fix issues for full outer joins in the pipeline executor when one of the dataset is empty.


guyd:
reviewed briefly +1

PR #769, opened 2016-12-07
MLDB-2077 allow merge() function to accept a single dataset

steadtler:
Note that single-dataset merged datasets were already accepted through a config instead of the builtin.


FinchPowers:
Test case missing. I'll take care of it thanks.

PR #772, opened 2016-12-09
Build for multiple CUDA architectures (MLDB-2085)

jeremybarnes:
Before this patch, the Tensorflow plugin needed to be rebuilt for each GPU architecture.  Now it should work out of the box on basically any deployed server architecture (see list below).

This enables multiple architectures to be built for CUDA, and makes a default list that covers the vast majority of deployed server CUDA architectures:

```
   Compute/shader model   Cards
   6.1			  P4, P40, Titan X
   6.0                    P100
   5.2                    M40
   3.7                    K80
   3.5                    K40, K20
   3.0                    K10, Grid K520 (AWS G2)
```

It's a Makefile-only fix, and has basically no effect unless `WITH_CUDA=1` is passed.  Passes all tests on a build with that flag set.


jraby:
LGTM  +1

PR #773, opened 2016-12-09
Mldb 1935 const binding

steadtler:
Add a test and fix const static analysis for everything expect expressions related to function calls (to be followed)


guyd:
+1 good test coverage

PR #774, opened 2016-12-09
MLDB-2088 tensorflow test gpu support

jeremybarnes:
This adds a test for running a Tensorflow graph on a GPU, enabled when WITH_CUDA=1 and the machine has GPUS.

To enable the functionality, we also add a device affinity configuration to the Tensorflow graph.

Finally, the `sessionsPerDevice`, which used to need to be high to get good throughput, now causes a crash when > 1, and so it's set back to 1.

This passes on a GPU machine with WITH_CUDA=1.


jraby:
I like the general idea.

Isn't that test very similar to MLDB-1203? Is there a way to pass an argument or something so that we don't duplicate the whole file just to add a single flag?

Also, I do not think we should enable that test since it relies on having `inception_dec_2015.zip` in the cwd.

If we want to enable it, I'd put inception_dec_2015.zip on public.mldb.ai and fetch it on every test. (we have a caching proxy for that use case)
If that file has an appropriate license, of course.

Same goes for all the other image fetches this test does. We should host a copy with our other testing data. (So that failure to fetch those are not dependent on a third party).

If you agree, I'll do the file copying and test fixing.

tensorflow/MLDB-2088-tensorflow-GPU-support.js
```diff
@@ -0,0 +1,228 @@
...
+// and return it as a blob.
+var fetcherConfig = {
+    id: 'fetch',
+    type: 'fetcher',
+    params: {
+    }
+};
+var fetcher = mldb.createFunction(fetcherConfig);
+
+//var res = mldb.query("SELECT tf_DecodeJpeg(fetch({url: 'file://mldb/ext/tensorflow/tensorflow/examples/label_image/data/grace_hopper.jpg'})[content], {ratio: 8}) as px");
```
steadtler: Did you mean to comment this? The log below will display the result of the previous query?

jeremybarnes:
OK with me.

May want to consider merging first(since in effect no CI machine has GPUs and so the new test will always be manual) as the dod is meet and adding enhancements to address those points.

jraby:
I'll have a look, I would very much like to avoid importing a new test that is the same as another test, but with 1 line change.

Especially when the original test is full of debug and commented out / dead code.

Do we have a way to pass flags to a test?
Or maybe the test could inspect argv[0] to see under what name it is called and enable the cpu test based on that?

simlmx:
FWIW using this branch and changing 
TF_CUDA_CAPABILITIES=3.0
to
TF_CUDA_CAPABILITIES=6.1
I was able to use cpus and gpus!

jeremybarnes:
I'm fine with whatever way makes sense to move this forwards.

jraby:
So, I've tweaked the branch a little...

  - `TF_CUDA_CAPABILITIES` is now 6.0 and overridable at build time
  - `MLDB-1203-tensorflow-plugin.js` has been cleaned a bit.
     It will now run on both CPU _and_ GPU if  `args.GPUS >= 1`
  - The test assets (inception zip and test image) are now hosted on `public.mldb.ai`, so the test has been enabled by default.
  - If we're building with `WITH_CUDA=1` and if there are GPUs in the build machine, the test will run on the CPU _and_ GPU.

Along the way I had to add support for passing arguments to `mldb_unit_test` test cases.

ext/tensorflow.mk
```diff
@@ -142,7 +142,7 @@ CUDA_MODELS=30 35 37 52 60 61
 
 # Which CUDA compute capabilities do we support?  3.0 is needed for the
```
guyd: The comment should be updated accordingly.  Is 6.0 the minimum supported?

mldb_macros.mk
```diff
@@ -5,9 +5,13 @@
 #      - manual: run the test manually
 #      - virtualenv: set up the Python virtualenv even for a non-python test
 #      - valgrind: run the test within valgrind
+# $(4) TODO: used but undocumented
+# $(5) Passed to --script-args if defined
```
guyd: Probably this should read `Passed to mldb_runner as a --script-args`.

tensorflow/MLDB-1203-tensorflow-plugin.js
```diff
@@ -1,20 +1,25 @@
...
 // https://storage.googleapis.com/download.tensorflow.org/models/inception_dec_2015.zip
+// Hosted on public.mldb.ai for caching.
+//var inceptionUrl = 'file://inception_dec_2015.zip';  // Use this to test with a local copy
+var inceptionUrl = 'https://public.mldb.ai/models/inception_dec_2015.zip';
+var imgUrl = "http://public.mldb.ai/testing/tensorflow/Calle_E_Monroe_St,_Chicago,_Illinois,_Estados_Unidos,_2012-10-20,_DD_04.jpg";
+var imgUrlSql = mldb.sqlEscape(imgUrl);
+
+// GPU testing
+var devices = [ "/cpu:.*" ];
+if (args && args.GPUS && args.GPUS >= 1) {
```
guyd: Cool!

tensorflow/tensorflow.mk
```diff
@@ -1,5 +1,12 @@
 # Makefile for tensorflow plugin for MLDB
 
+ifeq ($(WITH_CUDA),1)
+NUM_CUDA_GPUS:=$(shell nvidia-smi -L | grep 'GPU ' | wc -l)
+else
+NUM_CUDA_GPUS:=0
+endif
+MANUAL_IF_NO_GPUS:=$(if $(call seq,0,$(NUM_CUDA_GPUS)),manual)
```
guyd: This is no longer needed, right?

ext/tensorflow.mk
jraby: 6.0 is not the minimum supported, but it is what we use for our hardware.

mldb_macros.mk
jraby: will fix

tensorflow/tensorflow.mk
jraby: It is not needed, but I like the idea of being able to disable a test if we do not have cuda or if we don't have any GPUs.
So I left it in.

ext/tensorflow.mk
jraby: I'll revert that to 3.0 since the tests work fine with it and using something higher would break compat with AWS g2 instances.

PR #775, opened 2016-12-09
[MLDB-2086] passing the config to mldb_runner in debug and release

guyd:
- passing the mldb.conf when doing make <test> to allow logging
- storing the mldb.conf in the docker image and passing it to mldb_runner
- fix a crash when failing to parse the command line

### DIFF
--- a/container_files/init/mldb_runner.sh
+++ b/container_files/init/mldb_runner.sh
@@ -20,5 +20,6 @@ exec /sbin/setuser _mldb \
         --plugin-directory {{MLDB_GLOBAL_PLUGINS_PATH}} \
         --plugin-directory {{MLDB_LOCAL_PLUGINS_PATH}} \
         --http-base-url "${HTTP_BASE_URL}" \
+        --config-path=/etc/mldb.conf \
         {{MLDB_EXTRA_FLAGS}} \
         $MLDB_RUNNER_ARGS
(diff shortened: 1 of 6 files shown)


templated_files.mk
```diff
@@ -10,6 +10,7 @@ $(eval $(call mldb_install_templated_directory,mldb/container_files/public_html/
 
 $(eval $(call mldb_install_templated_file,mldb/container_files/init/05-mldb-id-mapping.sh,$(ETC)/my_init.d/05-mldb-id-mapping.sh,555))
 $(eval $(call mldb_install_templated_file,mldb/container_files/init/mldb_runner.sh,$(ETC)/service/mldb_runner/run,555))
+$(eval $(call mldb_install_templated_file,mldb/container_files/mldb.conf,$(ETC)/service/mldb_runner/run,664))
```
jraby: I think we could put `mldb.conf`  directly in `/etc`  instead of deep inside the `runit` directory.

Minor nit:  0644.

jraby:
I'm not sure if we should overload `MLDB_EXTRA_FLAGS` for this purpose.
`mldb.conf` might deserve its own template variable since we probably want to pass it all the time.

I see `EXTRA_FLAGS` as the hook which can be used to pass image specific parameters.
With this change, we would have to remember to set the config file along with the extra flags we want to pass.

I think we could add `--config-path=$(MLDB_CONFIG_FILE)` to `mldb_runner.sh` and set the path directly in `TEST_$(1)_RAW_COMMAND`  (or change MLDB_CONFIG_FILE when doing a debug build)

container_files/init/mldb_runner.sh
```diff
@@ -20,5 +20,6 @@ exec /sbin/setuser _mldb \
         --plugin-directory {{MLDB_GLOBAL_PLUGINS_PATH}} \
         --plugin-directory {{MLDB_LOCAL_PLUGINS_PATH}} \
         --http-base-url "${HTTP_BASE_URL}" \
+        --config-path=/etc/mldb.conf \
```
guyd: @jraby  Est-ce que je devrais utiliser une variable d'environnement ici?

PR #776, opened 2016-12-13
Mldb 1935 function static analysis

steadtler:
Fix const static-analysis for function calls, extract and like expressions.

### DIFF
--- a/rest/poly_collection.cc
+++ b/rest/poly_collection.cc
@@ -30,7 +30,8 @@ bool operator==(const PolyConfig & lhs, const PolyConfig & rhs)
     return  lhs.id == rhs.id &&
         lhs.type == rhs.type &&
         lhs.persistent == rhs.persistent &&
-        lhs.params == rhs.params;
+        lhs.params == rhs.params &&
+        lhs.deterministic == rhs.deterministic;
 }
 
 DEFINE_STRUCTURE_DESCRIPTION(PolyConfig);
@@ -51,6 +52,8 @@ PolyConfigDescription()
     addField("persistent", &PolyConfig::persistent,
              "If true, then this element will have its configuration stored "
              "and will be reloaded on startup", false);
+    addField("deterministic", &PolyConfig::deterministic,
+             "If true, the entity has no hidden state ", true);
 }
 
 
--- a/sql/builtin_functions.h
+++ b/sql/builtin_functions.h
@@ -150,12 +150,12 @@ struct RegisterBuiltin {
                         }
                     };
 
-                    bool constantArgs = determinism == DETERMINISTIC;
+                    bool constantArgs = true;
                     for (auto& arg : args) {
                         constantArgs = constantArgs && arg.metadata.isConstant;
                     }
 
-                    result.resultMetadata.isConstant = constantArgs;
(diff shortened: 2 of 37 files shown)


sql/builtin_functions.h
```diff
@@ -137,6 +150,13 @@ struct RegisterBuiltin {
                         }
                     };
 
+                    bool constantArgs = determinism == DETERMINISTIC;
+                    for (auto& arg : args) {
+                        constantArgs = constantArgs && arg.metadata.isConstant;
+                    }
+
+                    result.resultMetadata.isConstant = constantArgs;
```
guyd: Here also, I would put `constantArgs && (determinism == DETERMINISTIC)` and remove it from the above.

sql/sql_expression_operations.cc
```diff
@@ -1520,6 +1520,8 @@ BoundSqlExpression
 ReadColumnExpression::
 bind(SqlBindingScope & scope) const
 {
+    cerr << "ReadColumnExpression " << columnName << endl;
```
guyd: debug

sql/sql_expression_operations.cc
```diff
@@ -2427,12 +2429,28 @@ ExtractExpression::
...
     SqlExpressionExtractScope extractScope(outerScope, fromBound.info);
 
     BoundSqlExpression extractBound = extract->bind(extractScope);
 
-    return {[=] (const SqlRowScope & row,		
+    bool isConst = fromBound.metadata.isConstant;
+
+    //We know for sure the extract is const if it access no function, tables or wildcards *and*
+    //all the variables are in the (const) lhs
+    //This exclude many legit const cases.
```
guyd: exclude -> excludes

testing/MLDB-1935-const-binding.py
```diff
@@ -348,7 +349,84 @@ def test_const_wildcard_var(self):
...
+        ])
+
+    def test_const_extract_const(self):
+        res = mldb.query("SELECT __isconst(({2 as x})[x]) as isconst FROM ds1 ORDER BY rowName()")
+        self.assertTableResultEquals(res, [
+            ['_rowName', 'isconst',],
+            ['row1', True],
+        ])
+
+    @unittest.expectedFailure
```
guyd: This should work isn't it?  If so, remove the expectedFailure.

rest/poly_collection.cc
```diff
@@ -51,6 +52,8 @@ PolyConfigDescription()
     addField("persistent", &PolyConfig::persistent,
              "If true, then this element will have its configuration stored "
              "and will be reloaded on startup", false);
+    addField("deterministic", &PolyConfig::deterministic,
+             "If true, the entity has no hidden state ", true);
```
steadtler: This one is for all entities, FunctionPolyConfigDescription is specifically for functions, and functions *per definition* are not supposed to have internal states :)

testing/MLDB-1935-const-binding.py
steadtler: Does not work yet unfortunately, because we dont (yet) have the const metadata per column of a rowValueInfo. Thats my next refactor.

PR #777, opened 2016-12-14
MLDB-1935 merge bound expression metadata inside ExpressionValueInfo

steadtler:


sql/expression_value.h
```diff
@@ -344,6 +344,12 @@ struct ExpressionValueInfo {
         return true;
     }
 
+    // Does it always returns the same value
+    virtual bool isConst() const = 0;
+
+    //return an expression value info of the same type, with the assigned constness
+    virtual std::shared_ptr<ExpressionValueInfo> getConst(bool constant) const = 0;
```
guyd: Why not a setConst member instead?

sql/expression_value.h
```diff
@@ -1187,7 +1193,7 @@ struct ExpressionValue {
     /** Return the most specialized possible value info for this given value.
         Used for static analysis of constants.
     */
-    std::shared_ptr<ExpressionValueInfo> getSpecializedValueInfo() const;
+    std::shared_ptr<ExpressionValueInfo> getSpecializedValueInfo(bool constant) const;
```
guyd: It is strange that we have to pass the constness here?  Don't we already know if the `this` is const or not?

testing/MLDB-1984-constant-functions.js
```diff
@@ -14,7 +14,7 @@ function assertEqual(expr, val, msg)
 var query = "SELECT tokenize('a,b,c') AS *";
 var tokQuery = "SELECT tokenize('a,b,c') AS tok";
 
-var analysis = mldb.get("/v1/query", { q: "SELECT static_expression_info(pi())[metadata][isConstant] as isRow", format: 'table', headers: false, rowNames: false });
+var analysis = mldb.get("/v1/query", { q: "SELECT static_expression_info(pi())[\"info\"][isConstant] as isRow", format: 'table', headers: false, rowNames: false });
```
guyd: Have you created a bug for that parsing issue?  If so, I would mention it here.

server/dataset_context.cc
```diff
@@ -75,10 +75,10 @@ doGetFunction(const Utf8String & tableName,
                     }
                 };
 
-            BoundFunction boundFunction(exec, applier->info.output);
-            boundFunction.resultMetadata.isConstant = constantArgs && applier->info.deterministic;
+            bool isConst = constantArgs && applier->info.deterministic;
+            auto outputInfo = applier->info.output->getConst(isConst);
```
guyd: I am not sure why we have a getConst function.  I would set constness in constructor and have a setConst member function, that is:
```
applier->info.output.setConst(constantArgs && applier->info.deterministic);
return BoundFunction(exec, applier->info.output)
```

sql/builtin_functions.cc
```diff
@@ -3329,7 +3329,7 @@ bind(const std::vector<BoundSqlExpression> & args,
                 }
             };
 
-        result.resultInfo = std::move(bound.info);
+        result.resultInfo = bound.info->getConst(isConstant && bound.info->isConst());
```
guyd: Why do we need to do set constness after the creation of the BoundSqlExpression?  Why can this call
```
 BoundSqlExpression bound = parsed->bind(evalScope);
```
return the info with the constness already set?

sql/builtin_functions.h
```diff
@@ -152,10 +153,10 @@ struct RegisterBuiltin {
 
                     bool constantArgs = true;
                     for (auto& arg : args) {
-                        constantArgs = constantArgs && arg.metadata.isConstant;
+                        constantArgs = constantArgs && arg.info->isConst();
                     }
 
-                    result.resultMetadata.isConstant = constantArgs && determinism == DETERMINISTIC;
+                    result.resultInfo = result.resultInfo->getConst(constantArgs && determinism == DETERMINISTIC);
```
guyd: If I understand correctly, this is done here to prevent checking for constness in every single builtin?  Don't we have cases where a builtin function would be deterministic despite the non-constness of its arguments?

sql/eval_sql.cc
```diff
@@ -50,7 +50,7 @@ evalSql(SqlBindingScope & scope,
     std::vector<std::shared_ptr<ExpressionValueInfo> > info;
     info.reserve(numArgs);
     for (size_t i = 0;  i < numArgs;  ++i) {
-        info.emplace_back(argsVec[i].getSpecializedValueInfo());
+        info.emplace_back(argsVec[i].getSpecializedValueInfo(false /*isconst TODO not sure about this one*/));
```
guyd: I suggest we track with a JIRA and mention the number here.

sql/expression_value.h
steadtler: This is a member of ExpressionValue, nor ExpressionValueInfo.

testing/MLDB-1984-constant-functions.js
steadtler: I believe it acts as intended? if it finds the keyword "inf" it expects a space, parentesis, comma or such afterwards. The quotes are meant to differentiate between the keywords and the identifiers

sql/builtin_functions.h
steadtler: If we find one we can see if its worth tackling it as an optimization. Being falsely non-const has no impact other than performance.

sql/expression_value.h
guyd: I still don't fully understand why we need a clone and why the constness should not be evaluated and set when the expression value info is created but I don't want to hold the PR for that.  We might want to rename `getConst` to `cloneWithConst` or simply `clone` to avoid the confusion with `isConst`.

testing/MLDB-1984-constant-functions.js
guyd: Agreed!

sql/builtin_functions.h
guyd: Makes sense

guyd:
+1

PR #783, opened 2016-12-20
MLDB-2100 http arbitrary slow timeout

FinchPowers:


plugins/fetcher.cc
```diff
@@ -93,7 +93,9 @@ struct FetcherFunction: public ValueFunctionT<FetcherArgs, FetcherOutput> {
         FetcherOutput result;
         Utf8String url = args.url;
         try {
-            filter_istream stream(url.rawString(), { { "mapped", "true" } });
+                filter_istream stream(url.rawString(),
+                                      { { "mapped", "true" },
+                                        { "httpArbitraryTooSlowAbort", "1"} });
```
guyd: Why not "true".  Also, filter_istream.h has a list of the _known_ options.  You should add yours there.

http/http_rest_proxy.cc
```diff
@@ -306,7 +308,8 @@ perform(const std::string & verb,
         bool exceptions,
         OnData onData,
         OnHeader onHeader,
-        bool followRedirect) const
+        bool followRedirect,
+        bool arbitraryTooSlowAbort) const
```
guyd: Since this is a bool, I think that `abortOnSlowConnection` is more appropriate.

guyd: Also, we should use the same name everywhere, some are prefixed with http some are not.

guyd:
+1 beside a few minor comment.  Would it be possible to add a test as well?

http/http_rest_proxy.cc
FinchPowers: dans le "http_rest" mettre le préfixe "http" me semble su
superflu.

guyd: C'est juste que ca aide a faire des greps quand on cherche comment un parametre est utilise.

jraby: +1 for `abortOnSlowConnection`

plugins/fetcher.cc
FinchPowers: Où est cette liste?

FinchPowers:
Je regarde pour le test.

plugins/fetcher.cc
guyd: https://github.com/mldbai/mldb/blob/master/vfs/filter_streams.h#L208

http/http_rest_proxy.cc
guyd: yup

FinchPowers:
@guyd le test est écrit.

guyd:
+1 pour le test

testing/MLDB-2100_fetcher_timeout_test.py
```diff
@@ -0,0 +1,65 @@
...
+# accept connections from outside
+try:
+    (clientsocket, address) = serversocket.accept()
+except socket.timeout:
+    mldb.log("MLDB did not contact the socket")
+    raise
+
+# now do something with the clientsocket
+# in this case, we'll pretend this is a threaded server
+ct = client_thread(clientsocket)
```
FinchPowers: Python is the server, MLDB is the client.
From the python server point of view, this is the socket to the client and the thread that manages the client.

I copied python doc without paying too much attention to the variable names and comments.

PR #784, opened 2016-12-20
Crank LOGBUFSIZE from 1024 to 8192 lines

jraby:
Also remove it from the 'template file' since it is not templated.


FinchPowers:
+1 ding dong!

PR #785, opened 2016-12-20
Mldb 2095 boosting and stump

guyd:


jeremybarnes:
all_features should only return used features, not trained features

guyd:
@jeremybarnes I get it for the all_features.  I will update this PR since it has staled.  I will submit for review in the coming day.

PR #787, opened 2017-01-04
Band aid: MLDB-2065 prevents segfault on nullptr over transposed ds

FinchPowers:
At least it prevents MLDB to segfault.


steadtler:
You can close that one without merging since we have a proper fix now.

PR #800, opened 2017-01-10
MLDB-2109 serial proc progress enh

FinchPowers:


core/procedure.cc
```diff
@@ -317,8 +319,32 @@ run(const ProcedureRunConfig & run,
...
+        return onProgress(progress);
+    };
+
+    for (int i = 0; i < steps.size(); ++i ) {
+        auto & s = steps[i];
+        iterationStep->value = (float)i / steps.size() * 100;
+
+        bool keepGoing = onProg(Json::Value{});
+        if (!keepGoing) {
+                throw MLDB::CancellationException("Procedure mock cancelled");
```
guyd: Error message should be changed.

core/procedure.cc
```diff
@@ -317,8 +319,32 @@ run(const ProcedureRunConfig & run,
...
+            progress["subProgress"] = data;
+        }
+        return onProgress(progress);
+    };
+
+    for (int i = 0; i < steps.size(); ++i ) {
+        auto & s = steps[i];
+        iterationStep->value = (float)i / steps.size() * 100;
+
+        bool keepGoing = onProg(Json::Value{});
```
guyd: Fishy.  Usually, the progress payload should cumulate the progress information from all steps.

testing/MLDB-621_serial_procedure_deadlock.py
```diff
@@ -1,16 +0,0 @@
-#
-# MLDB-621_serial_procedure_deadlock.py
```
guyd: Why removing it?  Isn't still detecting that recursive references aren't deadlocking?

testing/serial_procedure_test.py
```diff
@@ -0,0 +1,51 @@
+#
+# serial_procedure_test.py
+# datacratic, 2015
```
guyd: 2015->2017

guyd:
+1 once the concerns have been reviewed

testing/serial_procedure_test.py
FinchPowers: It's the original creation of the original file. I'm unsure whether I should bump it or not.

core/procedure.cc
FinchPowers: How do we do that? The problem is that the "progress" is lost whenever a procedure finishes.

guyd: Above serial progress should contain a progress step of each serial step (e.g. using the serial step id).  Pass the index `i` as part of onProg param (you will need to bind it before calling `run` on the serial step).  Whenever, `i` is greater than the last value, increment the serial progress with `next`.

core/procedure.cc
```diff
@@ -317,10 +319,37 @@ run(const ProcedureRunConfig & run,
...
+        Json::Value progress = jsonEncode(serialProgress);
+        if (!data.empty()) {
+            progress["subProgress"] = data;
+        }
+        return onProgress(progress);
+    };
+
+    for (int i = 0; i < steps.size(); ++i ) {
+        auto & s = steps[i];
+        //iterationStep->value = (float)i / steps.size() * 100;
```
guyd: Commented code.

PR #805, opened 2017-01-12
MLDB-2112 500 to 404 when not an error

FinchPowers:


FinchPowers:
On hold. The "resource" should not even be created if it originally returned a 400.

guyd:
+1

PR #806, opened 2017-01-13
Mdlb 2110 merged dataset progress

guyd:
Added progress at the construction of datasets.  In particular, sub selects and merge are now returning their progress.  I changed the progress interface along the way.  As a result, simpler operations like the construction of a dataset or its iteration are returning a simple progress object `ProgressState`.  Other more complex operations like running a procedure are using the previous interface that returns a Json object.


server/bound_queries.cc
```diff
@@ -164,7 +164,8 @@ struct UnorderedExecutor: public BoundSelectQuery::Executor {
 
         // Get a list of rows that we run over
         // Ordering is arbitrary but deterministic
-        auto rows = whereGenerator(-1, Any(), BoundParameters(), onProgress).first;
+        //TODO MLDB-2110 put back progress
```
FinchPowers: Forgotten TODO?

sql/builtin_dataset_functions.cc
```diff
@@ -127,13 +131,14 @@ BoundTableExpression merge(const SqlBindingScope & context,
             datasets.push_back(arg.dataset);
         }
         else if (!!arg.table) {
+            //TODO - MLDB-2110 split the progress for each args
```
FinchPowers: Forgotten?

testing/MLDB-2110-merge-and-subselect-progress.py
```diff
@@ -0,0 +1,61 @@
+# MLDB-2110-merge-and-subselect-progress.py
+# Datacratic, 2017
```
FinchPowers: Guy Dumais

Did you use the script generator? It should have put your name there, otherwise it's a bug.

utils/progress.h
```diff
@@ -41,11 +43,30 @@ struct Progress {
...
 };
-DECLARE_STRUCTURE_DESCRIPTION(IterationProgress);
+
+DECLARE_STRUCTURE_DESCRIPTION(ProgressState);
+
+typedef std::function<bool(const ProgressState &)> ProgressFunc;
+
+/* This is a temporary conversion helper to avoid 
+   changing all the procedure run signature.
+   TODO - MLDB-2110 - fix all the progress signature */
```
FinchPowers: TODO!!

testing/MLDB-2110-merge-and-subselect-progress.py
guyd: I did not for this one.  I started from the script you provided.

guyd:
The review was getting too large so I decided to split some of the work.  For example, I did not want to change all the progress signatures.  This may be done later as a separate JIRA.  I will create it.

server/bound_queries.cc
guyd: This one I forgot indeed.  Good catch.

PR #807, opened 2017-01-17
Allow customization of Eigen kernel launch failure behavior (MLDB-2115)

jeremybarnes:
Modifies the default behaviour from assert() to throw so that a CUDA launch failure is recoverable.  This requires a change to the Eigen library too (behind the submodule).


guyd:
+1

PR #812, opened 2017-01-18
How to regen LIBGIT2_SOURCE for new libgit2

jraby:


jraby:
trivial, merging away.

PR #813, opened 2017-01-19
Lower curl connect(2) timeout from 300s to 20s

jraby:
This allows for ~3 SYN retransmits on a default linux config.

The idea is to avoid being stuck in connect(2) for too long while still
having a chance of success when going through flaky networks.

The timeout value is not configurable, but I think it is a "saner default" than 5 minutes.
(and you should probably fix your network if you need a higher value).

For the record, nginx's proxy_pass module uses a 60s connect timeout.

### DIFF
--- a/http/http_rest_proxy.cc
+++ b/http/http_rest_proxy.cc
@@ -336,7 +336,9 @@ perform(const std::string & verb,
         if (timeout != -1)
             myRequest.add_option(CURLOPT_TIMEOUT, timeout);
         else myRequest.add_option(CURLOPT_TIMEOUT, 0L);
+
         myRequest.add_option(CURLOPT_NOSIGNAL, 1L);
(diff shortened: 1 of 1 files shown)


guyd:
Je pense que la communication a S3 utilise HttpClient qui doit souffrir du meme probleme.  Sinon +1

guyd:
Oups! j'ai ferme le PR!

jeremybarnes:
I would like to make it configurable but it's OK for now.  Clearly depending on the application we would want to affect the timeout and retry behavior

jraby:
I agree that the retries should be configurable, but I'm not event sure if curl treats connect timeouts as transient/retry-able errors. (or if it need to be handled at a higher level).

I'm not sure a see a case for raising the connect timeout, but I certainly see why we'd like to make it even lower for resources that are close. connections to local http servers should be able to fail really fast if the server is gone.

PR #815, opened 2017-01-19
Avoid dependency on libhttp in libtensorflow-core

jraby:
Use a runtime_error instead of HttpReturnException

### DIFF
--- a/ext/tf_cuda_handlers.h
+++ b/ext/tf_cuda_handlers.h
@@ -1,0 +1,10 @@
+/* tf_cuda_handlers.h
+ * Jean Raby, January 2017
+ * Copyright (c) 2017 mldb.ai Inc. All rights reserved.
+ */
+
+#include <stdexcept>
+
+struct CudaLaunchException: std::runtime_error {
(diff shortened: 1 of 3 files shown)


ext/tf_cuda_handlers.h
```diff
@@ -0,0 +1,10 @@
+/* tf_cuda_handlers.h
+ * Jean Raby, January 2017
+ * Copyright (c) 2017 mldb.ai Inc. All rights reserved.
```
guyd: mldb.ai Inc -> mldb.ai inc

ext/tf_cuda_handlers.h
```diff
@@ -0,0 +1,10 @@
+/* tf_cuda_handlers.h
+ * Jean Raby, January 2017
+ * Copyright (c) 2017 mldb.ai inc. All rights reserved.
+ */
+
+#include <stdexcept>
+
+struct CudaLaunchException: std::runtime_error {
+    CudaLaunchException(std::string err): std::runtime_error(err){}
```
guyd: const std::string & err

PR #823, opened 2017-01-23
Mldb 2114 plugin post nodata 404

FinchPowers:


testing/mldb_python_plugin_test.cc
```diff
@@ -44,13 +44,13 @@ BOOST_AUTO_TEST_CASE( test_python_loading )
     scriptConfig["address"] = "file://mldb/testing/python_script_test1.py";
 
     auto output = proxy.post("/v1/types/plugins/python/routes/run", scriptConfig);
-    BOOST_CHECK_EQUAL(output.code(), 200);
+    BOOST_REQUIRE_EQUAL(output.code(), 200);
```
FinchPowers: I've updated BOOST_CHECK to BOOST_REQUIRE, which makes it easier to debug errors. The downside is that it makes a lot of noise here. I'll highlight the lines I did "really" change for that PR.

testing/mldb_python_plugin_test.cc
```diff
@@ -247,45 +247,45 @@ print "a"
...
-    BOOST_CHECK_EQUAL(getResult.code(), 200);
-    BOOST_CHECK(getResult.jsonBody().isArray());
+    BOOST_REQUIRE_EQUAL(getResult.code(), 200);
+    BOOST_REQUIRE(getResult.jsonBody().isArray());
     
     getResult = proxy.get("/v1/plugins/pyplugin/routes/emptyDict");
     cerr << "getResult = " << getResult << endl;
-    BOOST_CHECK_EQUAL(getResult.code(), 404);
-//     BOOST_CHECK(getResult.jsonBody().isObject());
+    BOOST_REQUIRE_EQUAL(getResult.code(), 200);
```
FinchPowers: Here. An empty dict previously meant a 404. Now it's a valid 200 with the empty dict.

testing/mldb_python_plugin_test.cc
```diff
@@ -348,20 +348,20 @@ print "a"
...
 
-    BOOST_CHECK_EQUAL(getResult.code(), 200);
-    BOOST_CHECK_EQUAL(getResult.jsonBody()["value"].asString(), "in function");
+    BOOST_REQUIRE_EQUAL(getResult.code(), 200);
+    BOOST_REQUIRE_EQUAL(getResult.jsonBody()["value"].asString(), "in function");
 
     getResult = proxy.get("/v1/plugins/pyplugin_local_multifile/routes/doc");
     cerr << getResult << endl;
-    BOOST_CHECK_EQUAL(getResult.code(), 404);
+    BOOST_REQUIRE_EQUAL(getResult.code(), 500);
```
FinchPowers: 404 must now be set by the plugin routes in their handler. Failing to set a return code is a 500 error.

plugins/lang/python/python_converters.cc
```diff
@@ -246,6 +246,7 @@ construct_recur(PyObject * pyObj)
         val = construct_lst<bp::list>(pyObj);
     }
     else if(PyDict_Check(pyObj)) {
+        val = Json::objectValue;
```
FinchPowers: An empty dict is no longer a null, it's ... an empty dict.

jeremybarnes:
+1

PR #828, opened 2017-01-25
Mldb 2120 procedure deadlock test

FinchPowers:


arch/rcu_protected.h
```diff
@@ -199,8 +199,9 @@ struct RcuProtected {
         return std::unique_ptr<T>(val.exchange(newVal));
     }
     
+    template<typename DELETER>
     bool cmp_xchg(RcuLocked<T> & current,
-                  std::unique_ptr<T> & newValue,
+                  std::unique_ptr<T, DELETER> & newValue,
```
FinchPowers: Not related to this PR, but adds flexibility. (And I used it while debugging.)

rest/rest_collection_impl.h
```diff
@@ -561,7 +561,8 @@ addBackgroundJobInThread(Key key,
 
         // Set up the task, without starting it yet
         auto task = std::make_shared<BackgroundTask>();
-        task->config = std::move(config);
+        task->running = false;
+        task->config = config;
```
FinchPowers: std::move(config)... if it did not work the first time, the config was lost.

rest/rest_collection_impl.h
```diff
@@ -615,6 +616,7 @@ addBackgroundJobInThread(Key key,
 
         if (impl->entries.cmp_xchg(oldEntries, newEntries, true)) {
             // Now we can start the task, since the commit succeeded
+            task->running = true;
```
FinchPowers: As long as it doesn't enter this condition scope, the task was not really running, hence the deadlock.

rest/rest_collection_impl.h
guyd: Good to know.

jeremybarnes:
Nice work Mich!

Please consider:
1.  Making the destructor noexcept, by catching an exception from cancel() and aborting immediately.  In the more distant future, cancel() should be made noexcept, but that will require deeper changes.
2.  Adding a test to rest_collection_test to reproduce the condition without the baggage of the rest of MLDB.  There are already other stress tests there.

rest/rest_collection_impl.h
```diff
@@ -561,7 +561,8 @@ addBackgroundJobInThread(Key key,
 
         // Set up the task, without starting it yet
         auto task = std::make_shared<BackgroundTask>();
-        task->config = std::move(config);
+        task->running = false;
```
guyd: Shouldn't we set it to false at construction?  I am surprised that this was not the case.

FinchPowers: Possibly. It crossed my mind but I did not ask. I'll dig it.

FinchPowers:
@jeremybarnes So far I was unsuccessful at reproducing the issue without the MLDB layer. I had to use inception which was both slow and time consuming enough to cause the issue.

I am opened to advice on how to try to reproduce it without the MLDB layer.

PR #829, opened 2017-01-25
MLDB-2113 run-script wait for all procs

FinchPowers:


server/mldb_runner.cc
```diff
@@ -408,6 +408,15 @@ int main(int argc, char ** argv)
...
         if (!dontExitAfterScript) {
+            int tasksBefore = -1;
+            int tasks;
+            while ((tasks = BackgroundTaskBase::getRunningTasks()) != 0) {
+                if (tasks != tasksBefore) {
+                    tasksBefore = tasks;
+                    cerr << "Still waiting for " << tasks << " task(s)" << endl;
+                }
+                std::this_thread::sleep_for(std::chrono::milliseconds(100));
+            }
```
FinchPowers: Some tasks seem to never be destroyed so we now have locks!

rest/rest_collection.h
```diff
@@ -117,7 +117,7 @@ struct BackgroundTaskBase {
     /** When destroying a background task, we make sure it's cancelled or
         finished so that there is nothing outstanding.
     */
-    ~BackgroundTaskBase();
+    virtual ~BackgroundTaskBase();
```
guyd: Euh!  Ce destructeur fait quelque chose?  Il aurait du etre virtuel.

FinchPowers: En effet! Scarry!

server/mldb_server.cc
```diff
@@ -389,6 +390,17 @@ shutdown()
...
+    int tasksBefore = -1;
+    int tasks;
+    while ((tasks = BackgroundTaskBase::getPendingTasks()) != 0) {
+        if (tasks != tasksBefore) {
+            tasksBefore = tasks;
+            INFO_MSG(logger) << "Still waiting for " << tasks
+                             << " task(s)";
+        }
+        std::this_thread::sleep_for(std::chrono::milliseconds(100));
+    }
```
FinchPowers: On hold. This is a band aid. Shutdown should shutdown and join services, thread pools and the like properly.

rest/rest_collection.cc
```diff
@@ -648,11 +648,19 @@ BackgroundTaskBase::
 BackgroundTaskBase()
     : running(false), state(State::INITIALIZING)
 {
+    ++pendingTasks;
+    needToDecrementPendingTasks.store(1);
+}
+
+atomic<int> BackgroundTaskBase::pendingTasks{0};
+int BackgroundTaskBase::getPendingTasks() {
+    return BackgroundTaskBase::pendingTasks.load();
```
jeremybarnes: no need for load()

rest/rest_collection.cc
```diff
@@ -648,11 +648,19 @@ BackgroundTaskBase::
 BackgroundTaskBase()
     : running(false), state(State::INITIALIZING)
 {
+    ++pendingTasks;
+    needToDecrementPendingTasks.store(1);
```
jeremybarnes: no need for store, use assignment

rest/rest_collection.h
jeremybarnes: It will only cause a problem if the derived class has a non-trivial destructor :)  but still scary

PR #830, opened 2017-01-25
Mldb 2111 allow select of more complex group by keys

steadtler:


server/bound_queries.cc
```diff
@@ -1620,6 +1666,34 @@ BoundGroupByQuery(const SelectExpression & select,
     //false means no implicit sort by rowhash, we want unsorted
     subSelect.reset(new BoundSelectQuery(subSelectExpr, from, alias, when, where, subOrderBy, calc, numBuckets));
 
+    std::vector<std::shared_ptr<ExpressionValueInfo> > groupInfo;
+    for (size_t c = 0; c < groupBy.clauses.size(); ++c) {
```
guyd: why not
```
for (const auto & boundCalc : subSelect->boundCalc)
```

sql/sql_expression_operations.cc
```diff
@@ -4197,6 +4197,52 @@ wildcards() const
...
+            },
+            this,
+            outputInfo};
+}
+
+Utf8String
+GroupByKeyExpression::
+print() const
+{
+    return "GroupBy Key";
```
guyd: Don't we need to have the index at least?

sql/sql_expression_operations.cc
```diff
@@ -4197,6 +4197,52 @@ wildcards() const
...
+}
+
+std::shared_ptr<SqlExpression>
+GroupByKeyExpression::
+transform(const TransformArgs & transformArgs) const
+{
+    return make_shared<GroupByKeyExpression>(index);
```
guyd: Shouldn't we throw here that it is not implemented?

sql/sql_expression_operations.h
```diff
@@ -599,4 +599,36 @@ struct SelectColumnExpression: public SqlRowExpression {
...
+    bind(SqlBindingScope & context) const override;
+
+    virtual Utf8String print() const override;
+
+    virtual std::shared_ptr<SqlExpression>
+    transform(const TransformArgs & transformArgs) const override;
+
+    virtual std::string getType() const override
+    {
+        return "groupbykey";
```
guyd: `SelectColumnExpression` uses camel case with first lowercase.

testing/MLDB-2111-group-by-expression.py
```diff
@@ -0,0 +1,127 @@
...
+            SELECT x+1 FROM (SELECT x:1) GROUP BY x+1
+        """)
+
+        self.assertTableResultEquals(res, 
+            [[ "_rowName", "x+1"],
+             [ "[2]", 2 ]])
+
+    def test_groupby_expression_named(self):
+        res = mldb.query("""
+            SELECT x+1 as z FROM (SELECT x:1) GROUP BY x+1
```
guyd: Group by with a single row might not reveal all the issues.

testing/MLDB-2111-group-by-expression.py
```diff
@@ -0,0 +1,127 @@
...
+            [[ "_rowName", "x+1"],
+             [ "[2,2]", 2 ],
+             [ "[2,4]", 2 ],
+             [ "[3,2]", 3 ],
+             [ "[3,4]", 3 ]])
+
+    ## Rowname & rowHash in the select is different than in the group by
+    def test_groupby_rowname(self):
+        res = mldb.query("""
+            SELECT rowName() FROM ds GROUP BY rowName() LIMIT 1
```
guyd: Why the limit?  I would like to see all the grouped by rows.

server/bound_queries.cc
steadtler: Because the size of the calc vector is greater or equal to the size of the group by keys, since we also put aggregator's arguments in the calc vector, after the group by clauses.

server/bound_queries.h
```diff
@@ -173,13 +174,13 @@ struct BoundGroupByQuery {
     BoundSqlExpression boundRowName;
 
     // Select Expression to resolve
-    const SelectExpression& select;
+    SelectExpression select;
 
     // Having Expression to resolve
-    const SqlExpression& having;
+    std::shared_ptr<SqlExpression> having;
```
steadtler: Because we replace it and SqlExpression should only be allocated as shared_ptr

sql/sql_expression_operations.cc
steadtler: it is implemented, but we dont need to transform the args because it has no children expressions.

sql/sql_expression_operations.cc
```diff
@@ -4197,6 +4197,52 @@ wildcards() const
...
+{
+    auto getGroupbyKey = scope.doGetGroupbyKey(index);
+
+    if (!getGroupbyKey.info) {
+        throw HttpReturnException(400, "scope " + MLDB::type_name(scope)
+                            + " doGetGroupbyKey '"
+                            + "' didn't return info");
+    }
+
+    auto outputInfo = getGroupbyKey.info->getConst(false);
```
steadtler: We need the (group by) row to evaluate its constValue(). Its a bit tricky, so I put a comment to show its not an oversight.

testing/MLDB-2111-group-by-expression.py
steadtler: Its good to have a variety. Some use the full dataset, some uses the select

steadtler:
@guyd 

I addressed most of the comments. Remember that this is not an optional paths, so all the other group by unit tests will still test regressions.

testing/MLDB-2111-group-by-expression.py
```diff
@@ -123,5 +133,28 @@ def test_groupby_having(self):
...
+        msg = "variable 'x' must appear in the GROUP BY clause or be used in an aggregate function"
+        with self.assertRaisesRegexp(mldb_wrapper.ResponseException, msg):
+            res = mldb.query("""
+                SELECT x+1 FROM (SELECT x:1) GROUP BY 1+x
+            """)
+
+        msg = "variable 'x' must appear in the GROUP BY clause or be used in an aggregate function"
+        with self.assertRaisesRegexp(mldb_wrapper.ResponseException, msg):
+            res = mldb.query("""
+                SELECT x+1*3 FROM (SELECT x:1) GROUP BY 1+x
```
guyd: Here I would have liked to see `GROUP BY x+1`.

testing/MLDB-2111-group-by-expression.py
```diff
@@ -123,5 +133,28 @@ def test_groupby_having(self):
             [["_rowName","z"],
              ["[3]", 0]])
 
+    def test_groupby_temporal(self):
+        res = mldb.query("""
+            select temporal_latest({*}) as * from ds2
```
guyd: I was thinking of `GROUP BY temporal_latest(x)`.

PR #831, opened 2017-01-26
MLDB-2126 use flattened column names in export.csv

steadtler:
We also need to find the correct ExpressionValueInfo for "select x": x is not necessarily an atom.


sql/expression_value.cc
```diff
@@ -263,6 +271,16 @@ allColumnNames() const
     return result;
 }
 
+std::vector<ColumnPath>
+ExpressionValueInfo::
+allAtomNames() const
+{
+    std::vector<ColumnPath> result;
+    for (auto & val: getKnownAtoms())
```
guyd: Could reserve here the space in the vector.

guyd:
+1

PR #833, opened 2017-01-27
MLDB-2105 select expr square bracket col name support

FinchPowers:


sql/sql_expression.cc
```diff
@@ -759,6 +759,32 @@ static Utf8String matchIdentifier(ParseContext & context, bool allowUtf8)
     return result;
 }
 
+void matchSingleQuoteStringUTF8(ParseContext & context,
```
FinchPowers: I had to move it ahead of the new code to allow the compiler to find it.

testing/MLDB-2105_select_expr_square_bracket_col_name_support.py
```diff
@@ -0,0 +1,77 @@
...
+#
+
+mldb = mldb_wrapper.wrap(mldb)  # noqa
+
+class Mldb2105SelectExprSquareBracketColNameSupport(MldbUnitTest):  # noqa
+
+    def test_match(self):
+        res = mldb.query("SELECT a['foo'] FROM (SELECT {foo: 123} AS a)")
+        self.assertEqual(res, [
+            ['_rowName', 'a.foo'],
```
FinchPowers: The resulting row name comes out as the dotted notation. Is it a problem?

PR #834, opened 2017-01-30
Square bracket behaviour RFC

FinchPowers:


testing/square_bracket_accessor_test.py
```diff
@@ -0,0 +1,150 @@
...
+# square_bracket_accessor_test.py
+# Francois-Michel L'Heureux, 2017-01-27
+# This file is part of MLDB. Copyright 2017 mldb.ai inc. All rights reserved.
+#
+import unittest
+
+mldb = mldb_wrapper.wrap(mldb)  # noqa
+
+class Mldb2105SelectExprSquareBracketColNameSupport(MldbUnitTest):  # noqa
+
```
FinchPowers: First half of the test. Accessors with single quotes. This is what MLDB already does.

testing/square_bracket_accessor_test.py
```diff
@@ -0,0 +1,150 @@
...
+        res = mldb.query("""
+            SELECT a['utéf8'] FROM (
+                SELECT parse_json('{"utéf8" : 123}') AS a
+            )
+                         """)
+        self.assertEqual(res, [
+            [u'_rowName', u"a['utéf8']"],
+            [u'result', u'utéf8']
+        ])
+
```
FinchPowers: Second half of the test, square bracket with unquoted string. All of those test currently fail, mostly with `Couldn't find column 'foo' in extract`. Should MLDB return the proper thing?

testing/square_bracket_accessor_test.py
```diff
@@ -0,0 +1,150 @@
...
+        res = mldb.query("""
+            SELECT a[utéf8] FROM (
+                SELECT parse_json('{"utéf8" : 123}') AS a
+            )
+                         """)
+        self.assertEqual(res, [
+            [u'_rowName', u"a[utéf8]"],
+            [u'result', 123]
+        ])
+
```
FinchPowers: I have no test with double quotes. Should I add some? What should it do for `SELECT a["foo"] FROM (SELECT {foo: 123} AS a)")` ?

testing/square_bracket_accessor_test.py
```diff
@@ -0,0 +1,150 @@
...
+            ['_rowName', "a['bar']"],
+            ['result', 'bar']
+        ])
+
+    def test_string_chaining(self):
+        res = mldb.query("SELECT a['f']['b']['z'] FROM (SELECT {f: {b: 123}} AS a)")
+        self.assertEqual(res, [
+            ['_rowName', "a['f']['b']['z']"],
+            ['result', 'z']
+        ])
```
FinchPowers: This surprises me. It keeps applying string accesses over strings.

steadtler: We ignore the context and just return the constant. Thats what I would expect.

testing/square_bracket_accessor_test.py
steadtler: a["foo"] and a[foo] should be the same thing.

testing/square_bracket_accessor_test.py
jeremybarnes: I think that it's OK that these fail for now; we know at bind time that we're asking for row[null] which isn't valid (will always fail at runtime).  If later on we want to make it so that row[null] = null then we should change that separately.

testing/square_bracket_accessor_test.py
jeremybarnes: Exactly... they should be the same thing.  However, a["hello world"] is valid, whereas a[hello world] should be a syntax error.

jeremybarnes:
I completely support this change; it was only in deference to Nic's position that we kept it in the past.  The semantics now match those of Python, Javascript and C++.  We should be aware however that it will break many existing tests and some existing code.

We should also (later) make the . operator a first class expression so that instead of func(x)['result'] as now, we can do func(x).result; that will make the syntax more consistent and easier to work with.

testing/square_bracket_accessor_test.py
FinchPowers: Good. I'll adapt the test accordingly.

PR #836, opened 2017-01-30
MLDB-2105 sql square bracket enh + test update

FinchPowers:


testing/square_bracket_accessor_test.py
```diff
@@ -73,153 +73,137 @@ def test_string_utf8(self):
...
-    @unittest.expectedFailure
-    def test_match_no_double_quotes(self):
+    def test_match_no_quotes(self):
         """
         It doesn't matter that it maches, it's a string.
         """
         res = mldb.query("SELECT a[foo] FROM (SELECT {foo: 123} AS a)")
         self.assertEqual(res, [
-            ['_rowName', "a[foo]"],
+            ['_rowName', "a.foo"],
```
FinchPowers: Whatever the operator used in the query, the result uses the dot notation.

testing/square_bracket_accessor_test.py
```diff
@@ -73,153 +73,137 @@ def test_string_utf8(self):
...
+
     def test_inner_whitespaces_double_quotes(self):
         res = mldb.query("""
             SELECT a["f b"] FROM (
                 SELECT parse_json('{"f b" : 123}') AS a
             )
                          """)
         self.assertEqual(res, [
-            ['_rowName', "a[f b]"],
+            ['_rowName', "a.f b"],
```
FinchPowers: The dot notation returns non "round trippable" result though.

FinchPowers:
I think we should have a solution for the "invalid dot notation" result before merging this. The safest constant way is to always use the bracket notation with double quotes in the results. Downside: Backward compatibility where scripts check resulting column name.

If we use a hybrid approcah, dot notation when possible, fallback to square bracket + double quote when impossible, this may be confusing.

Thoughts?

testing/square_bracket_accessor_test.py
FinchPowers: Ok never mind, MLDB returns that already when we do something like

`SELECT a."f b"`

steadtler:
+1 with some hesitation... now the bracket operator will mean slightly different things when whats on the left is an identifier and when its not. When its an identifier, it'll only select a column from it. When its any row expression on the left, it'll allow evaluation of an expression in the context of that row expression.

FinchPowers:
@jeremybarnes What do you think?

jeremybarnes:
I think that this needs to be thought through more.  We're now making the . an operator that can be used anywhere; I think it's unclear what the precedence is and how it interacts with other lexical constructs that include dots.  To me, [] and . are different constructs syntactically, it just so happens that . is implemented in terms of [].  Let's keep the code cleanup, and do a whiteboard session on the semantics.

PR #837, opened 2017-01-30
[MLDB-2131] added a macro to document all TF operations wrapped in MLDB

guyd:
Added a new macro so that we can automatically generate the documentation of all the tensorflow operations that are wrapped in MLDB.

### DIFF
--- a/tensorflow/tensorflow_plugin.cc
+++ b/tensorflow/tensorflow_plugin.cc
@@ -1727,6 +1727,13 @@ struct TensorflowPlugin: public Plugin {
                                       std::placeholders::_1,
                                       std::placeholders::_2,
                                       std::placeholders::_3));
+
+        tfOperationsMacroHandle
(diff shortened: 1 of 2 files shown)


tensorflow/tensorflow_plugin.cc
```diff
@@ -1803,6 +1810,31 @@ struct TensorflowPlugin: public Plugin {
     }
 
     /** Documentation macro for TF operations */
+    void tfOperationsMacro(MacroContext & context,
+                           const std::string & macroName)
+    {
+        auto writeTocEntry = [&context](const Utf8String & op) {
```
FinchPowers: writeTocEntry or
writeToEntry?

FinchPowers:
+1

tensorflow/tensorflow_plugin.cc
guyd: toc : table of contents

PR #842, opened 2017-02-07
MLDB-2139 plugin dependencies

FinchPowers:


server/mldb_server.cc
```diff
@@ -404,47 +404,68 @@ scanPlugins(const std::string & dir_)
...
+            logger->error() << "details:";
+            logger->error() << jsonEncode(exc.details);
+            logger->error() << "plugin will be ignored";
+            return;
+        } catch (const std::exception & exc) {
+            logger->error() << "loading plugin " << dir << ": " << exc.what();
+            logger->error() << "plugin will be ignored";
+            return;
+        }
+    };
```
FinchPowers: This error handler was copy pasted. I extracted it.

server/mldb_server.cc
```diff
@@ -404,47 +404,68 @@ scanPlugins(const std::string & dir_)
...
+
+    auto loadPlugin = [&] (PluginManifest & manifest) {
```
FinchPowers: Same code as before with the error handler extracted.

server/mldb_server.cc
```diff
@@ -404,47 +404,68 @@ scanPlugins(const std::string & dir_)
...
-                    auto config = manifest.config.params.convert<PluginResource>();
-                    config.address = dir;
-                    manifest.config.params = config;
-                    auto plugin = plugins->obtainEntitySync(
-                        manifest.config, nullptr /* on progress */);
+                for (const auto & dep: manifest.dependencies) {
+                    if (loadedPlugins.find(dep) == loadedPlugins.end()) {
+                        missingDependencies[manifest.config.id] = manifest;
+                        return;
+                    }
```
FinchPowers: Don't load the plugin now if there are missing deps.

server/mldb_server.cc
```diff
@@ -474,20 +495,36 @@ scanPlugins(const std::string & dir_)
...
-            return;
-        } catch (const std::exception & exc) {
-            logger->error() << "error scanning plugin directory  "
-                            << dir << ": " << exc.what();
-            logger->error() << "plugins will be ignored";
-            return;
+        });
+    }
+
+    // Dumb algo, loop until nothing is left or the last pass didn't load anything
```
FinchPowers: We resolve the dependencies with a dumb loop. There should not be a lot so perf is not an issue.

FinchPowers:
If someone has any ideas on how to test this, I'm opened to suggestions.

FinchPowers:
Major path bug found. Will fix.

server/plugin_manifest.h
```diff
@@ -21,6 +21,7 @@ namespace MLDB {
 /*****************************************************************************/
 
 struct PluginManifest {
+    std::vector<std::string> dependencies;
```
jeremybarnes: Should be UTF8String for consistency probably.

jeremybarnes:
My feeling is that this would be much cleaner if we simply added a "require" function for the dependencies, that was called when loading.  For example, in the main plugin script we would write

mldb.require('tensorflow')

which would then load up the given plugin if it wasn't already loaded.  It keeps it declarative, and allows us to avoid having huge lists of hardcoded dependencies which brings in the kitchen sink every time.

In the redesign of the Python plugin system, we could do something similar: eg mldb.pythonRequires('pippackagename >0.34')

PR #843, opened 2017-02-07
Fix wrong implementation of rowLength for embeddings

mathieumb:
Fix wrong implementation of rowLength for embeddings and implement getAtomCount.

rowLength is meant to return the number of columns in the first structured level, and that worked for row representations. However for embeddings representations is was returning the number of atoms. 

If you need the total number of atoms, the proper method is getAtomCount.


FinchPowers:
Test to avoid regression?

mathieumb:
@FinchPowers
I had to change from rowLength to getAtomCount in a bunch of places, so now both are tested by dozens of tests.

mathieumb:
verbal +1 !

PR #847, opened 2017-02-13
Mldb 2142 prefix suffix

mathieumb:
Two simple builtins to remove prefix and suffix from strings.


guyd:
`regex_replace` couldn't do that?

mathieumb:
Yes but this is just easier and clearer.

guyd:
+1

PR #848, opened 2017-02-13
MLDB-2143 fix issue with utf8 labels in classifier accuracy testing

mathieumb:
one-liner bug.


guyd:
So, if I understand correctly, the training and the prediction were not actually impacted by the bug. It was only the reporting of the stats that was not handling non-ascii characters.

guyd:
+1

mathieumb:
It was throwing on a generic "cannot cast Utf8 to ASCII" exception while formatting the result JSON.

PR #854, opened 2017-02-17
Mldb 2150 fetcher http return error

FinchPowers:


jeremybarnes:
Let me look at it properly before merging please

vfs/fs_utils.h
```diff
@@ -48,10 +48,22 @@ struct FsObjectInfo {
     std::map<std::string, Json::Value> objectMetadata;  ///< Metadata from obj
 
     operator bool () const { return exists; }
+
+    // make dynamic_pointer_cast work otherwise not detected as polymorphism
+    // struct
+    virtual ~FsObjectInfo() {}
```
jeremybarnes: Can't do this; the class is designed to be value copyable.  This will lead to strange and intermittent bugs due to object slicing.

vfs/fs_utils.h
```diff
@@ -48,10 +48,22 @@ struct FsObjectInfo {
...
 };
 
 DECLARE_STRUCTURE_DESCRIPTION(FsObjectInfo);
 
+struct ExcAwareObjectInfo : FsObjectInfo {
```
jeremybarnes: Why does this need to derive from FsObjectInfo?  Can't it include that as a member instead?

vfs/http_streambuf.cc
```diff
@@ -87,6 +87,14 @@ struct HttpStreamingDownloadSource {
         return future.get();
     }
 
+    shared_ptr<FsObjectInfo> getObjectInfo() const
```
jeremybarnes: What does this have to do with the exceptions?

vfs/fs_utils.h
FinchPowers: I could, but I did not want to contaiminate it with a new member.

vfs/http_streambuf.cc
```diff
@@ -294,6 +306,7 @@ struct HttpStreamingDownloadSource {
                 if (!headerSet.exchange(true)) {
                     headerPromise.set_exception(lastExc);
                 }
+                info->what = exc.what();
```
jeremybarnes: We should at least assert on info being not null here.

vfs/http_streambuf.cc
FinchPowers: I told you you would cry... This object is available to the caller so this is why I picked it.

vfs/http_streambuf.cc
FinchPowers: It can't be, it is instantiated in reset().
I can add an assert anyway.

FinchPowers:
@jeremybarnes I made some changes. Please rereview.

vfs/fs_utils.h
```diff
@@ -48,6 +48,8 @@ struct FsObjectInfo {
     std::map<std::string, Json::Value> objectMetadata;  ///< Metadata from obj
 
     operator bool () const { return exists; }
+
+    std::string what;                ///< MLDB-2150, hack, do not use
```
jeremybarnes: Instead of that, why don't we simply call it errorMessage.  Then it's more generic and we can use it.  What is a terrible name.

vfs/http_streambuf.cc
```diff
@@ -140,6 +148,8 @@ struct HttpStreamingDownloadSource {
 
         bool httpAbortOnSlowConnection;
 
+        shared_ptr<FsObjectInfo> info;
```
jeremybarnes: Instead of info, I would simply store a std::exception_ptr.  Use std::get_exception()

vfs/http_streambuf.cc
```diff
@@ -294,6 +305,9 @@ struct HttpStreamingDownloadSource {
                 if (!headerSet.exchange(true)) {
                     headerPromise.set_exception(lastExc);
                 }
+                if (info) {
+                    info->what = exc.what();
```
jeremybarnes: And here, we can set to getExceptionString from the saved exception object

vfs/http_streambuf.cc
```diff
@@ -315,18 +329,19 @@ struct HttpStreamingDownloadSource {
     {
         impl.reset();
     }
+
 };
 
-std::pair<std::unique_ptr<std::streambuf>, FsObjectInfo>
+std::pair<std::unique_ptr<std::streambuf>, shared_ptr<FsObjectInfo>>
```
jeremybarnes: No need to pass a shared_ptr.  Just pass the FsObjectInfo directly.

FinchPowers: I think I need to, otherwise it copy constructs a new one before the exception is set.

plugins/fetcher.cc
```diff
@@ -113,9 +112,16 @@ struct FetcherFunction: public ValueFunctionT<FetcherArgs, FetcherOutput> {
                 blob = CellValue::blob(streamo.str());
             }
 
-            result.content = ExpressionValue(std::move(blob), info.lastModified);
+            FsObjectInfo info = stream.info();
```
jeremybarnes: OK, so... sorry to do this again... the filter_streams will throw their exceptions when close() is called.  So whenever we want to make sure that an exception is thrown, we need to call close().  Why?  Because sometimes the exception happens in a worker thread, which will crash the exception if it's thrown.  Can we try to make close() throw the exception, in other words complete any missing logic in the filter_stream::close() method rather than putting it here?

vfs/fs_utils.h
```diff
@@ -48,6 +49,8 @@ struct FsObjectInfo {
     std::map<std::string, Json::Value> objectMetadata;  ///< Metadata from obj
 
     operator bool () const { return exists; }
+
+    std::exception_ptr excPtr;  ///< MLDB-2150, hack, do not use
```
jeremybarnes: If we make close() work properly, we won't need this and/or we can make an "error" field that's just a string.

vfs/http_streambuf.cc
jeremybarnes: No, it should be possible to move it.

plugins/fetcher.cc
FinchPowers: It still doesn't work. The exception is caught somewhere within boost::iostreams and never makes it out.

PR #855, opened 2017-02-20
Mldb 2152 fetcher concurrency limit

FinchPowers:


testing/http_streambuf_test.cc
```diff
@@ -72,7 +72,7 @@ struct TestServer {
 
         // this is ok because it doesn't require a virtualenv
         runner.run({"/usr/bin/env", "python",
-                    "mldb/testing/http_streambuf_test_server.py",
+                    "mldb/testing/test_server.py",
```
FinchPowers: Renamed it because it's useful to other tests.

testing/test_server.py
```diff
@@ -20,8 +21,13 @@
 Send a POST request:
     curl -d "foo=bar&bin=baz" http://localhost
 """
+from SocketServer import ThreadingMixIn
```
FinchPowers: Updated to be multithreaded.

plugins/fetcher.cc
```diff
@@ -25,9 +28,15 @@ namespace MLDB {
 /** Function that fetches a row from a url. */
 
 struct FetcherFunctionConfig {
-    FetcherFunctionConfig()
+    FetcherFunctionConfig() : maxConcurrentFetch(-1),
```
jeremybarnes: Do these inline and remove the constructor, eg int maxConcurrentFetch = -1.  Also, there shouldn't be a mutex or a condition variable in the config object; that belongs in the fetcher itself.

plugins/fetcher.cc
```diff
@@ -86,10 +107,37 @@ struct FetcherFunction: public ValueFunctionT<FetcherArgs, FetcherOutput> {
     {
         functionConfig = config.params.convert<FetcherFunctionConfig>();
     }
+
+    struct ConcurrencyHandler {
+        ConcurrencyHandler(FetcherFunctionConfig & config) :
```
jeremybarnes: The config should be a POD type.  You will need to pass the variables in.

plugins/fetcher.cc
```diff
@@ -124,7 +172,7 @@ struct FetcherFunction: public ValueFunctionT<FetcherArgs, FetcherOutput> {
         return result;
     }
 
-    FetcherFunctionConfig functionConfig;
+    mutable FetcherFunctionConfig functionConfig;
```
jeremybarnes: No, this should remain const.  You can add a mutable CV and mutex to this class.

testing/test_server.py
jeremybarnes: Do we need a way for MLDB to run something (eg a function) in multiple threads?  It would be nice to avoid having to jump outside of the normal MLDB functionality for this.

FinchPowers: This is a dummy test server, I updated it to be multithreaded to be able to test the fetcher concurrency limit.

PR #856, opened 2017-02-23
ubuntu-toolchain-r/test to get recent libstdc++

jraby:


jeremybarnes:
We should put a decent description in the PR!

I don't think this is a good idea, as we lose any kind of pinning of the version (we will end up pulling in whatever packages were the latest in the PPA when we create an image).  Isn't there an official updated libstdc++ that we can more explicitly point to?

jraby:
Maybe. I'll revisit if the ABI breaks.

PR #857, opened 2017-02-24
Fix one route in reporting progress of transform dataset

mathieumb:


plugins/sql_functions.cc
```diff
@@ -729,9 +729,7 @@ run(const ProcedureRunConfig & run,
     auto transformingStep = bindingStep->nextStep(1);
     auto onTransformingProgress = [&](const ProgressState & percent) {
         lock_guard<mutex> lock(progressMutex);    
-        if (transformingStep->value < (float) percent.count / *percent.total) {       
-            transformingStep->value = (float) percent.count / *percent.total;     
-        }
+        transformingStep->updateValue((float) percent.count / *percent.total);
```
FinchPowers: I think you should leave the check, we don't want it to go back. It can occur because of race conditions.

mathieumb: I move the check inside updateValue to avoid copy pasting it everywhere

FinchPowers: I don't see it in the PR. Can you point it to me?

PR #860, opened 2017-03-08
MLDB-2161 - fixed the handling of utf8 character in json encoder that was using t…

guyd:
…he wrong overload


FinchPowers:
You should also put the ticket number in the title of the PR as well as in the commit messages.

PR #862, opened 2017-03-09
[MLDB-2163] add support to call functions and query with POST

guyd:


guyd:
@FinchPowers What is missing?

FinchPowers:
* routes /v1/query and /v1/functions must not to be modified.
* route /v1/get must be added.

server/mldb_server.cc
```diff
@@ -217,6 +217,14 @@ initRoutes()
                                      "Do we sort the column names",
                                      false));
 
+        addRouteAsync(
+            versionNode, "/redirect/get", {"POST"}, "Redirect POST as GET with body. "
+            "Use this route only with system that does not support sending a GET with a body.",
```
FinchPowers: system>s that do not<

server/mldb_server.cc
```diff
@@ -263,6 +271,32 @@ runHttpQuery(const Utf8String& query,
                        rowNames, rowHashes, sortColumns);
 }
 
+void
+MldbServer::
+handleRedirectToGet(const string & uri,
+                    RestConnection & connection,
+                    const Json::Value & body) const
+{
```
FinchPowers: Is there any risk that someone could make a call to this route to get something not related to mldb?
Say I am on a LAN. I have no access to the extranet but I have access to MLDB. MLDB has access to the extranet. Can I use MLDB to get out?

This is a security risk question.

guyd: Yes.  That seems possible.

FinchPowers: Would checking that the target starts with `/v1` be enough to block such attempts?

jeremybarnes: It's not possible.  We don't actually make an HTTP call; we simply call the handleRequest method.  So there is no security issue here.

jeremybarnes:
+1 from me

PR #865, opened 2017-03-10
Mldb 2166 mime type

mailletf:


sql/builtin_functions.cc
```diff
@@ -3439,6 +3440,54 @@ bind(const std::vector<BoundSqlExpression> & args,
     }
 }
 
+
+BoundFunction get_mime_type(const std::vector<BoundSqlExpression> & args)
```
FinchPowers: rm `get_`

sql/builtin_functions.cc
```diff
@@ -3439,6 +3440,54 @@ bind(const std::vector<BoundSqlExpression> & args,
...
+                }
+
+                if(magic_load(magic, NULL) != 0) {
+                    magic_close(magic);
+                    throw ML::Exception("Error loading magic database - %s\n", magic_error(magic));
+                }
+
+                if(!args[0].isAtom())
+                    throw MLDB::Exception("Mime type extraction requires that an atomic value "
+                            "of type BLOB is passed to it. Use the fetcher function to get the file blob.");
```
FinchPowers: Should we be as explicit as that? (Use the fetcher function... which may be renamed and is not the only way of doing it.)
If we want to be that explicit, then we could say `fetcher()[content]`

sql/builtin_functions.cc
```diff
@@ -3439,6 +3440,54 @@ bind(const std::vector<BoundSqlExpression> & args,
...
+
+                const unsigned char * data = input.blobData();
+                const size_t len = input.blobLength();
+
+                mime = magic_buffer(magic, data, len);
+
+                Utf8String str(mime);
+                magic_close(magic);
```
FinchPowers: What happens if this doesn't get called? Does it leak? If so we need to wrap it into an RAII structure.

FinchPowers:
You need to add the new dependency to `mldb_base/docker_create_mldb_base.sh` too.

PR #866, opened 2017-03-15
Trivial changes

jeremybarnes:
Nothing to see here, move on... indentation, and making one frequently inlined function always inline.


mailletf:
+1

PR #867, opened 2017-03-15
Fixed offset error on last line of chunk (MLDB-2167)

jeremybarnes:
Fixed error using gcount() after readline(), leading to thinking that the chunk is bigger than it is and junk characters in between chunks.


mailletf:
Is there a small test that we can add?

jeremybarnes:
It's not easy to reproduce, and I only managed to do so using customer data.  I'll look to see if I can add a hook to help make testing easier.

jeremybarnes:
We need at least 20MB of data in order to reproduce in a test case.

PR #868, opened 2017-03-15
MLDB-2168 import text skip lines

jeremybarnes:
Allows a `skipLineRegex` parameter in `import.text` that can for example skip blank lines or those starting with a comment.  Particularly useful when the file starts with comments which need to be skipped before the headers can be read.


mathieumb:
+1, however Im worrying about how big this procedure is getting.

PR #870, opened 2017-03-15
Allow Excel formulas (=...) to be processed in CSV (MLDB-2170)

jeremybarnes:
Allows for the `="123"` found in some CSV exports from Excel to be properly parsed (by ignoring it).  This was seen in the wild.


plugins/importtext_procedure.cc
```diff
@@ -92,6 +92,9 @@ ImportTextConfigDescription::ImportTextConfigDescription()
              "If true, the indexes of the columns will be used to name them."
              "This cannot be set to true if headers is defined.",
              false);
+    addAuto("processExcelFormulas", &ImportTextConfig::processExcelFormulas,
+            "Process formulas in `=\"...\"` format from Excel export of "
```
guyd: This suggests that we do more that we actually do.  It is not all formulas but only the simple formulas `="n"` where `n` is a numeral.

jeremybarnes:
@mailletf There are lots of datasets out there produced by Excel with this exact format.  It's not overfitting a single dataset.

mailletf:
allrighty +1

PR #871, opened 2017-03-15
Multilabel classification

mathieumb:
PR FOR FEEDBACK NOT FOR MINUTE REVIEW.
@jeremybarnes @mailletf 

Ive applied the "onevsall" to the text tags problem and it works beautifully, it really simplifies things.
It didnt turn out that complicated.


ml/jml/feature_set.cc
```diff
@@ -94,9 +94,9 @@ replace(const Feature & feat, float val)
             it->second = val;
             ++it;
             int start = it - features.begin();
-            int pos = 0;
+            int pos = start;
```
jeremybarnes: This should go in separately as a separate PR and issue; it's a pure bug fix that should not depend on the rest.

ml/jml/onevsall.cc
```diff
@@ -0,0 +1,157 @@
+// This file is part of MLDB. Copyright 2017 mldb.ai inc. All rights reserved.
```
jeremybarnes: Naming convention is one_vs_all (minor)

ml/jml/onevsall.h
```diff
@@ -0,0 +1,72 @@
...
+
+
+namespace ML {
+
+
+/*****************************************************************************/
+/* ONEVSALL_CLASSIFIER                                                       */
+/*****************************************************************************/
+
+/** Classifier that always returns the zero label. */
```
jeremybarnes: I don't understand the design here.  Could you add a comment that describes what we're attempting to do with this classifier?

ml/jml/onevsall_generator.cc
```diff
@@ -0,0 +1,139 @@
...
+         const std::vector<Feature> & features,
+         int) const
+{
+    std::shared_ptr<OneVsAllClassifier> current = make_shared<OneVsAllClassifier>(model);
+
+    ML::Mutable_Feature_Info labelInfo = ML::Mutable_Feature_Info(ML::BOOLEAN);
+
+    int labelValue = 0; 
+
+    std::shared_ptr<Mutable_Feature_Space> fs2 = std::dynamic_pointer_cast<Mutable_Feature_Space>(make_sp(feature_space->make_copy()));
```
jeremybarnes: This won't be true in most cases.  If it's not a mutable feature space, we will need to copy into a new one.

ml/jml/training_data.h
```diff
@@ -85,6 +85,11 @@ class Training_Data {
     {
         return feature_space_;
     }
+
+    void set_feature_space(std::shared_ptr<const Feature_Space> fs)
```
jeremybarnes: this shouldn't be necessary

jeremybarnes:
I think it would be helpful if there was a high level description of what problem motivated this changeset.

PR #872, opened 2017-03-17
Wrap error message when retrieving the arguments of a tensorflow builtin

mathieumb:


tensorflow/MLDB-1736-tensorflow-builtins.js
```diff
@@ -60,4 +59,8 @@ mldb.log(png);
 var conv_output = mldb.query("SELECT tf_Conv2D({input: [[[[1], [1], [1]], [[1], [1], [1]], [[1], [1], [1]]]], filter: [[[[1]], [[2]]], [[[2]], [[1]]]]}, {T: { type: 'DT_FLOAT'}, padding: 'SAME', strides: [ 1, 1, 1, 1] }) AS res")
 mldb.log(conv_output)
 
+//check that we get the name of the node in the error message
+var jpeg = mldb.get('/v1/query', {q: "SELECT tf_EncodeJpeg([[]]) AS *", format: 'table'});
+assertEqual(jpeg.json.error, "Fetching tensor for input node `image`: Can't convert value '' of type 'EMPTY' to unsigned integer")
```
guyd: We use back ticks in the docs but for error message I think we have mostly used single or double quotes.

PR #874, opened 2017-03-22
Fix small logic issue with feature_set

mathieumb:
Small bug I found while working on multilabel classifications. Multilabel tests will regression-test it.

MDLB-2159


FinchPowers:
Ticket number please!

FinchPowers:
And a test please.

PR #875, opened 2017-03-23
MLDB-2176 Multilabel classification

mathieumb:
The goal is to handle classification problems where each example has a set of labels instead of a single one. (for example, tagging content)

With 3 options, two trivial for comparison purpose and one-vs-all, which will train a probabilized, binary classifier for each possible label.


mathieumb:
@jeremybarnes @mailletf

plugins/experiment_procedure.cc
```diff
@@ -84,6 +84,14 @@ ExperimentProcedureConfigDescription()
              "Model mode: `boolean`, `regression` or `categorical`. "
```
mailletf: missing multilabel

plugins/experiment_procedure.cc
```diff
@@ -84,6 +84,14 @@ ExperimentProcedureConfigDescription()
...
              "Controls how the label is interpreted and what is the output of the classifier. "
              , CM_BOOLEAN);
+    addField("multilabelStrategy", &ExperimentProcedureConfig::multilabelStrategy,
+             "Multilabel strategy: `random` or `decompose`. "
+             "Controls how examples are prepared to handle multilabel classification. "
+             , MULTILABEL_RANDOM);
+    addField("accuracyOverN", &ExperimentProcedureConfig::accuracyOverN,
+              "Calculate a recall score over the top scoring labels."
+              "Does not apply to boolean or regression modes."
```
mailletf: et categorical ?

plugins/experiment_procedure.cc
```diff
@@ -84,6 +84,14 @@ ExperimentProcedureConfigDescription()
              "Model mode: `boolean`, `regression` or `categorical`. "
              "Controls how the label is interpreted and what is the output of the classifier. "
              , CM_BOOLEAN);
+    addField("multilabelStrategy", &ExperimentProcedureConfig::multilabelStrategy,
+             "Multilabel strategy: `random` or `decompose`. "
+             "Controls how examples are prepared to handle multilabel classification. "
```
mailletf: Ajouter this only applies when mode=multilabel

container_files/public_html/doc/builtin/procedures/Classifier.md
```diff
@@ -25,9 +25,21 @@ The `mode` field controls which mode the classifier will operate in:
...
+  style therefore produces multiple outputs. The `multilabelStrategy` field
+  controls how multilabel classification is handled.
+
+## Multilabel classification
+
+In all operation modes but `multilabel`, the label is a single scalar value. The `multilabel` handles
+categorial classification problems where each example has a set of labels instead of a single one.
+To this end the `label` input must be a row. In this row each column with a non-null value will be a
+label value in the example's set. The column name is used to identify the label, while the value itself is disregarded.
+This makes multi-label classification easy to use with bag of words, for example.
```
mailletf: ça nous prendrait un notebook de démo... je vais pe le faire

container_files/public_html/doc/builtin/procedures/Classifier.md
```diff
@@ -25,9 +25,21 @@ The `mode` field controls which mode the classifier will operate in:
...
 - `regression` mode will use a numeric label, and will predict the value of
   the label itself.
-- `categorical` model will use a categorical (multi-class) label, and will
+- `categorical` mode will use a categorical (multi-class) label, and will
   predict the probability of each of the categories independently.  This
   style therefore produces multiple outputs.
+- `multilabel` mode will use a set of categorical (multi-class) labels, and will
+  predict the probability of each of the categories independently.  This
+  style therefore produces multiple outputs. The `multilabelStrategy` field
+  controls how multilabel classification is handled.
```
mailletf: `multilabel` mode will do [multi-label classification](https://en.wikipedia.org/wiki/Multi-label_classification) by using a set

container_files/public_html/doc/builtin/procedures/Accuracy.md
```diff
@@ -198,6 +198,53 @@ with the same row name as in the input, and the following columns:
...
+The `weighted_statistics` represents the average of the 
+per-label statistics.
+
+Here is a sample output, with `accuracyOverN` set to `3`:
+
+```python
+{
+    "status": {
+                "weightedStatistics": {
+                    "recall over top 3": 0.6666666666666666
```
mailletf: "recall over top 3" -> "recallOverN"

container_files/public_html/doc/builtin/procedures/Accuracy.md
```diff
@@ -198,6 +198,53 @@ with the same row name as in the input, and the following columns:
...
+predictions for each example. I.e, each example will register a positive for 
+each of its label(s) that is found among the set of highest scoring labels returned by the classifier.
+The size of this set is determined by the `accuracyOverN` parameter. It follows that a `1.0` recall rate cannot
+be obtained if any example contains more unique labels than the value of the `accuracyOverN`
+parameter.
+
+The `weighted_statistics` represents the average of the 
+per-label statistics.
+
+Here is a sample output, with `accuracyOverN` set to `3`:
```
mailletf: accuracyOverN -> recallOverN

container_files/public_html/doc/builtin/procedures/Accuracy.md
```diff
@@ -198,6 +198,53 @@ with the same row name as in the input, and the following columns:
...
+
+```python
+{
+    "status": {
+                "weightedStatistics": {
+                    "recall over top 3": 0.6666666666666666
+                },
+                "labelStatistics": {
```
mailletf: Si on veut mettre le "3", je le mettrais dans une clé au dessus de labelStatistics. Genre rajouter "accuracyOverN": 3. Comme ça tu as pas beosin dem odifier ton code si tu change le 3. Le nom des clés vont rester stable.

container_files/public_html/doc/builtin/procedures/Accuracy.md
```diff
@@ -198,6 +198,53 @@ with the same row name as in the input, and the following columns:
...
 * `weight`: the row's assigned weight
 
+### <a name="multilabel"></a>Multi-label mode
+
+The `status` field will contain the recall among top-N performance 
+statistic (e.g. [recall](http://en.wikipedia.org/wiki/Precision_and_recall)) 
+for the classifier, where the labels with the highest score will be chosen as the
+predictions for each example. I.e, each example will register a positive for 
+each of its label(s) that is found among the set of highest scoring labels returned by the classifier.
+The size of this set is determined by the `accuracyOverN` parameter. It follows that a `1.0` recall rate cannot
```
mailletf: accuracy -> recall

container_files/public_html/doc/builtin/procedures/Accuracy.md
```diff
@@ -198,6 +198,53 @@ with the same row name as in the input, and the following columns:
...
+be obtained if any example contains more unique labels than the value of the `accuracyOverN`
```
mailletf: accuracy -> recall

container_files/public_html/doc/builtin/procedures/Accuracy.md
```diff
@@ -198,6 +198,53 @@ with the same row name as in the input, and the following columns:
 * `score`: predicted value
 * `weight`: the row's assigned weight
 
+### <a name="multilabel"></a>Multi-label mode
```
mailletf: Je regarde les métriques ici est sont intéressantes: http://scikit-learn.org/stable/modules/model_evaluation.html#multilabel-ranking-metrics

plugins/experiment_procedure.cc
mathieumb: Oui, ca s'applique a Categorical! Il y a un unit test qui le démontre.

mathieumb:
@jeremybarnes

mathieumb:
@jeremybarnes ping

ml/jml/multilabel_training_data.h
```diff
@@ -0,0 +1,78 @@
...
+            newScores[predictedIndex] = newScore;
+        }
+
+        void initialize(Feature predicted);
+        virtual void sort();
+        virtual Feature_Set * make_copy() const;
+
+    private:
+        const std::shared_ptr<ML::Feature_Set> inner;
+        std::vector<float> newScores;
```
jeremybarnes: We could save a lot of space by using a compact_vector here...

ml/jml/onevsall.cc
```diff
@@ -0,0 +1,165 @@
+// This file is part of MLDB. Copyright 2017 mldb.ai inc. All rights reserved.
+
+/* onevsall.h                                          -*- C++ -*-
+   Mathieu Marquis Bolduc, 8 March 2017
+   Copyright (c) 2017 MLDB.ai  All rights reserved.
+   $Source$
+
+   onevsall classifier
```
jeremybarnes: We need a much better comment here.

ml/jml/onevsall.h
```diff
@@ -0,0 +1,75 @@
...
+
+
+namespace ML {
+
+
+/*****************************************************************************/
+/* ONEVSALL_CLASSIFIER                                                       */
+/*****************************************************************************/
+
+/** Classifier that trains a binary classifier for each possible label */
```
jeremybarnes: Actually a much better comment here.

ml/jml/onevsall.h
```diff
@@ -0,0 +1,75 @@
...
+    virtual OneVsAllClassifier * make_copy() const;
+
+
+    void push(std::shared_ptr<Classifier_Impl> subClassifier,std::shared_ptr<ProbabilizerModel> probabilizer) {
+        subClassifiers.push_back(subClassifier);
+        probabilizers.push_back(probabilizer);
+    }
+
+    std::vector<std::shared_ptr<Classifier_Impl>> subClassifiers;
+    std::vector<std::shared_ptr<ProbabilizerModel>> probabilizers;
```
jeremybarnes: Why do we need probabilizers to come into play here?  Normally you use them on the output, not on sub-classifeirs.

ml/jml/onevsall_generator.cc
```diff
@@ -0,0 +1,168 @@
...
+    weak_learner.reset();
+}
+
+Config_Options
+OneVsAll_Classifier_Generator::
+options() const
+{
+    Config_Options result = Classifier_Generator::options();
+    result       
+        .subconfig("weak_leaner", weak_learner,
```
jeremybarnes: weak_learner has a very specific meaning for boosting and bagging, it's probably not the right name here.

ml/jml/onevsall_generator.cc
```diff
@@ -0,0 +1,168 @@
...
+{
+    auto logger = MLDB::getMldbLog<OneVsAll_Classifier_Generator>();
+
+    std::shared_ptr<OneVsAllClassifier> current = make_shared<OneVsAllClassifier>(model);
+
+    ML::Mutable_Feature_Info labelInfo = ML::Mutable_Feature_Info(ML::BOOLEAN);
+
+    int labelValue = 0; 
+
+    std::shared_ptr<Info_Override_Feature_Space> fs2 = std::make_shared<Info_Override_Feature_Space>(feature_space);
```
jeremybarnes: You should assert on the label being categorical here, so that we get a proper message if we pass a regressor.

ml/jml/onevsall.h
mathieumb: Because we will order the output of the binary classifiers, so they need to be comparable. This was a request from @mailletf.

ml/jml/onevsall_generator.cc
```diff
@@ -0,0 +1,168 @@
...
+           const auto& combinaison = multiLabelList[combinaisonIndex];
+           return std::count(combinaison.begin(), combinaison.end(), labelValue) > 0 ? 1.0f : 0.0f;
+        };
+
+        mutable_trainingData->changePredictedValue(getPredictedScore);
+
+        auto subClassifier = weak_learner->generate(context, *mutable_trainingData, weights, features);
+
+        //Probabilize it
+        std::mutex fvsLock;
```
jeremybarnes: Either lock it or lose it ;)

ml/jml/probabilizer.h
```diff
@@ -332,6 +336,31 @@ operator << (DB::Store_Writer & store, const GLZ_Probabilizer & prob);
...
+    ProbabilizerModel(std::shared_ptr<spdlog::logger> logger);
+
+    //Score, label, weight
+    void train(const std::vector<std::tuple<float, float, float> >& fvs,
+               ML::Link_Function link);
+
+    void serialize(DB::Store_Writer & store) const;
+    void reconstitute(DB::Store_Reader & store);
+
+    std::shared_ptr<spdlog::logger> logger;
```
jeremybarnes: Ugh, no thanks.  We need to get out of the habit of putting shared pointers to loggers everywhere.  If we want to log, then we should get the logger at the point of logging, or pass it in, or make it thread-local, or something.

ml/jml/probabilizer.h
```diff
@@ -332,6 +336,31 @@ operator << (DB::Store_Writer & store, const GLZ_Probabilizer & prob);
 DB::Store_Reader &
 operator >> (DB::Store_Reader & store, GLZ_Probabilizer & prob);
 
+/*****************************************************************************/
+/* PROBABILIZER                                                          */
+/*****************************************************************************/
```
jeremybarnes: What is this class doing?  It's a factoring out of the actual probabilizer data parameters?  Then it should be POD, have no logger, and have no constructor.

ml/jml/training_data.h
```diff
@@ -247,6 +248,8 @@ class Training_Data {
     }
 
     const Dataset_Index & generate_index() const;
+
+    friend Multilabel_Training_Data;
```
jeremybarnes: Which function is needed?  Ideally we would wrap the necessary functionality in a protected virtual function.  This way we're coupling the two together too tightly.

plugins/accuracy.cc
```diff
@@ -243,15 +248,15 @@ runBoolean(AccuracyConfig & runAccuracyConf,
 RunOutput
 runCategorical(AccuracyConfig & runAccuracyConf,
                BoundSelectQuery & selectQuery,
-               std::shared_ptr<Dataset> output)
+               std::shared_ptr<Dataset> output,
+               int accuracyOverN)
 {
-    typedef vector<std::tuple<CellValue, CellValue, double, double, RowPath>> AccumBucket;
+    typedef vector<std::tuple<CellValue, CellValue, std::vector<CellValue>, double>> AccumBucket;
```
jeremybarnes: This will slow everything down, massively.  We are copying a vector on every accumulation.  Instead, make it take a `const CellValue *` or ideally a `double *` and a size_t, and have a separate path for each possibility.

plugins/accuracy.cc
```diff
@@ -261,6 +266,9 @@ runCategorical(AccuracyConfig & runAccuracyConf,
             std::vector<std::tuple<RowPath, CellValue, Date> > outputRow;
 
             static const ColumnPath score("score");
+
+            std::vector<std::pair<double, CellValue>> bestlabelsCandidates;
```
jeremybarnes: Use PossiblyDynamicBuffer instead of vector here

plugins/accuracy.cc
```diff
@@ -261,6 +266,9 @@ runCategorical(AccuracyConfig & runAccuracyConf,
...
+            std::vector<CellValue> bestlabels;
```
jeremybarnes: And here

plugins/accuracy.cc
```diff
@@ -86,6 +86,11 @@ AccuracyConfigDescription()
               "test set is very large and aggregate statistics for each unique score is "
               "sufficient, for instance to generate a ROC curve. This has no effect "
               "for other values of `mode`.", false);
+    addField("accuracyOverN", &AccuracyConfig::accuracyOverN,
```
jeremybarnes: I would make this a vector, so that you can ask for accuracyOverN: [ 1, 3, 10 ] and get all of the values that you're after.  We want accuracy to be one-pass, and in my experience of using this you always want more than one n.

plugins/accuracy.cc
```diff
@@ -277,14 +285,31 @@ runCategorical(AccuracyConfig & runAccuracyConf,
                         outputRow.emplace_back(score + columnName, v, recordDate);
                     }
 
+                    if (accuracyOverN > 1) {
+                        bestlabelsCandidates.emplace_back(v, 
+                            CellValue(jsonDecodeStr<CellValue>(columnName.toSimpleName())));
```
jeremybarnes: You don't need the CellValue constructor here.  Also, you should use CellValue(columnName), not pass through strings.

plugins/accuracy.cc
```diff
@@ -277,14 +285,31 @@ runCategorical(AccuracyConfig & runAccuracyConf,
...
+                    }
+
                     return true;
                 };
             scoreLabelWeight[0].forEachAtom(onAtom);
 
-            auto label = scoreLabelWeight[1].getAtom();
+            if (bestlabelsCandidates.size() > accuracyOverN) {
+                std::sort(bestlabelsCandidates.begin(),
```
jeremybarnes: You should use std::partial_sort here

plugins/accuracy.cc
```diff
@@ -277,14 +285,31 @@ runCategorical(AccuracyConfig & runAccuracyConf,
...
-            auto label = scoreLabelWeight[1].getAtom();
+            if (bestlabelsCandidates.size() > accuracyOverN) {
+                std::sort(bestlabelsCandidates.begin(), 
+                          bestlabelsCandidates.end(), 
+                          std::greater<std::pair<double, CellValue>>());
+                bestlabelsCandidates.erase(bestlabelsCandidates.begin()+accuracyOverN, bestlabelsCandidates.end());
+            }
+
+            for (const auto& pair : bestlabelsCandidates) {
+                bestlabels.push_back(pair.second);
```
jeremybarnes: Reserve bestLabels, since we know the size of it.

plugins/accuracy.cc
```diff
@@ -277,14 +285,31 @@ runCategorical(AccuracyConfig & runAccuracyConf,
...
                     return true;
                 };
             scoreLabelWeight[0].forEachAtom(onAtom);
 
-            auto label = scoreLabelWeight[1].getAtom();
+            if (bestlabelsCandidates.size() > accuracyOverN) {
+                std::sort(bestlabelsCandidates.begin(), 
+                          bestlabelsCandidates.end(), 
+                          std::greater<std::pair<double, CellValue>>());
+                bestlabelsCandidates.erase(bestlabelsCandidates.begin()+accuracyOverN, bestlabelsCandidates.end());
```
jeremybarnes: Big problem here: you aren't handling ties properly.  This will cause problems with reproducibility.  If there are 5 values tied for 3rd place, then you need to give 1/5 of a point for the top 3.  This section needs to be redone.

plugins/accuracy.h
```diff
@@ -28,7 +28,7 @@ struct AccuracyConfig : public ProcedureConfig {
     static constexpr const char * name = "classifier.test";
 
     AccuracyConfig()
-          : mode(CM_BOOLEAN), uniqueScoresOnly(false)
+          : mode(CM_BOOLEAN), uniqueScoresOnly(false), accuracyOverN(1)
```
jeremybarnes: Use member initialization, not a constructor for default values (the configuration classes should have no constructor)

plugins/experiment_procedure.cc
```diff
@@ -84,6 +84,15 @@ ExperimentProcedureConfigDescription()
              "Model mode: `boolean`, `regression` or `categorical`. "
              "Controls how the label is interpreted and what is the output of the classifier. "
              , CM_BOOLEAN);
+    addField("multilabelStrategy", &ExperimentProcedureConfig::multilabelStrategy,
```
jeremybarnes: Need much better documentation here; I don't understand the difference here.

plugins/experiment_procedure.h
```diff
@@ -107,6 +109,11 @@ struct ExperimentProcedureConfig : public ProcedureConfig {
     /// What mode to run in
     ClassifierMode mode;
 
+    // Strategy to handle multilabel
```
jeremybarnes: Again, remove the constructor and use direct member initialization.

plugins/probabilizer.cc
```diff
@@ -195,85 +172,12 @@ run(const ProcedureRunConfig & run,
...
-    repr.style = "glz";
-    repr.link = link;
-    repr.params = probParams;
+    ProbabilizerModel probabilizer(logger);
+    probabilizer.train(fvs, runProcConf.link);
 
     if (!runProcConf.modelFileUrl.toString().empty()) {
         filter_ostream stream(runProcConf.modelFileUrl);
-        stream << jsonEncode(repr);
+        stream << jsonEncode(probabilizer);
```
jeremybarnes: Are you 100% sure that the serialization format is exactly compatible with what was there before?

jeremybarnes:
Lots of small-ish comments; and a couple of more serious issues.  Also careful in the copying of vectors; the accuracy code is part of the basic workflow and it is a productivity sink if it's slow.

In general the approach looks good, it just needs some whipping into shape.  I would also consider moving this to a couple of smaller PRs where you do the refactoring first, secondly you build the basic machinery, and thirdly you modify MLDB to incorporate it.  That way it would be much clearer in terms of impact and intent.

ml/jml/onevsall_generator.cc
mathieumb: Multilabel checks for Null-or-not, and we force the label to be categorical. The only possible mismatch I see is the sub classifier not being able to handle binary classification, in which case it is its role to throw the error.

ml/jml/multilabel_training_data.h
mathieumb: But the raw pointer interface forces me to use the data(), and I hesitate to use unsafe_raw_data()

ml/jml/probabilizer.h
mathieumb: Its my attempt to factor out the ML part of the probabilizer procedure. I found the whole thing a bit confused.

ml/jml/training_data.h
mathieumb: I needed the data_ and index_ of the *inner*. I can get access to the inherited ones but not of the inner ones. I can get around this by either copying the index or exposing its shared pointer publicly.

plugins/accuracy.cc
mathieumb: I'll create a Jira for that if its ok for you

plugins/accuracy.h
mathieumb: Why is that? All of the plugins config classes initialize the defaults in the constructor.

plugins/experiment_procedure.cc
mathieumb: I'll write a bit more, but there is more description in the documentation of the enum and the documentation of the classifier. Should we copy the description of each enum values in the description of every field that uses it?

container_files/public_html/doc/builtin/procedures/Accuracy.md
jeremybarnes: Agreed, spaces in the key names are problematic

container_files/public_html/doc/builtin/procedures/Accuracy.md
jeremybarnes: As I specified elsewhere, this should be recallOverTopN: { 1: x, 3: y , 10 z ] with the values corresponding to configuration parameters.

ml/jml/multilabel_training_data.h
```diff
@@ -57,15 +57,7 @@ class Multilabel_Training_Data : public Training_Data
     };
```
jeremybarnes: Can we remove the include of training_index.h?

plugins/accuracy.cc
```diff
@@ -251,8 +252,12 @@ runCategorical(AccuracyConfig & runAccuracyConf,
                std::shared_ptr<Dataset> output,
                int accuracyOverN)
 {
-    typedef vector<std::tuple<CellValue, CellValue, std::vector<CellValue>, double>> AccumBucket;
+    typedef vector<std::tuple<CellValue, CellValue, double>> AccumBucket;
+    typedef vector<std::tuple<CellValue, CellValue, std::vector<CellValue>, double>> AccumBucketWithBestLabels;
```
jeremybarnes: The inner vector should be compact_vector<CellValue, 1> as normally there will only be one best label, and this avoids memory allocations in that case.

plugins/accuracy.cc
```diff
@@ -286,30 +290,38 @@ runCategorical(AccuracyConfig & runAccuracyConf,
...
                 };
             scoreLabelWeight[0].forEachAtom(onAtom);
 
             if (bestlabelsCandidates.size() > accuracyOverN) {
-                std::sort(bestlabelsCandidates.begin(), 
-                          bestlabelsCandidates.end(), 
+                std::partial_sort(bestlabelsCandidates.data(), 
+                          bestlabelsCandidates.data() + accuracyOverN,
+                          bestlabelsCandidates.data() + bestlabelsCandidates.size(), 
                           std::greater<std::pair<double, CellValue>>());
```
jeremybarnes: Danger danger: if there are NaN values, this std::sort will segfault as x < y and y < x simultaneously.  You need to use a safe wrapper that handles NaN.

ml/jml/multilabel_training_data.h
jeremybarnes: OK

ml/jml/onevsall.h
jeremybarnes: OK, true.  Can you put a comment to that effect in?

ml/jml/probabilizer.h
jeremybarnes: OK, it's much better now it's a pure POD structure.  The last thing I would consider is using a free function to train it rather than a member function, but not a big deal.

plugins/accuracy.cc
jeremybarnes: I don't think it's a lot of work, and it avoids us shipping something with all the technical work done but a poor UX.

plugins/accuracy.h
jeremybarnes: They were written before C++11.  Now we're moving to in-member wherever possible; we should fix where we can.

plugins/experiment_procedure.cc
jeremybarnes: No.  We can embed an enum's documentation in the containing page.  I don't understand why we need this option or what it does.  Maybe a higher level description, or could we just not have an option at all and leave it as a default?  What is the user story for this?

plugins/probabilizer.cc
mathieumb: Yes. I just moved the code.

plugins/accuracy.cc
mathieumb: If we keep only one label I use the other structure.

mathieumb:
Verbal +1 from @jeremybarnes

PR #878, opened 2017-04-07
Mldb 2181 null feature model fix

FinchPowers:


testing/MLDB-2181_null_feature_model_test.py
```diff
@@ -0,0 +1,164 @@
...
+            'id': 'test_c_over_a_or_b_ds',
+            'type': 'sparse.mutable'
+        })
+        for idx in xrange(10):
+            ds.record_row('row{}'.format(idx), [
+                ['line', idx, 0],
+                ['label', 0, 0],
+                ['feature', random.random() + 0.6, 0],
+                ['noise', random.random(), 0],
+                ['a_or_b', 'c', 0]
```
guyd: This wasn't the same problem, right?  Here it is a new value but the type does not change.

FinchPowers: Correct. Do you think I should move it to a separate test file, relabel it, or else?

plugins/dataset_feature_space.cc
```diff
@@ -114,6 +114,9 @@ getColumnInfo(std::shared_ptr<Dataset> dataset,
                 auto categorical = std::make_shared<ML::Fixed_Categorical_Info>(allValues);
                 result.info = ML::Feature_Info(categorical);
             }
+            else if (types.size() == 1 && types.begin()->first == CellValue::EMPTY) {
```
FinchPowers: When I tested, the value type I got was `EMPTY`. If I switch the comparison to `UNKNOWN`, I doubt this will still work.

FinchPowers: As for the case of a constant int, you are right, we would then keep it. If I keep a very narrow view, the current issue is about dealing with NULLs.

We could open a separate ticket to deal with non null features of cardinality one, but they *should* be less problematic as they *should* be of the right type. (Vs null that defaults to numerical which is impossible to determine as correct or wrong.)

testing/MLDB-2181_null_feature_model_test.py
guyd: No.  I just wanted to make sure.  May be a docstring with what the test is doing would be nice.

plugins/dataset_feature_space.cc
guyd: Not sure I understand your comment.  I mean this if case is taken, we should set the `result.info` to `ML::UNKNOWN` and not `ML::INUTILE`.

FinchPowers: Makes sense. Thanks.

testing/MLDB-2181_null_feature_model_test.py
FinchPowers: Done!

PR #884, opened 2017-05-03
Mldb 1209 statstable fncts noise injection

guyd:
This is Francois' branch with a change by vbisserie to make it compile and a unit test for the noise generation functions.  A small issue was found in the NoiseInjector::add_noise where the rounding was always done to the lower value.


plugins/stats_table_procedure.cc
```diff
@@ -372,11 +376,19 @@ run(const ProcedureRunConfig & run,
 /* STATS TABLE FUNCTION                                                      */
 /*****************************************************************************/
 
+static const std::string INJECT_NOISE_DOC_STR =
+     "Inject laplacian noise to counts. This is useful when training "
+     "a classifier on the examples that were used to genereate the counts. "
```
guyd: genereate -> generate

guyd: Should we say that this is acceptable only if the count values are relatively large compare to the noise?  How large?

testing/MLDB-1209_statstable_bias_noise_test.py
```diff
@@ -0,0 +1,132 @@
...
+# Francois Maillet, 2016-07-04
+# This file is part of MLDB. Copyright 2016 Datacratic. All rights reserved.
+#
+
+import datetime
+import random
+
+mldb = mldb_wrapper.wrap(mldb)  # noqa
+
+class MLDB1209StatstableBiasNoiseTest(MldbUnitTest):  # noqa
```
guyd: I had not realized that this is also lacking some validation point.  I will add these too.

plugins/stats_table_procedure.cc
guyd: I suggest we add more on label data leakage in the documentation of stats table functions.  I feel this comment is hard to understand without an example.

testing/MLDB-1209_statstable_bias_noise_test.py
guyd: I've done that.

plugins/stats_table_procedure.h
```diff
@@ -113,10 +113,11 @@ struct StatsTableProcedure: public Procedure {
 
 struct StatsTableFunctionConfig {
     StatsTableFunctionConfig(const Url & modelFileUrl = Url())
-        : modelFileUrl(modelFileUrl)
+        : injectNoise(false), modelFileUrl(modelFileUrl)
     {
     }
 
+    bool injectNoise;
```
mathieumb: We're trying to transition to C++ 11 member initialization (i.e. bool injectNoise = false)

plugins/stats_table_procedure.h
```diff
@@ -245,6 +249,7 @@ struct StatsTablePosNegFunctionConfig {
 
     std::string outcomeToUse;
 
+    bool injectNoise;
```
mathieumb: Same.

utils/noise_injector.h
```diff
@@ -0,0 +1,48 @@
...
+#include <limits>
+
+namespace MLDB {
+/*****************************************************************************/
+/* NOISE INJECTOR                                                            */
+/*****************************************************************************/
+
+struct NoiseInjector {
+
+    NoiseInjector() : mu(0), b(3)
```
mathieumb: Initialize defaults in the declaration and kill the constructor.

PR #885, opened 2017-05-04
MLDB-2188 Classifier label validation

mathieumb:
Validate that no column is used as both label and feature in classifier.

### DIFF
--- a/plugins/classifier.cc
+++ b/plugins/classifier.cc
@@ -303,7 +303,7 @@ run(const ProcedureRunConfig & run,
                     throw HttpReturnException
                             (400, "Dataset column '" + 
                              v.first.toUtf8String()
-                             + "' is used in both label and feature ");
+                             + "' cannot be used in both label and feature ");
             }
             else {
                 for (const auto& c : knownInputColumns) {
@@ -311,7 +311,7 @@ run(const ProcedureRunConfig & run,
                         throw HttpReturnException
                                 (400, "Dataset column '" + 
                                  c.toUtf8String()
-                                 + "' is used in both label and feature");
+                                 + "' cannot be used in both label and feature");
                     }
                 }
             }
@@ -326,7 +326,7 @@ run(const ProcedureRunConfig & run,
                     throw HttpReturnException
                         (400, "Dataset column '" + 
                          c.toUtf8String()
-                         + "' is used in both label and feature "
+                         + "' cannot be used in both label and feature "
                          + "because of label wildcard '" +
                          prefix.toUtf8String() + 
                          "'");
--- a/plugins/classifier.cc
+++ b/plugins/classifier.cc
@@ -298,11 +298,23 @@ run(const ProcedureRunConfig & run,
 
         for (auto& v : labelUnbound.vars) {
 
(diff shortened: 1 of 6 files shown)


testing/MLDB-2143-classifier-utf8.py
```diff
@@ -63,13 +63,14 @@ def test_utf8_category(self):
...
-        "f1Score": 1.0,
-        "precision": 1.0,
-        "accuracy": 1.0
-    })
+            "recall": 0.9444444444444444,
+            "support": 72.0,
+            "f1Score": 0.9446548821548821,
+            "precision": 0.9537037037037037,
+            "accuracy": 0.9645061728395061
+        })
```
FinchPowers: Why are those values changing? Were they incorrect before? Or because the only feature that was used was the label?

testing/MLDB-2188-label_feature_validation.py
```diff
@@ -0,0 +1,216 @@
...
+        cls_config = {
+                "my_fasttext": {
+                    "type": "fasttext",
+                    "verbosity" : 0,
+                    "dims" : 4,
+                    "epoch" : 5,
+                }
+            }
+
+        with self.assertMldbRaises(expected_regexp="Dataset column 'tokens.alabama' is used in both label and feature because of label wildcard 'tokens'"):
```
FinchPowers: A couple of too long lines in this file.

plugins/classifier.cc
```diff
@@ -283,20 +283,60 @@ run(const ProcedureRunConfig & run,
...
+        UnboundEntities labelUnbound = label->getUnbound();
+
+        for (auto& v : labelUnbound.vars) {
+
+            if (runProcConf.mode != CM_MULTILABEL) {
+                if (std::find(knownInputColumns.begin(), knownInputColumns.end(), v.first) != knownInputColumns.end())
+                    throw HttpReturnException
+                            (400, "Dataset column '" + 
+                             v.first.toUtf8String()
+                             + "' is used in both label and feature ");
```
mailletf: is used -> cannot be used

plugins/classifier.cc
```diff
@@ -283,20 +283,60 @@ run(const ProcedureRunConfig & run,
...
+            }
+            else {
+                for (const auto& c : knownInputColumns) {
+                    if (c.startsWith(v.first)) {
+                        throw HttpReturnException
+                                (400, "Dataset column '" + 
+                                 c.toUtf8String()
+                                 + "' is used in both label and feature");
```
mailletf: is used -> cannot be used

testing/MLDB-2143-classifier-utf8.py
mathieumb: We were incorrectly leaving some label columns with the features. This PR is meant to avoid this mistake.

testing/MLDB-2188-label_feature_validation.py
mathieumb: Making multiline strings is a pain though because then you can't use multiline strings to comment out some tests.

PR #887, opened 2017-05-05
Allow compilation with gcc6 and clang; fixes various low-level issues (MLDB-2190)

jeremybarnes:


jeremybarnes:
Note: test failure is spurious (already failing in master since trivial change, azure_blob_storage_test)

PR #888, opened 2017-05-06
MLDB-2191 import text memory allocations

jeremybarnes:
Reduces memory allocations in the tight loop of text import, improving throughput by around 10%.

The test failure is spurious.


jeremybarnes:
It's a private customer benchmark, but any with an import text phase will benefit.

PR #889, opened 2017-05-06
MLDB-2192 value info embedding scalar

jeremybarnes:
Fixes incorrect return shape inference for binary operations between an embedding and a scalar.  Includes a test case.

The test failure is spurious.

### DIFF
--- a/testing/MLDB-1317-tensor-datatype.js
+++ b/testing/MLDB-1317-tensor-datatype.js
@@ -30,6 +30,56 @@ var expected = [
 
 assertEqual(resp, expected);
 
+var resp = mldb.query("select static_expression_info([ [ 10, 20 ], [ 30, 40 ] ]) as *");
+
+var expected = [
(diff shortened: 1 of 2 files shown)


testing/MLDB-1317-tensor-datatype.js
```diff
@@ -30,6 +30,30 @@ var expected = [
 
 assertEqual(resp, expected);
 
+var resp = mldb.query("select static_expression_info([ [ 10, 20 ], [ 30, 40 ] ]) as *");
+
+var expected = [{"columns":[["expr","[ [ 10, 20 ], [ 30, 40 ] ]","-Inf"],["info.isConstant",1,"-Inf"],["info.kind","embedding","-Inf"],["info.shape.0",2,"-Inf"],["info.shape.1",2,"-Inf"],["info.type","INT64","-Inf"]],"rowName":"result"}];
```
FinchPowers: Line too long

PR #890, opened 2017-05-06
MLDB-2193 import text optimizations

jeremybarnes:
Several micro-optimizations that between them allow for roughly doubling the speed of the import.text procedure.

Test coverage on these is pretty extensive on existing tests.

### DIFF
--- a/plugins/importtext_procedure.cc
+++ b/plugins/importtext_procedure.cc
@@ -216,7 +216,8 @@ struct SqlCsvScope: public SqlExpressionMldbScope {
                      const VariableFilter & filter) -> const ExpressionValue &
                 {
                     auto & row = scope.as<RowScope>();
-                    return storage = ExpressionValue(row.row[index], row.ts);
+                    return storage = ExpressionValue(row.row[index],
+                                                     row.ts);
                 },
                 std::make_shared<AtomValueInfo>()};
     }
@@ -968,6 +969,10 @@ struct ImportTextProcedureWorkInstance
         // call the SQL parser
         bool isWhereTrue = config.where->isConstantTrue();
 
+        // Do we have a "NAMED lineNumber()"?  In that case, we can short
+        // circuit the evaluation of that expression.
+        bool isNamedLineNumber = config.named->surface == "lineNumber()";
+
         std::atomic<uint64_t> numSkipped(0);
         std::atomic<uint64_t> totalLinesProcessed(0);
 
@@ -1119,10 +1124,16 @@ struct ImportTextProcedureWorkInstance
                                          0 /* todo: chunk ofs */);
 
             ExpressionValue nameStorage;
-            RowPath rowName(namedBound(row, nameStorage, GET_ALL)
-                                .toUtf8String());
-            row.rowName = &rowName;
+            RowPath rowName;
 
+            if (isNamedLineNumber) {
+                rowName = Path(actualLineNum);
+            }
+            else {
+                rowName = namedBound(row, nameStorage, GET_ALL).coerceToPath();
+            }
+            row.rowName = &rowName;
+            
             // If it doesn't match the where, don't add it
             if (!isWhereTrue) {
                 ExpressionValue storage;
--- a/sql/expression_value.cc
+++ b/sql/expression_value.cc
@@ -1749,7 +1749,10 @@ ExpressionValue(RowValue row) noexcept
     // Structure a flattened representation.  The range between first
     // and last must all start with the same prefix, of length at least
     // level.
-    std::function<Structured (RowValue::iterator first,
+    //
+    // Returns the flattened representation, and a bool telling whether
+    // or not there are duplicate values in it.
+    std::function<std::pair<Structured, bool> (RowValue::iterator first,
                               RowValue::iterator last,
                               size_t level)>
         doLevel = [&] (RowValue::iterator first,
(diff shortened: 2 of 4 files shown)


jeremybarnes:
Failing test is already on master.

sql/expression_value.cc
```diff
@@ -3292,15 +3315,42 @@ void
 ExpressionValue::
 mergeToRowDestructive(StructValue & row)
 {
-    row.reserve(row.size() + rowLength());
+    // This is optimized here because the SelectExpression uses it whenever
+    // merging
+    switch (type_) {
+    case Type::STRUCTURED:
+        if (structured_.unique()) {
+            // We have the only reference in the shared pointer, AND we have
```
mathieumb: Couldnt there still be many references to a shared ptr of the Expression Value? Would it be safe then?

plugins/importtext_procedure.cc
```diff
@@ -1119,10 +1124,16 @@ struct ImportTextProcedureWorkInstance
...
-            RowPath rowName(namedBound(row, nameStorage, GET_ALL)
-                                .toUtf8String());
-            row.rowName = &rowName;
+            RowPath rowName;
 
+            if (isNamedLineNumber) {
+                rowName = Path(actualLineNum);
+            }
+            else {
+                rowName = namedBound(row, nameStorage, GET_ALL).coerceToPath();
```
FinchPowers: `coerceToPath()` vs `toUtf8String`, will the result be utf-8 proof?

sql/sql_expression.cc
```diff
@@ -3119,49 +3128,246 @@ bind(SqlBindingScope & context) const
...
+
+                StructValue row;
+                if (&v == &clauseStorage) {
+                    clauseStorage.mergeToRowDestructive(row);
+                }
+                else {
+                    v.appendToRow(Path(), row);
+                }
+
+                //cerr << "clause " << clauses[i]->surface
```
FinchPowers: log debug

sql/sql_expression.cc
```diff
@@ -3119,49 +3128,246 @@ bind(SqlBindingScope & context) const
...
+                //     << " had result " << jsonEncodeStr(row) << endl;
+
+                // If this throws, the ExpressionValueInfo says that there
+                // are a fixed set of columns, but what returned wasn't
+                // the same (there are more or less of them than expected)
+                if (row.size() != clauseInstructions.size()) {
+                    cerr << "row.size() = " << row.size() << endl;
```
FinchPowers: use the logger

sql/sql_expression.cc
```diff
@@ -3119,49 +3128,246 @@ bind(SqlBindingScope & context) const
...
 
-    return BoundSqlExpression(exec, this, outputInfo);
+        std::vector<ClauseInstructions> instructions;
+        
+        // Now for each one, either place or merge it with the rest
+        for (size_t i = 0;  i < boundClauses.size();  ++i) {
+            const auto & c = boundClauses[i];
+            auto knownColumns = c.info->getKnownColumns();
+
+            //cerr << "knownColumns are " << jsonEncode(knownColumns) << endl;
```
FinchPowers: logger

sql/sql_expression.cc
```diff
@@ -3119,49 +3128,246 @@ bind(SqlBindingScope & context) const
...
+            for (auto & c: knownColumns)
+                clausePrefixes.insert(c.columnName.at(0));
+            
+            std::vector<ClauseInstruction> clauseInstructions;
+            
+            for (auto & p: clausePrefixes) {
+                int idx = elementIndexes[p];
+                int outputNumber = donePrefixes[idx];
+                ++donePrefixes[idx];
+                //cerr << "clause " << i << " " << clauses[i]->surface
```
FinchPowers: logger

sql/sql_expression.cc
```diff
@@ -3119,49 +3128,246 @@ bind(SqlBindingScope & context) const
...
+                //     << " prefix " << p
+                //     << " output " << idx << " output number " << outputNumber
+                //     << endl;
+                clauseInstructions.emplace_back(p, idx, outputNumber == 0);
+            }
+            
+            //cerr << "clauseInstructions.size() = " << clauseInstructions.size()
```
FinchPowers: logger

sql/sql_expression.cc
```diff
@@ -3119,49 +3128,246 @@ bind(SqlBindingScope & context) const
...
+             isConstant);
+
+        auto exec = [=] (const SqlRowScope & context,
+                         ExpressionValue & storage,
+                         const VariableFilter & filter) -> const ExpressionValue &
+            {
+                StructValue result;
+                result.resize(allPrefixes.size());
+
+                //cerr << "executing optimized merge clause" << endl;
```
FinchPowers: ...

sql/sql_expression.cc
```diff
@@ -3119,49 +3128,246 @@ bind(SqlBindingScope & context) const
...
+
+                for (size_t i = 0;  i < clauses.size();  ++i) {
+                    instructions[i].apply(context, filter, result);
+                }
+
+                //cerr << "merged result is " << jsonEncodeStr(result) << endl;
```
FinchPowers: ...

sql/expression_value.cc
```diff
@@ -3292,15 +3315,42 @@ void
...
+    default:
+        break;
+    };
 
+    // Default implementation
+    row.reserve(row.size() + rowLength());
+    
     auto onSubexpr = [&] (PathElement & columnName,
-                          ExpressionValue & val)
+                              ExpressionValue & val)
```
FinchPowers: Bad indent

sql/sql_expression.cc
mathieumb: If we use the logger in cases like this, it'll most likely be turned off if the error happens, which means you wont have the info to help you understand or reproduce the error...

FinchPowers: If the log level is error, I hope it will not be off!!

sql/expression_value.cc
jeremybarnes: Yes... That's the whole point of unique ().  Note that this is a non const method and so only this thread can be in a method now.  It's used in several other places.

plugins/importtext_procedure.cc
jeremybarnes: Yes.  Everything is utf8 proof.

sql/sql_expression.cc
jeremybarnes: In general we have been ok with leaving commented out debug traces in the past.  Have we changed this?

sql/sql_expression.cc
jeremybarnes: This will only print something off there is an exception.  It provides more info for if we get an asset for.

sql/sql_expression.cc
FinchPowers: I push toward using the logger. It's my opinion that if this code is important enough to leave there, it should compile. If it's commented, it will rot.

So it's up to you.

jeremybarnes: The logger has a major overhead a) it needs to be passed everywhere, and b) calling it even when disabled increases code size and runtime.  Leaving them commented out lets a future debugger start off (possibly with updates to the comments) where left off.

plugins/importtext_procedure.cc
jeremybarnes: And note, toUtf8String() is completely wrong to use there, as it doesn't allow for any structured row names.  In other words, this is the only correct way to do this.

jeremybarnes:
@FinchPowers test coverage was addressed in the description.  I will fix the indentation.  I don't think we should make the other changes.  Anything else?

FinchPowers:
Pray proceed. :)

PR #891, opened 2017-05-06
MLDB-2194 double timestamp column format

jeremybarnes:
Reduces memory usage for doubles from 12 to 8 bytes/entry, and from timestamps from 12 to 8 or much less if they are integers.

It also puts in place a mechanism that frozen columns can use other frozen columns as part of their implementation.

### DIFF
--- a/plugins/frozen_column.cc
+++ b/plugins/frozen_column.cc
@@ -13,6 +13,7 @@
 #include "mldb/jml/utils/lightweight_hash.h"
 #include "mldb/http/http_exception.h"
 #include "mldb/utils/atomic_shared_ptr.h"
+#include "mldb/types/value_description.h"
 #include <mutex>
 
 using namespace std;
@@ -58,7 +59,8 @@ struct TableFrozenColumn: public FrozenColumn {
         }
     }
 
-    virtual bool forEach(const ForEachRowFn & onRow) const
+    virtual bool forEachImpl(const ForEachRowFn & onRow,
+                             bool keepNulls) const
     {
         ML::Bit_Extractor<uint32_t> bits(storage.get());
 
@@ -69,6 +71,8 @@ struct TableFrozenColumn: public FrozenColumn {
             if (hasNulls) {
                 if (index > 0)
                     val = table[index - 1];
+                else if (!keepNulls)
+                    continue;  // skip nulls
             }
             else {
                 val = table[index];
@@ -81,9 +85,14 @@ struct TableFrozenColumn: public FrozenColumn {
         return true;
     }
 
+    virtual bool forEach(const ForEachRowFn & onRow) const
+    {
+        return forEachImpl(onRow, false /* keep nulls */);
+    }
+
     virtual bool forEachDense(const ForEachRowFn & onRow) const
     {
-        return forEach(onRow);
+        return forEachImpl(onRow, true /* keep nulls */);
     }
 
     virtual CellValue get(uint32_t rowIndex) const
@@ -605,7 +614,7 @@ struct IntegerFrozenColumn: public FrozenColumn {
             int64_t val = bits.extract<uint64_t>(entryBits);
             if (hasNulls) {
                 if (val == 0) {
-                    if (keepNulls && !onRow(i + firstEntry, CellValue()))
+                    if (keepNulls && (!onRow(i + firstEntry, CellValue())))
                         return false;
                 }
                 else {
@@ -651,7 +660,6 @@ struct IntegerFrozenColumn: public FrozenColumn {
             else return result = val + offset - 1;
         }
         else {
-            TRACE_MSG(logger) << "got val " << val << " " << val + offset;
             return result = val + offset;
         }
     }
@@ -687,7 +695,6 @@ struct IntegerFrozenColumn: public FrozenColumn {
             if (val == 0 && hasNulls)
                 continue;
             allVals.push_back(val);
-            bits.advance(entryBits);
         }
 
         std::sort(allVals.begin(), allVals.end());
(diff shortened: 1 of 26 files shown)


jeremybarnes:
Test failure is again the same spurious on on master.

plugins/frozen_column.cc
```diff
@@ -760,6 +762,349 @@ RegisterFrozenColumnFormatT<IntegerFrozenColumnFormat> regInteger;
...
+        }
+
+        uint64_t val;
+
+        static const uint64_t NULL_BITS
+            = 0ULL  << 63 // sign
+            | (0x7ffULL << 53) // exponent is all 1s for NaN
+            | (0xe1a1ULL); // mantissa
+
+        union U {
```
FinchPowers: Could this have a more meaningful name?

plugins/frozen_column.cc
```diff
@@ -760,6 +762,349 @@ RegisterFrozenColumnFormatT<IntegerFrozenColumnFormat> regInteger;
...
+        SizingInfo info(column);
+        ExcAssertNotEqual(info.bytesRequired, -1);
+
+        firstEntry = column.minRowNumber;
+        numEntries = info.numEntries;
+
+        // Check it's really feasible
+        ExcAssert(column.columnTypes.onlyDoublesAndNulls());
+        Entry * data = new Entry[info.numEntries];
+        storage = std::shared_ptr<Entry>(data, [] (Entry * p) { delete[] p; });
```
FinchPowers: Possibly dumb question, what happens if there is not enough memory to allocate the sared_ptr? Will we leak data? Then the whole program will crash so we don't care much?

plugins/frozen_column.cc
```diff
@@ -760,6 +762,349 @@ RegisterFrozenColumnFormatT<IntegerFrozenColumnFormat> regInteger;
...
+            const CellValue & v = column.indexedVals[r_i.second];
+            if (!v.empty()) {
+                data[r_i.first] = v.toDouble();
+            }
+        }
+    }
+
+    bool forEachImpl(const ForEachRowFn & onRow, bool keepNulls) const
+    {
+        throw HttpReturnException(400, "forEachImpl");
```
FinchPowers: Should the user ever see this message? If yes, the message should be improved. If not, the status should be changed to 500.

plugins/frozen_column.cc
jeremybarnes: It's pretty common to give type punning unions a short name.  I could call it UnionToAccessBitsInDouble but it's imo less readable.

plugins/frozen_column.cc
jeremybarnes: That could only happen on a stack overflow, which will crash the whole program.  There is no way to avoid that; the new has to happen in a separate clause, and good practice is to immediately put the new data into a smart pointer like we're doing here.

plugins/frozen_column.cc
jeremybarnes: In fact, this should be implemented and tested.

jeremybarnes:
@FinchPowers could you have another look please?

sql/cell_value.cc
```diff
@@ -64,6 +64,16 @@ CellValue(const Path & path)
         return;
     }
 
+    if (path.size() == 0) {
+        type = ST_SHORT_PATH;
+        strLength = 0;
+        strFlags = 0;
+
+        cerr << "returned " << jsonEncodeStr(*this) << endl;
```
FinchPowers: Do we want that in prod?

sql/cell_value.cc
```diff
@@ -1009,29 +1038,51 @@ operator <  (const CellValue & other) const
...
-                bool smaller = timeInterval.months < other.timeInterval.months || 
-                   (timeInterval.months == other.timeInterval.months && timeInterval.days < other.timeInterval.days) ||
-                   (timeInterval.months == other.timeInterval.months && timeInterval.days == other.timeInterval.days 
-                    && abs(timeInterval.seconds) < abs(other.timeInterval.seconds));
+                bool nan1 = std::isnan(timeInterval.seconds);
+                bool nan2 = std::isnan(other.timeInterval.seconds);
+
+                if (nan1 || nan2) {
+                    bool smaller = timeInterval.months < other.timeInterval.months || 
+                                                         (timeInterval.months == other.timeInterval.months && timeInterval.days < other.timeInterval.days) ||
```
FinchPowers: Line too long

sql/cell_value.cc
jeremybarnes: no...fixed

jeremybarnes:
@FinchPowers one last time please?

testing/MLDB-1710-left-right-rowname.py
```diff
@@ -78,7 +78,7 @@ def test_outer(self):
         res1 = mldb.query("select rightRowName() from ds1 left join ds2 on ds1.a = ds2.a + 1")
         mldb.log(res1)
 
-        expected = [["_rowName", "rightRowName()"], ["[x]-[]", '""']]
+        expected = [["_rowName", "rightRowName()"], ["[x]-[]", '']]
```
FinchPowers: Is that really supposed to change?

testing/builtin_hash_fct_test.py
```diff
@@ -19,7 +19,7 @@ def test_it(self):
         self.assertEqual(res[1][1], 9281407066453831864)
 
         res = mldb.query("SELECT hash({a: 12, b: 'coco'})")
-        self.assertEqual(res[1][1], 12351211796938911870)
+        self.assertEqual(res[1][1], 17779398697240081926)
```
FinchPowers: Does this mean it breaks predictability?

testing/MLDB-1710-left-right-rowname.py
jeremybarnes: Yes, the previous version was wrong.  This bug was fixed in this branch and the test updated.

testing/builtin_hash_fct_test.py
jeremybarnes: No.  Previously there was a bug where timestamps were taken into account in the hash.  The timestamp included was wrong due to a different bug which was fixed in this branch.  Once fixed, the hash changed; I modified the code such that it shouldn't ever change any more.

If we do have some code that is deployed that relies on hashes, we will need to plan how to recover from that.

PR #895, opened 2017-09-21
Fixed slow make due to re-hashing over and over

jeremybarnes:
Any command line with a semicolon, quote, or parentheses couldn't be saved and hence was re-hashed over and over leading to slow make build times.  This replaces them with underscores and allows this to work, halving the build latency.

### DIFF
--- a/jml-build/functions.mk
+++ b/jml-build/functions.mk
@@ -50,14 +50,22 @@ NOTHING :=
 SPACE := $(NOTHING) $(NOTHING)
 TAB := $(NOTHING)	$(NOTHING)
 DOLLAR := $$
+LPARAN:=(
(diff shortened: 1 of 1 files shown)


jml-build/functions.mk
```diff
@@ -50,14 +50,22 @@ NOTHING :=
 SPACE := $(NOTHING) $(NOTHING)
 TAB := $(NOTHING)	$(NOTHING)
 DOLLAR := $$
+LPARAN:=(
+RPARAN:=)
+SEMICOLON:=;
+QUOTE:="
+#"  
+# (previous line is for emacs)
```
jraby: :rofl:

PR #896, opened 2017-09-25
Ubuntu 1604

jeremybarnes:
This series of commits allows MLDB to be built and tests to pass on either Ubuntu 14.04 or Ubuntu 16.04.

It has been verified on both OSs.

### DIFF
--- a/plugins/sqlite_dataset.cc
+++ b/plugins/sqlite_dataset.cc
@@ -359,7 +359,8 @@ struct SqliteSparseDataset::Itl
     {
         try {
             auto rowNum = runScalarQuery<int>("SELECT rowNum FROM rows WHERE rowHash = ? AND rowName = ?",
-                                              RowHash(rowName), rowName);
+                                              RowHash(rowName), rowName.toUtf8String().rawString().c_str());
+
             auto cols = runQuery<std::tuple<ColumnPath, CellValue, Date> >
                 ("SELECT cols.colName, vals.val, vals.ts FROM vals JOIN cols ON vals.colNum = cols.colNum WHERE rowNum = ?", rowNum);
 
@@ -387,7 +388,9 @@ struct SqliteSparseDataset::Itl
                 explanation.append(record);
             }
 
-            INFO_MSG(logger) << queryStr << "\n" << explanation;
+            cerr << "rowHash = " << int64_t(RowHash(rowName)) << endl;
+            cerr << "rowName = " << rowName << endl;
+            cerr << queryStr << "\n" << explanation << endl;
             abort();
 
             throw;
@@ -480,8 +483,9 @@ struct SqliteSparseDataset::Itl
 
 
 
-        sqlite3pp::query query(db, "SELECT rowNum FROM rows WHERE rowHash = ? LIMIT 1");
+        sqlite3pp::query query(db, "SELECT rowNum FROM rows WHERE rowHash = ? AND rowName = ? LIMIT 1");
         bindArg(query, 1, rowHash);
+        bindArg(query, 2, rowNameStr.c_str());
         for (sqlite3pp::query::iterator i = query.begin(); i != query.end(); ++i) {
             return (*i).get<int>(0);
         }
@@ -627,13 +631,13 @@ struct SqliteSparseDataset::Itl
         doCommand("CREATE TABLE IF NOT EXISTS rows ("
                   "  rowNum INTEGER PRIMARY KEY, "
                   "  rowHash INT NOT NULL, "
-                  "  rowName INT NOT NULL "
+                  "  rowName TEXT NOT NULL "
                   ")");
 
         doCommand("CREATE TABLE IF NOT EXISTS cols ("
                   "  colNum INTEGER PRIMARY KEY, "
                   "  colHash INT NOT NULL, "
-                  "  colName INT NOT NULL "
+                  "  colName TEXT NOT NULL "
                   ")");
 
         doCommand("CREATE UNIQUE INDEX IF NOT EXISTS byrow ON vals (rowNum, colNum, val, ts)");
--- a/sql/path.cc
+++ b/sql/path.cc
@@ -196,8 +196,8 @@ compareElements(const char * s0, size_t l0,
 
 PathElement::
 PathElement()
-    : words{0, 0, 0}
 {
+    std::memset(this, 0, sizeof(PathElement));
 }
 
 PathElement::
@@ -227,14 +227,6 @@ PathElement(const char * str, size_t len)
 PathElement::
 PathElement(const char * str, size_t len, int digits)
 {
-#if 0
-    if (digits != calcDigits(str, len)) {
-        cerr << "for string '" << string(str, len) << "' with length " << len
-             << ": digits = " << digits
-             << endl;
-    }
-    ExcAssertEqual(digits, calcDigits(str, len));
-#endif
     initChars(str, len, digits);
 }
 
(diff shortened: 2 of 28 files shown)


plugins/sqlite_dataset.cc
```diff
@@ -387,7 +388,9 @@ struct SqliteSparseDataset::Itl
                 explanation.append(record);
             }
 
-            INFO_MSG(logger) << queryStr << "\n" << explanation;
+            cerr << "rowHash = " << int64_t(RowHash(rowName)) << endl;
+            cerr << "rowName = " << rowName << endl;
+            cerr << queryStr << "\n" << explanation << endl;
```
FinchPowers: Seems to be the opposite of where we want to go.

plugins/sqlite_dataset.cc
```diff
@@ -627,13 +631,13 @@ struct SqliteSparseDataset::Itl
...
                   "  rowHash INT NOT NULL, "
-                  "  rowName INT NOT NULL "
+                  "  rowName TEXT NOT NULL "
                   ")");
 
         doCommand("CREATE TABLE IF NOT EXISTS cols ("
                   "  colNum INTEGER PRIMARY KEY, "
                   "  colHash INT NOT NULL, "
-                  "  colName INT NOT NULL "
+                  "  colName TEXT NOT NULL "
```
FinchPowers: Is that really realted to Ubuntu 16.04?

sql/path.cc
```diff
@@ -760,18 +727,17 @@ initStringUnchecked(T && str)
...
+    if (l == 0) {
+        storage_.append("\0", 1);
     }
     else {
-        complex_ = 1;
-        new (&this->str.str) Utf8String(std::move(str));
+        storage_.append(d, l);
     }
+
+    //ExcAssertEqual(toUtf8String(), str);
```
FinchPowers: rm if unecessary

sql/path.cc
```diff
@@ -783,22 +749,17 @@ void
...
+        storage_.append("\0", 1);
     }
     else {
-        complex_ = 1;
-        new (&this->str.str) Utf8String(str, len);
+        storage_.append(str, len);
     }
+    digits_ = digits;
+
+    //ExcAssertEqual(toUtf8String(), Utf8String(str, len));
```
FinchPowers: rm

sql/testing/path_order_test.cc
```diff
@@ -0,0 +1,85 @@
...
+        }
+
+        BOOST_CHECK_EQUAL(ordered.size(), paths.size());
+        BOOST_CHECK_EQUAL(unordered.size(), paths.size());
+        BOOST_CHECK_EQUAL(orderedHashes.size(), paths.size());
+        BOOST_CHECK_EQUAL(unorderedHashes.size(), paths.size());
+
+        for (const Path & p: paths) {
+            if (!ordered.count(p))
+                cerr << "missing path " << p << endl;
```
FinchPowers: Is that an error? Should the test fail?

FinchPowers: I see it's a code move, but I still ask the question.

testing/MLDB-2166_mime_type.py
```diff
@@ -14,32 +14,32 @@ def setUpClass(cls):
...
+        #         [
+        #             "_rowName",
+        #             "mime"
+        #         ],
+        #         [
+        #             "result",
+        #             "Python script, ASCII text executable"
+        #         ]
+        #     ]
+        # )
```
FinchPowers: Why?

utils/interned_string.h
```diff
@@ -0,0 +1,222 @@
...
+            std::swap(internalWords[i], other.internalWords[i]);
+        }
+    }
+
+    size_t size() const
+    {
+        return isExt() ? extLength_ : intLength_;
+    }
+
+    size_t length() const
```
FinchPowers: Since this is called from a nothrow ctor, I think this should also be marked as nothrow.

utils/interned_string.h
```diff
@@ -0,0 +1,222 @@
...
+    }
+
+    void reserve(size_t newCapacity)
+    {
+        if (newCapacity < capacity())
+            return;
+
+        bool wasExt = isExt();
+
+        char * newBytes = new Char[newCapacity];
```
FinchPowers: I think in case of error we may have a memory leak here.

utils/interned_string.h
```diff
@@ -0,0 +1,222 @@
...
+        size_t l = size();
```
FinchPowers: Or if this is a "no except" section, size() should also be declared with "noexcept".
If it is the case, I would surroumd the section with clear comments that this is a "memory safe zone".

plugins/sqlite_dataset.cc
jeremybarnes: It's a pre-crash log.  We don't want to log to a logging subsystem just before we crash the system.

plugins/sqlite_dataset.cc
jeremybarnes: Yes, it fixes something that worked on 14.04 but no longer does on 16.04.

testing/MLDB-2166_mime_type.py
jeremybarnes: The library in 16.04 incorrectly detects that as C++ source, not Python.

utils/interned_string.h
jeremybarnes: This is a code move.  Note that the nothrow does not require functions it calls to be nothrow; it just says that if one of them does throw, the exception won't propagate outside.

utils/interned_string.h
jeremybarnes: You are right, but again this moves outside of the scope (it was a code move).

utils/interned_string.h
FinchPowers: Doesn't it also make it un catchable and crash the system very badly?

sql/path.cc
jeremybarnes: For me it's helpful documentation.  This is a postcondition, that is too expensive to check, but is helpful to understand what is going on.

sql/path.cc
jeremybarnes: Same, this is documentation.

plugins/sqlite_dataset.cc
FinchPowers: Adding that as a comment in the code would be helpful.

testing/MLDB-2166_mime_type.py
FinchPowers: Meh weird regression... Although it's not core, are we ok with it?

plugins/sqlite_dataset.cc
jeremybarnes: It does call abort() immediately after... it's pretty clear from the context.

testing/MLDB-2166_mime_type.py
jeremybarnes: We're testing a library that's provided with the OS, so this is basically an OS regression.  I don't really see what we can do about it.

utils/interned_string.h
jeremybarnes: Yes!  Which is kind of the point: if something in there does throw (note that marking nothrow doesn't change at all whether something will) then it will noisily crash the program, as it's a precondition violation.  Which is exactly what you would want it to do.

Note that the reason noexcept was added was to enable things like vector resizes to be both performant and strongly exception safe; that's the reason for the annotation here.

sql/testing/path_order_test.cc
jeremybarnes: It fails further down; this provides a more helpful message for if it does fail.

utils/interned_string.h
jeremybarnes: I have added a comment and a static_assert for the PODness to make it clear.

PR #903, opened 2018-09-06
Python3

jeremybarnes:
This portemanteau branch was aimed at the modernization of the MLDB codebase and includes the following:
- Move from Python 2 to Python 3
- Updates to run on Ubuntu 18.04 while maintaining compatibility with ubuntu 16.04
- Major refactoring to move MLDB towards a more modular structure and eventually separate plugins entirely from the main tree
- Major compile time and test run time improvements; it's now quite feasible to develop MLDB on a 2013 era laptop and it compiles and runs tests in around 5 minutes on a modern server machine
- Disabling of the Tensorflow, MongoDB and Azure plugins for now, with a view to them eventually becoming separate plugins
- A complete overhaul of the Python support, fixing numerous GIL locking and re-entrancy bugs
- Removal of magic objects injected into the Python interpreter; MLDB is now accessed using `from mldb import mldb`
- Ability to import mldb from a pure Python script, which will embed MLDB within the Python instance without code changes.  This is an important step to having MLDB able to be distributed by pip
- Numerous test fixes and bug fixes
- MLDB now uses C++17 and requires G++ 6+ or clang 6 to compile
- Compilation fixes for gcc 7 and gcc8
- Numerous code standardizations to use C++17 features instead of earlier alternatives where possible
- Numerous decoupling of internal and external interfaces
- Much dead code removed
- Much smaller boost footprint in the code
- Sensor support in a plugin, with skeleton audio and video sensors


FinchPowers:
🎉

PR #914, opened 2021-08-03
OSX support

jeremybarnes:
Removes the Linuxisms from the code and build process of MLDB and generalizes the interfaces to many of the OS primitives which allow for MLDB to be built (and all tests to pass) under Darwin / OSX.


FinchPowers:
@jeremybarnes I will try this branch here as I have OSX but I don't have an M1.

FinchPowers:
```
$ make -j12
mldb/jml-build/clang.mk:13: building with clang++ version Apple clang version 12.0.0 (clang1200.0.32.29)
mldb/jml-build/python.mk:6: PYTHON_VERSION_DETECTED=3.9
mldb/testing//testing.mk:39: mldb/testing/mldb_sample_plugin//mldb_sample_plugin.mk: No such file or directory
make: *** No rule to make target 'mldb/testing/mldb_sample_plugin//mldb_sample_plugin.mk'.  Stop.
```

PR #942, opened 2021-08-24
OSX build instructions

jeremybarnes:
Update build instructions for OSX (they have been tested).

### DIFF
--- a/Building-OSX.md
+++ b/Building-OSX.md
@@ -1,2 +1,107 @@
-brew install rust gnu-time icu4c libmagic boost boost-python3 coreutils libssh2 lz4 openssl python@3.9 snappy v8 xz yaml-cpp
+# Building MLDB on OSX
+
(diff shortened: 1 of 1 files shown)


Building-OSX.md
```diff
@@ -1,2 +1,171 @@
...
+```
+
+## Compile MLDB
+
+This process is long.  You should use an argument to `-j` which is equal to the number of processors cores
+in your system; be aware however that some files require lots of memory to build (more than 500MB) and
+so you also need to ensure that your machine has sufficient RAM for the compilation.
+
+```
+make -j6 -k compile
```
FinchPowers, line 121: we used to do `nice make -j $(nproc) compile` :)

jeremybarnes: That unfortunately runs my laptop out of memory (it thinks it has 12 cores, but there are only 6 real ones, and I don't have enough memory to compile simultaneously on the 12).

PR #972, opened 2024-12-02
Compile on OSX Sonoma

jeremybarnes:


jeremybarnes:
Various updates to target the new compiler versions on OSX sonoma and generally make it build again.

Notable:
- A new way of filtering options by compiler
- Python now sets its directories carefully on startup to look in the virtualenv so no need to install system packages anymore

